PR #5, opened 2020-03-09
#171691191 configure database (mongodb)

tobieeniafe:
**What does this PR do**
1. Add environment variables
2. Add mongoose 
3. Add mongo db connectivity

**Description of task to be completed**
Setup environment variables so app will run as it should in development and production modes

**How should this be manually tested?**
1. Clone the rope to your machine running nodejs and mongodb
2. Open terminal in the cloned folder and run `npm install && npm run start:dev`
 
**Any background context you want to provide?**
`GET /api/v1/welcome` is the only available endpoint

**What are the relevant Pivotal tracker stories?**
#171691191]


config/config.js
```diff
@@ -0,0 +1,15 @@
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+if (process.env.NODE_ENV === 'production') {
```
andela-hpopoola: I think it will be better to consolidate the environments together and load the development configuration if .env is not found. This is to ensure that we are not forced to use the default you defined (irrespective of the environment) and also caters for other environments (e.g STAGING and TEST). Something like this will be more productive imho.

```suggestion
module.exports = {
    DB_URL: process.env.DB_URL || 'mongodb://localhost:27017/ballers',
    PORT: process.env.PORT || 3000,
  };
```

tobieeniafe: the `DB_URL` localhost would have to come first, or we'd be pushing to the production db everytime

PR #9, opened 2020-03-23
#171808935 create registration endpoint

tobieeniafe:
#### What does this PR do?
Creates registration endpoint

Have the following endpoints working
`POST /api/v1/user/welcome`

From postman, Send a `POST` request to `api/v1/user/welcome`
With the following parameters in JSON format
`"firstName"
    "lastName"
    "email"
    "password"
    "confirmPassword"
    "phone"`

#171808935


config/config.js
```diff
@@ -4,5 +4,6 @@ dotenv.config();
 
 const DB_URL = process.env.DB_URL || 'mongodb://localhost:27017/ballers';
 const PORT = process.env.PORT || 3000;
+const { USER_SECRET } = process.env;
```
andela-hpopoola: Can we have a default value for the USER_SECRET here so that we are not forced to create a `.env` file on development for the USER_SECRET.

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
+import mongoose from 'mongoose';
+import bcrypt from 'bcryptjs';
+
+const UserSchema = new mongoose.Schema({
+  firstname: {
```
andela-hpopoola: Can we update this to a camel case instead to be more consistent with the name convention? I mean we should use `firstName` instead of `firstname`.

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+    type: String,
+    required: true,
+  },
```
andela-hpopoola: Can we ensure that the value saved in firstName is trimmed before saving? Adding `trim: true` should do the trick. This should also be added to the lastName.

andela-hpopoola:
I noticed that you placed the user model file (controllers/user.model.js) in the controller folder. Can we ensure that the controllers folder is for controllers files only and the model folder for model files only. Simply put, the `user.model.js` should be placed in a model folder.

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+  lastname: {
```
andela-hpopoola: Same camelCase comment as stated above -`lastName` instead of `lastname`

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+    type: String,
+    required: true,
+  },
```
andela-hpopoola: Can we trim this too, please?

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+  email: {
+    type: String,
+    required: true,
```
andela-hpopoola: Can we add the `unique` option here to ensure that emails are always unique in the database?

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+  password: {
+    type: String,
+    required: true,
+  },
+  phone: {
+    type: String,
+    required: true,
+  },
+  createdAt: {
+    type: String,
```
andela-hpopoola: **Wrong Datatype:** createdAt and updatedAt should have a data type of `Date` not `String`. However, from mongoose documentation, it is better to use timestamp so that both fields are handled by mongoose instead of manually. https://mongoosejs.com/docs/guide.html#timestamps

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+    required: true,
+  },
+  updatedAt: {
```
andela-hpopoola: Same as above.

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+    type: String,
+    required: true,
+  },
+
+});
+
+const User = mongoose.model('User', UserSchema);
+
+User.getUserByEmail = (email, callback) => {
```
andela-hpopoola: **Separation of Concerns:** Can we move this into another file, please? I think it will be better to move the codes here to a services file (services/user.services.js) to ensure that we have appropriate separation of concerns. The code will be cleaner and easier to maintain if the model files handle the database fields and the services files handle the database manipulation.

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+  const query = { email };
+  User.findOne(query, callback);
```
andela-hpopoola: **Nit:** Can we update the callbacks to promises, please? It is preferable to use promises over callbacks to avoid callback hell. Promises are easier to work with compared to callbacks.

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+User.getUserByPhone = (phone, callback) => {
```
andela-hpopoola: Can we remove this for the main time? I can think of a suitable reason why we will want to get a user by phone? If you have a good reason, can you explain it please?

controllers/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+  const query = { phone };
+  User.findOne(query, callback);
+};
+
+User.addUser = (newUser, callback) => {
```
andela-hpopoola: This should be moved to the services class and converted to a promise.

routes/user.js
```diff
@@ -0,0 +1,58 @@
+import express from 'express';
+import moment from 'moment';
+import jwt from 'jsonwebtoken';
+import { USER_SECRET } from '../config/config';
+import User from '../controllers/user.model';
+
+const router = express.Router();
+const now = `${moment().format('LLLL')} ${new Date().toString().match(/([A-Z]+[\+-][0-9]+.*)/)[1]}`;
```
andela-hpopoola: Since we are autogenerating the createdAt and updatedAt fields, I don't think we need this line again.

routes/user.js
```diff
@@ -0,0 +1,58 @@
+import express from 'express';
+import moment from 'moment';
```
andela-hpopoola: Can we uninstall `moment` and use `datefns` for handling date functionalities when needed? They do virtually the same thing but `datefns` has a smaller file size and is recommended to smaller projects like this. https://javascript-conference.com/blog/why-you-should-use-date-fns-for-manipulating-dates-with-javascript/

routes/user.js
```diff
@@ -0,0 +1,58 @@
...
+import jwt from 'jsonwebtoken';
+import { USER_SECRET } from '../config/config';
+import User from '../controllers/user.model';
+
+const router = express.Router();
+const now = `${moment().format('LLLL')} ${new Date().toString().match(/([A-Z]+[\+-][0-9]+.*)/)[1]}`;
+
+router.post('/register', (req, res) => {
+  const newUser = new User({
```
andela-hpopoola: **Separation of Concerns:** Can we move all the codes in the post request into a controller file to ensure proper separation of concerns. Here is a good example showing proper separation on concern - https://riptutorial.com/node-js/example/32332/model-routes-controllers-services-code-structure (Please note you will need to convert the code to ES6 if you are using any part)

routes/user.js
```diff
@@ -0,0 +1,58 @@
...
+    firstname: req.body.firstname,
+    lastname: req.body.lastname,
+    email: req.body.email,
+    password: req.body.password,
+    phone: req.body.phone,
+    createdAt: now,
+    updatedAt: now,
+  });
+
+  if (newUser.password !== req.body.confirmPassword) {
```
andela-hpopoola: I think it would be easier to use an external package for data validation. I will recommend Joi (https://github.com/hapijs/joi).

routes/user.js
```diff
@@ -0,0 +1,58 @@
...
+        res.status(200).send({ success: false, message: 'Email is linked to another account' });
+      } else {
+        User.getUserByPhone(newUser.phone, (error1, user1) => {
+          if (error1) {
+            res.status(500).send({ success: false, message: 'Something\'s not right (phone)' });
+          }
+          if (user1) {
+            res.status(200).send({ success: false, message: 'Phone number is linked to another account' });
+          } else {
+            User.addUser(newUser, (error2, user2) => {
```
andela-hpopoola: The code here is a little difficult to understand due to the use of callbacks (user, user1, user2, error, error1, error2), I believe it will be more readable when you update it to promises.

routes/user.js
```diff
@@ -0,0 +1,58 @@
...
+              if (error2) {
+                res.status(500).send({ success: false, message: 'Something\'s not right (add)' });
+              } else {
+                user2.email = null;
+                user2.password = null;
+                user2.phone = null;
+                const token = jwt.sign(user2.toJSON(), USER_SECRET, {
```
andela-hpopoola: Nice work in encrypting the token. However, can we use the user Id for now instead of user2.toJSON(). Can we also update the expiry date to 30days for now?

andela-hpopoola:
Thank you for working on this. Can you ensure that you write tests for the new codes you added?

controllers/user.model.js
tobieeniafe: I feel emails and phone numbers should be unique 🤔so to check if the number to be registered already exists in the db

tobieeniafe: I feel emails and phone numbers should be unique 🤔so to check if the number to be registered already exists in the db

routes/user.js
tobieeniafe: According to the article, the size only matters in frontend applications. And this isn't a frontend application. Besides there's more pressing work to do than having to go through the documentation to find the best format again.

> This doesn’t matter too much in a backend application with Node.js though, since the package size is a minor concern. You can use date-fns just like Moment.js, in the frontend browser. It’s here that the package size is decisive.

routes/user.js
tobieeniafe: adding `joi` would require either adding `joigoose` or removing `mongoose`. So what's it going to be?

andela-hpopoola: Let's add `joigoose`, please.

controllers/user.model.js
andela-hpopoola: Nice thoughts, @tobieeniafe. However, from experience, some people find it difficult to release their personal phone numbers on any platform which would make the phone number field to be sometimes null. I don't want us to focus on things we might not use.

routes/user.js
```diff
@@ -0,0 +1,40 @@
...
+    phone: req.body.phone,
+  });
+  if (req.body.password === req.body.confirmPassword) {
+    getEmail(newUser.email)
+      .then((user) => {
+        if (user === true) {
+          res.status(200).json({ success: false, message: 'Email is linked to another account' });
+        } else {
+          addUser(newUser).then(() => {
+            const token = jwt.sign(newUser._id.toJSON(), USER_SECRET, {
```
andela-hpopoola: `newUser._id` is currently not defined. This is causing the code to throw an error, hence the catch block is executed.

PR #15, opened 2020-04-06
#172045791 Login endpoint

tobieeniafe:
#### What does this PR do?
Add login endpoint

Add login endpoint for user authentication

Visit `/api/v1/user/login`
Parameters `email` `password`

#172045791

### DIFF
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -12,8 +12,8 @@ const UserController = {
   login(req, res, next) {
     const user = req.locals;
     loginUser(user)
-      .then((token) => {
-        res.status(200).json({ success: true, message: 'Login successful', token });
+      .then((payload) => {
+        res.status(200).json({ success: true, message: 'Login successful', payload });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -1,4 +1,4 @@
-import { addUser } from '../services/user.service';
+import { addUser, loginUser } from '../services/user.service';
 
 const UserController = {
   register(req, res, next) {
@@ -6,6 +6,14 @@ const UserController = {
     addUser(user)
       .then((token) => {
         res.status(201).json({ success: true, message: 'User registered', token });
+      })
+      .catch((error) => next(error));
+  },
+  login(req, res, next) {
+    const user = req.locals;
+    loginUser(user)
+      .then((token) => {
+        res.status(200).json({ success: true, message: 'Login successful', token });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -10,10 +10,10 @@ const UserController = {
       .catch((error) => next(error));
   },
   login(req, res, next) {
-    const user = req.locals;
-    loginUser(user)
-      .then((payload) => {
-        res.status(200).json({ success: true, message: 'Login successful', payload });
+    const loginDetails = req.locals;
+    loginUser(loginDetails)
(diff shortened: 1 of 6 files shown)


server/controllers/user.controllers.js
```diff
@@ -9,6 +9,14 @@ const UserController = {
       })
       .catch((error) => next(error));
   },
+  login(req, res, next) {
+    const user = req.locals;
+    loginUser(user)
+      .then((payload) => {
+        res.status(200).json({ success: true, message: 'Login successful', payload });
```
andela-hpopoola, line 16: Can we use `user` in the returned object instead of payload? I feel it will be easier to reference user when you are consuming the API than using payload.

server/services/user.service.js
```diff
@@ -35,3 +43,32 @@ export const addUser = async (user) => {
...
+        const payload = {
+          firstName: existingUser.firstName,
+          lastName: existingUser.lastName,
+          email: existingUser.email,
+          phone: existingUser.phone,
+          token,
+        };
+        return payload;
+      }
+      throw new ErrorHandler(500, 'Internal Server Error NE');
```
andela-hpopoola, line 67: I think the ending `NE` is a typo. Can you remove it, please?

test/controllers/user.controller.test.js
```diff
@@ -205,3 +205,58 @@ describe('Register Route', () => {
...
+    });
+    it('returns successful payload', (done) => {
+      request()
+        .post('/api/v1/user/login')
+        .send(userLogin)
+        .end((err, res) => {
+          expect(res).to.have.status(200);
+          expect(res.body.success).to.be.eql(true);
+          expect(res.body.message).to.be.eql('Login successful');
+          expect(res.body.payload).to.have.property('firstName');
```
andela-hpopoola, line 224: I think it will be better to compare `firstName`, `lastName`, and `email` to the original value (user) saved in the database instead of testing if the properties are present. This will make the test fail if the code is retrieving thee wrong record due to coding bug / or any other reason.

test/services/user.service.test.js
```diff
@@ -173,4 +201,58 @@ describe('User Service', () => {
...
+
+    context('when a valid user is entered', () => {
+      it('comparePassword should return true', async () => {
+        const hash = (await getUserByEmail(userDetails.email)).password;
+        const compareResponse = await comparePassword(userDetails.password, hash);
+        expect(compareResponse).to.eql(true);
+      });
+    });
+
+    context('when existingUser returns an error', () => {
```
andela-hpopoola, line 220: The context here should be `when the User model returns an error`

test/controllers/user.controller.test.js
```diff
@@ -221,10 +221,9 @@ describe('Login Route', () => {
           expect(res).to.have.status(200);
           expect(res.body.success).to.be.eql(true);
           expect(res.body.message).to.be.eql('Login successful');
-          expect(res.body.payload).to.have.property('firstName');
-          expect(res.body.payload).to.have.property('lastName');
-          expect(res.body.payload).to.have.property('email');
-          expect(res.body.payload).to.have.property('token');
+          expect(res.body.user.firstName).to.be.eql(user.firstName);
+          expect(res.body.user.lastName).to.be.eql(user.lastName);
+          expect(res.body.user.email).to.be.eql(user.email);
```
andela-hpopoola, line 226: Can you return the test for the `token`? We need to confirm that the token is returned when a user login. Not compulsory now, but we might need to update the tests later to confirm that the right token is returned.
```suggestion
          expect(res.body.user.email).to.be.eql(user.email);
          expect(res.body.payload).to.have.property('token');
```

PR #16, opened 2020-04-10
#172169387 Social media endpoint

tobieeniafe:
#### What does this PR do?
Adds social media login

Allows registered users to login into the platform with their Facebook and Google accounts respectively

`/api/v1/user/auth/facebook/` for Facebook
`/api/v1/user/auth/google/` for Google

#172169387

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -3,6 +3,7 @@ import bodyParser from 'body-parser';
 import cors from 'cors';
 import morgan from 'morgan';
 import winston from 'winston';
+import passport from 'passport';
 import { PORT, startDBConnection } from './config';
 import logger from './config/winston';
 import routes from './routes';
@@ -14,6 +15,7 @@ app.use(cors());
 app.use(bodyParser.urlencoded({ extended: false }));
 app.use(bodyParser.json());
 app.use(morgan('combined', { stream: winston.stream.write }));
+app.use(passport.initialize());
 
 routes(app);
 app.use(errorMiddleware);
--- a/server/routes/user.routes.js
+++ b/server/routes/user.routes.js
@@ -1,4 +1,5 @@
 import express from 'express';
+import passport from 'passport';
 import { registerSchema, loginSchema } from '../schemas/user.schema';
 import { schemaValidation } from '../helpers/middleware';
 import UserController from '../controllers/user.controllers';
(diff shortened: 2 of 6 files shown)


server/index.js
```diff
@@ -14,6 +15,7 @@ app.use(cors());
 app.use(bodyParser.urlencoded({ extended: false }));
 app.use(bodyParser.json());
 app.use(morgan('combined', { stream: winston.stream.write }));
+app.use(passport.initialize());
```
andela-hpopoola, line 18: Can we move this to the passport config file?

server/services/user.service.js
```diff
@@ -1,9 +1,17 @@
+/* eslint-disable no-underscore-dangle */
```
andela-hpopoola, line 1: Can you add this to the line where it is used? It is not a good idea to disable esliint for the the whole file.

server/services/user.service.js
```diff
@@ -72,3 +80,69 @@ export const loginUser = async (user) => {
     throw new ErrorHandler(401, 'Invalid email or password');
   }
 };
+
+export const loginSocialMedia = async (email) => {
```
andela-hpopoola, line 84: I believe we can improve the implementation here. When a user tries to login via social media, we need to check if the user email is in the database (i.e. the user has previously registered), if yes, we can log them in via the available account in the database. If No, we use create an account for them and log them in with the newly created account.

Also, can the name be updated to `loginViaSocialMedia`?

server/routes/user.routes.js
```diff
@@ -65,4 +66,51 @@ router.post('/register', schemaValidation(registerSchema), UserController.regist
 
 router.post('/login', schemaValidation(loginSchema), UserController.login);
 
-module.exports = router;
+router.get(
```
andela-hpopoola, line 69: Let's move all passport related routes into passport.routes.js to make it easier to maintain.

server/services/user.service.js
tobieeniafe: line 114 had two flags so moved to underscore up

andela-hpopoola: Can you remove it now? I don't see any use for it in the codebase.

server/services/passport.service.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-underscore-dangle */
```
andela-hpopoola, line 1: Lets remove this and add `_json` to the allowable values in .eslintrc. https://github.com/Highrachy/ballers-api/blob/ac1c9dd2bda8c2a60f311edbae21f1232d121b0c/.eslintrc#L19

PR #22, opened 2020-07-03
#173216044 add property endpoint

tobieeniafe:
#### What does this PR do?
Add property andpoint for dashboard

Create endpoints to add, view, update & delete properties

`POST /api/v1/property/add`
`PUT /api/v1/property/update`
`DELETE /api/v1/property/delete/:id`
`GET /api/v1/property/all`
`GET /api/v1/property/:id`

`{
  "name": "5 bedroom terrace duplex",
  "location": "lagos",
  "price": 40000000,
  "units": 3,
  "houseType": "5 bedroom terrace duplex",
  "bedrooms": 5,
  "toilets": 7,
  "description": "Newly built 5 bedroom terrace duplex",
  "floorPlans": "http://linktoplan.ng",
  "mapLocation": {"longitude": "1.23456", "latitude": "2.34567"},
  "neighborhood": ["Lekki phase 1"],
  "mainImage": "https://picsum.photos/200",
  "gallery": ["https://picsum.photos/200", "https://picsum.photos/200", "https://picsum.photos/200"]
}`

All requests should be sent with token as header `Authorization`

#173216044

### DIFF
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -50,7 +50,7 @@ const PropertyController = {
   getOne(req, res, next) {
     getPropertyById(req.params.id)
       .then((property) => {
-        if (property.length < 1) {
+        if (!property) {
           res
             .status(httpStatus.NOT_FOUND)
             .json({ success: false, message: 'Property does not exist' });
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -21,8 +21,8 @@ const PropertyController = {
     const updatedproperty = req.locals;
     const { user } = req;
     updateProperty({ ...updatedproperty, updatedBy: user._id })
-      .then(() => {
-        res.status(httpStatus.OK).json({ success: true, message: 'Property updated' });
+      .then((property) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Property updated', property });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -10,8 +10,7 @@ const PropertyController = {
   add(req, res, next) {
     const property = req.locals;
     const { user } = req;
-    property.owner = user._id;
-    addProperty(property)
+    addProperty({ ...property, owner: user._id })
       .then(() => {
         res.status(201).json({ success: true, message: 'Property added' });
       })
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -13,7 +13,7 @@ const PropertyController = {
     const { user } = req;
     addProperty({ ...property, addedBy: user._id, updatedBy: user._id })
       .then(() => {
-        res.status(httpStatus.CREATED).json({ success: true, message: 'Property added' });
+        res.status(httpStatus.CREATED).json({ success: true, message: 'Property added', property });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -2,82 +2,60 @@ import {
   addProperty,
   updateProperty,
   deleteProperty,
-  getPropertyByRefNo,
-  getOwnedProperties,
+  getPropertyById,
+  getAllProperties,
 } from '../services/property.service';
+import httpStatus from '../helpers/httpStatus';
 
 const PropertyController = {
   add(req, res, next) {
     const property = req.locals;
     const { user } = req;
-    addProperty({ ...property, owner: user._id })
+    addProperty({ ...property, addedBy: user._id, updatedBy: user._id })
       .then(() => {
-        res.status(201).json({ success: true, message: 'Property added' });
+        res.status(httpStatus.CREATED).json({ success: true, message: 'Property added' });
       })
       .catch((error) => next(error));
   },
   update(req, res, next) {
     const updatedproperty = req.locals;
     const { user } = req;
-    const owner = user._id;
-    if (updatedproperty.owner === owner) {
-      updateProperty(updatedproperty)
-        .then(() => {
-          res.status(200).json({ success: true, message: 'Property updated', updatedproperty });
-        })
-        .catch((error) => next(error));
-    } else {
-      res.status(403).json({ success: true, message: 'Forbidden access' });
-    }
-  },
-  getAll(req, res, next) {
-    const { user } = req;
-    const owner = user._id;
-    getOwnedProperties(owner)
-      .then((properties) => {
-        if (properties.length > 0) {
-          res.status(200).json({ success: true, properties });
-        } else {
-          res.status(404).json({ success: true, message: 'No properties available' });
-        }
-      })
-      .catch((error) => next(error));
-  },
-  get(req, res, next) {
-    const { refNo } = req.params;
-    const { user } = req;
-    const owner = user._id;
-    getPropertyByRefNo(refNo)
-      .then((property) => {
-        // eslint-disable-next-line eqeqeq
-        if (property.owner == owner) {
-          if (property == null) {
-            res.status(404).json({ success: true, message: 'Property does not exist' });
-          } else {
-            res.status(200).json({ success: true, property });
-          }
-        } else {
-          res.status(403).json({ success: true, message: 'Forbidden access' });
-        }
+    updateProperty({ ...updatedproperty, updatedBy: user._id })
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Property updated' });
       })
       .catch((error) => next(error));
   },
   delete(req, res, next) {
-    const { refNo } = req.params;
-    const { user } = req;
-    const owner = user._id;
-
-    getPropertyByRefNo(refNo)
-      .then((property) => {
-        // eslint-disable-next-line eqeqeq
-        if (property.owner == owner) {
-          deleteProperty(refNo)
-            .then(() => {
-              res.status(200).json({ success: true, message: 'Property deleted' });
-            })
-            .catch((error) => next(error));
+    const { id } = req.params;
+    deleteProperty(id)
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Property deleted' });
+      })
+      .catch((error) => next(error));
+  },
+  getMultiple(req, res, next) {
+    getAllProperties()
+      .then((properties) => {
+        if (properties.length > 0) {
+          res.status(httpStatus.OK).json({ success: true, properties });
         } else {
-          res.status(403).json({ success: true, message: 'Forbidden access' });
+          res
+            .status(httpStatus.NOT_FOUND)
+            .json({ success: false, message: 'No properties available' });
+        }
+      })
+      .catch((error) => next(error));
+  },
+  getOne(req, res, next) {
+    getPropertyById(req.params.id)
+      .then((property) => {
+        if (property.length < 1) {
+          res
+            .status(httpStatus.NOT_FOUND)
+            .json({ success: false, message: 'Property does not exist' });
+        } else {
+          res.status(httpStatus.OK).json({ success: true, property });
         }
       })
       .catch((error) => next(error));
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -1,0 +1,88 @@
+import {
+  addProperty,
+  updateProperty,
+  deleteProperty,
+  getPropertyByRefNo,
+  getOwnedProperties,
+} from '../services/property.service';
+
+const PropertyController = {
+  add(req, res, next) {
+    const property = req.locals;
+    const { user } = req;
+    property.owner = user._id;
+    addProperty(property)
+      .then(() => {
+        res.status(201).json({ success: true, message: 'Property added' });
+      })
+      .catch((error) => next(error));
+  },
+  update(req, res, next) {
+    const updatedproperty = req.locals;
+    const { user } = req;
+    const owner = user._id;
+    if (updatedproperty.owner === owner) {
+      updateProperty(updatedproperty)
+        .then(() => {
+          res.status(200).json({ success: true, message: 'Property updated', updatedproperty });
+        })
+        .catch((error) => next(error));
+    } else {
+      res.status(403).json({ success: true, message: 'Forbidden access' });
+    }
+  },
+  getAll(req, res, next) {
+    const { user } = req;
+    const owner = user._id;
+    getOwnedProperties(owner)
+      .then((properties) => {
+        if (properties.length > 0) {
+          res.status(200).json({ success: true, properties });
+        } else {
+          res.status(404).json({ success: true, message: 'No properties available' });
+        }
+      })
+      .catch((error) => next(error));
+  },
+  get(req, res, next) {
+    const { refNo } = req.params;
+    const { user } = req;
+    const owner = user._id;
+    getPropertyByRefNo(refNo)
+      .then((property) => {
+        // eslint-disable-next-line eqeqeq
+        if (property.owner == owner) {
+          if (property == null) {
+            res.status(404).json({ success: true, message: 'Property does not exist' });
+          } else {
+            res.status(200).json({ success: true, property });
+          }
+        } else {
+          res.status(403).json({ success: true, message: 'Forbidden access' });
+        }
+      })
+      .catch((error) => next(error));
+  },
+  delete(req, res, next) {
+    const { refNo } = req.params;
+    const { user } = req;
+    const owner = user._id;
+
+    getPropertyByRefNo(refNo)
+      .then((property) => {
+        // eslint-disable-next-line eqeqeq
+        if (property.owner == owner) {
+          deleteProperty(refNo)
+            .then(() => {
+              res.status(200).json({ success: true, message: 'Property deleted' });
+            })
+            .catch((error) => next(error));
+        } else {
+          res.status(403).json({ success: true, message: 'Forbidden access' });
+        }
+      })
+      .catch((error) => next(error));
+  },
+};
+
+export default PropertyController;
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -9,10 +9,10 @@ import httpStatus from '../helpers/httpStatus';
 
 const PropertyController = {
   add(req, res, next) {
-    const property = req.locals;
+    const newProperty = req.locals;
     const { user } = req;
-    addProperty({ ...property, addedBy: user._id, updatedBy: user._id })
-      .then(() => {
+    addProperty({ ...newProperty, addedBy: user._id, updatedBy: user._id })
+      .then((property) => {
         res.status(httpStatus.CREATED).json({ success: true, message: 'Property added', property });
       })
       .catch((error) => next(error));
--- a/server/helpers/middleware.js
+++ b/server/helpers/middleware.js
@@ -34,7 +34,7 @@ export const authenticate = async (req, res, next) => {
     } else {
       return res.status(httpStatus.NOT_FOUND).json({
         success: false,
-        message: 'User not found',
+        message: 'Invalid token',
       });
     }
   } catch (error) {
--- a/server/helpers/middleware.js
+++ b/server/helpers/middleware.js
@@ -1,4 +1,5 @@
 import httpStatus from './httpStatus';
+import userRole from './userRole';
 import { decodeToken, getUserById } from '../services/user.service';
 
 export const schemaValidation = (schema) => {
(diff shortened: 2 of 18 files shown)


server/controllers/property.controllers.js
```diff
@@ -0,0 +1,88 @@
...
+  deleteProperty,
+  getPropertyByRefNo,
+  getOwnedProperties,
+} from '../services/property.service';
+
+const PropertyController = {
+  add(req, res, next) {
+    const property = req.locals;
+    const { user } = req;
+    property.owner = user._id;
```
andela-hpopoola, line 13: Can we try to merge this with the method invocation below? It is good practises to avoid object mutation.

server/controllers/property.controllers.js
```diff
@@ -0,0 +1,88 @@
...
+    addProperty(property)
```
andela-hpopoola, line 14:
```suggestion
    addProperty({...property, owner: user._id })
```

server/schemas/property.schema.js
```diff
@@ -0,0 +1,14 @@
...
+export const propertySchema = Joi.object({
+  refNo: Joi.string().label('Reference number').required(),
+  category: Joi.string().label('Product category').required(),
+  type: Joi.string().label('Product type').required(),
+  price: Joi.number().label('Product price').required(),
+  area: Joi.string().label('Product area').required(),
+  state: Joi.string().label('Product state').required(),
+});
+
+export const propertyUpdateSchema = Joi.object({
```
andela-hpopoola, line 12: Not sure why you are verifying only the email address for an updated property. I think we need to verify all the input submitted by the admin to be sure that they are valid.

server/controllers/property.controllers.js
```diff
@@ -0,0 +1,87 @@
...
+      .catch((error) => next(error));
+  },
+  delete(req, res, next) {
+    const { refNo } = req.params;
+    const { user } = req;
+    const owner = user._id;
+
+    getPropertyByRefNo(refNo)
+      .then((property) => {
+        // eslint-disable-next-line eqeqeq
```
andela-hpopoola, line 72: Even though the code below will likely be removed, I see no reason to disable eslint here. If the code below doesn't give you the expected result in this case, then it means you are doing something wrong (for the values not be equal) and would need to check your code again.

server/models/user.model.js
```diff
@@ -54,6 +54,13 @@ const UserSchema = new mongoose.Schema(
       select: false,
     },
     phone: String,
+    role: {
+      type: Number,
+      default: 1,
+    },
+    assignedProperties: {
+      type: [],
```
andela-hpopoola, line 62: This should be a list of ObjectIds. Something like this (You might need to confirm from mongoose documentation):
```suggestion
      type: [{ type : ObjectId, ref: 'Property' }],
```

server/controllers/property.controllers.js
tobieeniafe: the values are equal, but the data types differ

andela-hpopoola: That is exactly my point sir, the values and the data type should be the same. If they are not the same, you are doing something wrong. From the code, it seems one is an ObjectId and the other is a Number. Both of them should be ObjectId in this case.

server/controllers/property.controllers.js
```diff
@@ -0,0 +1,65 @@
...
+} from '../services/property.service';
+import httpStatus from '../helpers/httpStatus';
+
+const PropertyController = {
+  add(req, res, next) {
+    const property = req.locals;
+    const { user } = req;
+    addProperty({ ...property, addedBy: user._id, updatedBy: user._id })
+      .then(() => {
+        res.status(httpStatus.CREATED).json({ success: true, message: 'Property added' });
```
andela-hpopoola, line 16: Let's return the newly added property to the response.

server/helpers/middleware.js
```diff
@@ -43,6 +44,18 @@ export const authenticate = async (req, res, next) => {
       error,
     });
   }
+  return null;
+};
 
+export const isAdmin = async (req, res, next) => {
+  const { user } = req;
+  if (user.role === userRole.ADMIN) {
```
andela-hpopoola, line 52: Lets verify if user is valid here, so that we can get an error if used without authenticating (i.e. if user is null)
```suggestion
  if (user && user.role === userRole.ADMIN) {
```

server/helpers/middleware.js
```diff
@@ -43,6 +44,18 @@ export const authenticate = async (req, res, next) => {
...
+    next();
+  } else {
+    return res.status(httpStatus.FORBIDDEN).json({
+      success: false,
+      message: 'Action can only be performed by an Admin',
```
andela-hpopoola, line 57:
```suggestion
      message: 'You are not permitted to perform this action',
```

server/models/property.model.js
```diff
@@ -0,0 +1,126 @@
...
+    },
+    toilets: {
+      type: Number,
+      required: true,
+    },
+    description: {
+      type: String,
+      required: true,
+    },
+    floorPlans: {
```
andela-hpopoola, line 100: I think this will be a link to an uploaded floor plan document from the mockup. Let's make it a string for the main time.

server/models/property.model.js
```diff
@@ -0,0 +1,126 @@
...
+      type: Boolean,
+      required: true,
```
andela-hpopoola, line 102: I think this is better as optional

server/models/property.model.js
```diff
@@ -0,0 +1,126 @@
...
+    },
+    mapLocation: {
+      type: String,
+      required: true,
+    },
+    neighborhood: {
```
andela-hpopoola, line 108: From the Mockup, this should be an array of strings.

server/models/property.model.js
```diff
@@ -0,0 +1,126 @@
...
+      type: String,
+      required: true,
+    },
+    addedBy: {
+      type: String,
```
andela-hpopoola, line 113: This should be an ObjectId so that it can have the same data type as the user id.

server/models/property.model.js
```diff
@@ -0,0 +1,126 @@
...
+      required: true,
+    },
+    updatedBy: {
+      type: String,
```
andela-hpopoola, line 117: Same comment as above.

server/models/user.model.js
```diff
@@ -47,14 +47,38 @@ const UserSchema = new mongoose.Schema(
...
     },
     phone: String,
-    activated: { type: Boolean, default: false },
+    role: {
+      type: Number,
+      default: 1,
+    },
+    assignedProperties: [
+      {
+        propertyId: String,
```
andela-hpopoola, line 67: This should be an ObjectId imho

server/models/user.model.js
```diff
@@ -47,14 +47,38 @@ const UserSchema = new mongoose.Schema(
...
+        assignedBy: String,
```
andela-hpopoola, line 68: Also this, ObjectId

server/models/user.model.js
```diff
@@ -47,14 +47,38 @@ const UserSchema = new mongoose.Schema(
...
+        assignedDate: Date,
+        approved: {
+          type: Boolean,
+          default: false,
+        },
+        approvedBy: String,
```
andela-hpopoola, line 74: This too 😬

server/models/property.model.js
```diff
@@ -98,23 +100,22 @@ const PropertySchema = new mongoose.Schema(
...
     },
     mapLocation: {
       type: String,
       required: true,
     },
     neighborhood: {
-      type: String,
+      type: [String],
       required: true,
```
andela-hpopoola, line 109: Please let's make this optional, please?

server/models/property.model.js
```diff
@@ -98,23 +100,22 @@ const PropertySchema = new mongoose.Schema(
...
     },
     floorPlans: {
-      type: Boolean,
-      required: true,
+      type: String,
     },
     mapLocation: {
       type: String,
       required: true,
```
andela-hpopoola: Lets make this optional too?

server/models/property.model.js
```diff
@@ -0,0 +1,125 @@
...
+ *          - location
+ *          - price
+ *          - units
+ *          - houseType
+ *          - bedrooms
+ *          - toilets
+ *          - description
+ *          - floorPlans
+ *          - mapLocation
+ *          - neighborhood
```
andela-hpopoola, line 20: We need to add the `property image` and `gallery` to the property schema. I mentioned you(your highrachy mail) on the spreadsheet document when I noticed that it was missing.

server/schemas/property.schema.js
```diff
@@ -0,0 +1,30 @@
...
+  name: Joi.string().label('Property name').required(),
+  location: Joi.string().label('Property location').required(),
+  price: Joi.number().label('Property price').required(),
+  units: Joi.number().label('Property units').required(),
+  houseType: Joi.string().label('Property type').required(),
+  bedrooms: Joi.number().label('Bedroom number').required(),
+  toilets: Joi.number().label('Toilet number').required(),
+  description: Joi.string().label('Property description').required(),
+  floorPlans: Joi.boolean().label('Property floor plans').required(),
+  mapLocation: Joi.string().label('Map location').required(),
```
andela-hpopoola, line 13: This should test for an optional object

server/schemas/property.schema.js
```diff
@@ -0,0 +1,30 @@
...
+  neighborhood: Joi.string().label('Map location').required(),
+});
+
+export const updatePropertySchema = Joi.object({
```
andela-hpopoola, line 17: Instead of repeating the same values here as above, can you assign the same values into an object and spread it in the two objects to ensure that updates are made in one place.

server/schemas/property.schema.js
```diff
@@ -0,0 +1,30 @@
...
+  location: Joi.string().label('Property location').required(),
+  price: Joi.number().label('Property price').required(),
+  units: Joi.number().label('Property units').required(),
+  houseType: Joi.string().label('Property type').required(),
+  bedrooms: Joi.number().label('Bedroom number').required(),
+  toilets: Joi.number().label('Toilet number').required(),
+  description: Joi.string().label('Property description').required(),
+  floorPlans: Joi.boolean().label('Property floor plans').required(),
+  mapLocation: Joi.string().label('Map location').required(),
+  neighborhood: Joi.string().label('Map location').required(),
```
andela-hpopoola, line 14: This should test for an optional array

server/schemas/property.schema.js
```diff
@@ -9,22 +14,16 @@ export const addPropertySchema = Joi.object({
...
   toilets: Joi.number().label('Toilet number').required(),
   description: Joi.string().label('Property description').required(),
-  floorPlans: Joi.boolean().label('Property floor plans').required(),
-  mapLocation: Joi.string().label('Map location').required(),
-  neighborhood: Joi.string().label('Map location').required(),
-});
+  floorPlans: Joi.string().label('Property floor plans'),
+  mapLocation,
+  neighborhood: Joi.array().label('Property neighborhood'),
+  mainImage: Joi.string().label('Property neighborhood').required(),
```
andela-hpopoola, line 20: Let's make the main image and gallery optional, please.

server/models/property.model.js
```diff
@@ -0,0 +1,136 @@
...
+    description: {
+      type: String,
+      required: true,
+    },
+    floorPlans: {
+      type: String,
+    },
+    mapLocation: {
+      longitude: {
+        type: String,
```
andela-hpopoola, line 107: Can we make this a number instead of a string so that it tallies with the example you gave in the API docs?

server/controllers/property.controllers.js
```diff
@@ -0,0 +1,65 @@
...
+} from '../services/property.service';
+import httpStatus from '../helpers/httpStatus';
+
+const PropertyController = {
+  add(req, res, next) {
+    const property = req.locals;
+    const { user } = req;
+    addProperty({ ...property, addedBy: user._id, updatedBy: user._id })
+      .then(() => {
+        res.status(httpStatus.CREATED).json({ success: true, message: 'Property added', property });
```
andela-hpopoola, line 16: Can you return the property returned by the database instead of the one submitted by the user. I think it is preferable for the returned property to contain an id (even if we want to omit other things).

server/schemas/property.schema.js
```diff
@@ -0,0 +1,29 @@
...
+  floorPlans: Joi.string().label('Property floor plans').optional(),
+  mapLocation,
+  neighborhood: Joi.array().label('Property neighborhood').optional(),
+  mainImage: Joi.string().label('Property neighborhood').optional(),
+  gallery: Joi.array().label('Property neighborhood').optional(),
+};
+
+export const addPropertySchema = Joi.object({ ...product });
+
+export const updatePropertySchema = Joi.object({
```
andela-hpopoola, line 26: I think I am wrong here. We should be able to update a property without stating all the values in the property. We should be able to update only one section (maybe the units with the id alone without stating other properties).

Lets readd the properties and make them optional instead of required so that the schema validation will allow us to add a subset of the property instead of adding more.

server/schemas/property.schema.js
```diff
@@ -25,5 +25,17 @@ export const addPropertySchema = Joi.object({ ...product });
 
 export const updatePropertySchema = Joi.object({
   id: Joi.string().label('Property id').required(),
-  ...product,
+  units: Joi.number().label('Property units').required(),
```
andela-hpopoola, line 28: Not sure why this is still `required`. I thought we wanted to make them optional?

test/services/property.service.test.js
```diff
@@ -0,0 +1,63 @@
+import mongoose from 'mongoose';
+import { expect, useDatabase } from '../config';
+import {
+  getPropertyById,
+  addProperty,
+  //   updateProperty,
```
andela-hpopoola, line 6: I am not sure why you commented this out, we need to write tests to check for property updates, delete and getAllProperties.

test/factories/property.factory.js
```diff
@@ -0,0 +1,19 @@
+import { Factory } from 'rosie';
+
+export default new Factory()
```
andela-hpopoola, line 3: Can you update the factory so that it contains the most updated version of property schema. I noticed that mainImage and gallery fields are missing.

test/controllers/property.controller.test.js
```diff
@@ -0,0 +1,296 @@
...
+
+describe('Update Property', () => {
+  const id = mongoose.Types.ObjectId();
+  const property = PropertyFactory.build({ _id: id, addedBy: _id, updatedBy: _id });
+
+  beforeEach(async () => {
+    await addProperty(property);
+  });
+
+  const newProperty = {
```
andela-hpopoola, line 65: This might look like I am asking for too much, but to ensure that our test works in the best way for us, we also need to test for all that we can update all the fields one by one with our test (not a part of it). Updating just a part of it works to test that the update controller works, however, we also need to test that each field will not cause an error / bug and would properly be updated when they are done by the user.

Can you add the other fields in this object and test that all of them are properly updated?

test/services/property.service.test.js
```diff
@@ -0,0 +1,157 @@
+import mongoose from 'mongoose';
+import { expect, sinon, useDatabase } from '../config';
+import {
+  getPropertyById,
+  addProperty,
+  updateProperty,
+  deleteProperty,
+  // getAllProperties,
```
andela-hpopoola, line 8: Not sure why you commented this out? Can we add a basic test for it, please?

tobieeniafe:
> Thank you for working on this @tobieeniafe. We are also through with this, thank you for your patience and endurance (I know adding tests can sometimes but frustrating).
> 
> I added a few more comments on updating a few tests. Also, I didn't see any implementation for this as stated in one of my comments above.
> 
> > getAllProperties endpoint needs to be properly fleshed out. The current implementation is not usable and it will be better if we can split it to the following:
> > a. getAllUserProperties - This should get all the properties owned by a user. Each property should contain the admin details.
> > b. getAllAdminAddedProperties - Get all the properties added by an admin. Each property should have the user information that has also been assigned to the property.
> > c. getAllProperties: Get all properties in the database with the assigned user and admin details.
> 
> I am happy if you want to implement them on another Pull Request. Just inform me and I will be glad to create a new story for it. Thank you and Nice Progress.
> 
> 

Let's add this to another PR please

server/schemas/property.schema.js
```diff
@@ -0,0 +1,41 @@
...
+  price: Joi.number().label('Property price').required(),
+  units: Joi.number().label('Property units').required(),
+  houseType: Joi.string().label('Property type').required(),
+  bedrooms: Joi.number().label('Bedroom number').required(),
+  toilets: Joi.number().label('Toilet number').required(),
+  description: Joi.string().label('Property description').required(),
+  floorPlans: Joi.string().label('Property floor plans').optional(),
+  mapLocation,
+  neighborhood: Joi.array().label('Property neighborhood').optional(),
+  mainImage: Joi.string().label('Property mainImage').optional(),
```
andela-hpopoola, line 20:
```suggestion
  mainImage: Joi.string().label('Property main image').optional(),
```

server/schemas/property.schema.js
```diff
@@ -0,0 +1,41 @@
...
+  location: Joi.string().label('Property location').optional(),
+  price: Joi.number().label('Property price').optional(),
+  houseType: Joi.string().label('Property type').optional(),
+  bedrooms: Joi.number().label('Bedroom number').optional(),
+  toilets: Joi.number().label('Toilet number').optional(),
+  description: Joi.string().label('Property description').optional(),
+  floorPlans: Joi.string().label('Property floor plans').optional(),
+  mapLocation,
+  neighborhood: Joi.array().label('Property neighborhood').optional(),
+  mainImage: Joi.string().label('Property mainImage').optional(),
```
andela-hpopoola, line 39:
```suggestion
  mainImage: Joi.string().label('Property main image').optional(),
```

test/controllers/property.controller.test.js
```diff
@@ -51,28 +90,235 @@ describe('Add Property Route', () => {
...
+        const property = PropertyFactory.build({ mainImage: '' });
+        request()
+          .post('/api/v1/property/add')
+          .set('authorization', adminToken)
+          .send(property)
+          .end((err, res) => {
+            expect(res).to.have.status(412);
+            expect(res.body.success).to.be.eql(false);
+            expect(res.body.message).to.be.eql('Validation Error');
+            expect(res.body.error).to.be.eql('"Property mainImage" is not allowed to be empty');
```
andela-hpopoola, line 248: Based on the above nits:
```suggestion
            expect(res.body.error).to.be.eql('"Property main image" is not allowed to be empty');
```

test/controllers/property.controller.test.js
```diff
@@ -51,28 +90,235 @@ describe('Add Property Route', () => {
...
+            done();
+          });
+      });
+    });
+    context('when gallery is empty', () => {
```
andela-hpopoola, line 253: The gallery is supposed to be an array, so we need to add other tests to test that we can add an empty array. The same should apply for other optional fields.

test/controllers/property.controller.test.js
```diff
@@ -230,12 +739,29 @@ describe('Get all properties', () => {
...
     });
 
+    context('when user token is used', () => {
+      beforeEach(async () => {
+        await User.findByIdAndDelete(_id);
+      });
+      it('returns token error', (done) => {
+        request()
+          .get('/api/v1/property/all')
+          .set('authorization', adminToken)
```
andela-hpopoola, line 749: The context states that `when user token is used`, however, you used the adminToken here.

test/controllers/property.controller.test.js
```diff
@@ -268,11 +794,28 @@ describe('Delete property', () => {
...
   });
 
+  context('when user token is used', () => {
+    beforeEach(async () => {
+      await User.findByIdAndDelete(_id);
+    });
+    it('returns token error', (done) => {
+      request()
+        .delete(`/api/v1/property/delete/${id}`)
+        .set('authorization', adminToken)
```
andela-hpopoola, line 804: Same comment as above

test/services/property.service.test.js
```diff
@@ -154,4 +154,17 @@ describe('Property Service', () => {
...
+  describe('#getAllProperties', () => {
+    const _id = mongoose.Types.ObjectId();
+
+    before(async () => {
+      await Property.create(PropertyFactory.build({ _id, addedBy: _id, updatedBy: _id }));
+    });
+
+    it('returns a valid property by Id', async () => {
+      const property = await getAllProperties();
+      expect(property).to.be.an('array');
```
andela-hpopoola, line 167: Though we are fixing this in another PR, however, testing that the response is an array is not a valid way to test that all the properties are returned. Here are my thought process:

1. Find a way to add 2 properties into the database for this test
2. Get all the properties and test that it returns all the properties
**(Optional tests but nice to have as they gives more confidence)**
3. Test that the returned properties are the same as the added ones
4. Add one more property
5. Verify that getAllProperties returns 3 properties

andela-hpopoola:
Thank you for working on this. I found a few nits but they are nothing too serious. Feel free to merge this PR after you resolve them. Thank you and Nice work.

PR #23, opened 2020-07-13
#173766649 edit user endpoint

tobieeniafe:
#### What does this PR do?
Add edit user endpoint for dashboard

Create endpoint to edit a user

`PUT /api/v1/user/update`
Sample data
`{
  "firstName": "User Firstname",
  "lastName": "User Lastname",
  "phone": "08012345678"
}`

All requests should be sent with token as header `Authorization`

#173766649

### DIFF
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -4,6 +4,7 @@ import {
   activateUser,
   forgotPasswordToken,
   resetPasswordViaToken,
+  updateUser,
 } from '../services/user.service';
 import { sendMail } from '../services/mailer.service';
 import EMAIL_CONTENT from '../../mailer';
@@ -82,6 +83,16 @@ const UserController = {
     return res
       .status(httpStatus.OK)
       .json({ success: true, message: 'Your information has been successfully retrieved', user });
+  },
+
+  update(req, res, next) {
+    const updateduser = req.locals;
+    const { user } = req;
+    updateUser({ ...updateduser, id: user._id })
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'User updated' });
+      })
+      .catch((error) => next(error));
   },
 };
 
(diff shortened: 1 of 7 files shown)


server/controllers/user.controllers.js
```diff
@@ -83,6 +84,16 @@ const UserController = {
       .status(httpStatus.OK)
       .json({ success: true, message: 'Your information has been successfully retrieved', user });
   },
+
+  update(req, res, next) {
+    const updateduser = req.locals;
+    const { user } = req;
+    updateUser({ ...updateduser, id: user._id })
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'User updated' });
```
andela-hpopoola, line 93: Can you return the updated user information here?

test/controllers/user.controller.test.js
```diff
@@ -601,3 +601,75 @@ describe('Current User', () => {
...
+    it('returns a updated user', (done) => {
+      request()
+        .put('/api/v1/user/update')
+        .set('authorization', token)
+        .send(newUser)
+        .end((err, res) => {
+          expect(res).to.have.status(200);
+          expect(res.body.success).to.be.eql(true);
+          expect(res.body).to.have.property('updateduser');
+          done();
```
andela-hpopoola, line 629: You need to add tests to verify the firstName, lastName and phone number has truly been updated.

test/controllers/user.controller.test.js
```diff
@@ -601,3 +601,75 @@ describe('Current User', () => {
...
+
+  context('without token', () => {
+    it('returns error', (done) => {
+      request()
+        .put('/api/v1/user/update')
+        .send(newUser)
+        .end((err, res) => {
+          expect(res).to.have.status(403);
+          expect(res.body.success).to.be.eql(false);
+          done();
```
andela-hpopoola, line 642: Can add test for the message or error that is returned to be sure that the right message is returned.

test/controllers/user.controller.test.js
tobieeniafe: the scenario to making sure the values were updated was tested for in the `user.service.test.js`. Line `380`. If I understand you correctly.

PR #24, opened 2020-07-16
#173766652 add referral code

tobieeniafe:
#### What does this PR do?
Add referral code

Add referral code to user schema and automatically generate code when user registers

`POST /api/v1/user/register`

#173766652

### DIFF
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -47,7 +47,11 @@ const UserSchema = new mongoose.Schema(
       type: String,
       required: true,
     },
-    email: { type: String, unique: true, required: true },
+    email: {
+      type: String,
+      unique: true,
+      required: true,
+    },
     password: {
       type: String,
       required: true,
@@ -56,8 +60,12 @@ const UserSchema = new mongoose.Schema(
     phone: String,
     referralCode: {
       type: String,
+      unique: true,
     },
-    activated: { type: Boolean, default: false },
+    activated: {
+      type: Boolean,
+      default: false,
+    },
     activationDate: Date,
   },
   { timestamps: true },
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -37,6 +37,7 @@ import mongoose from 'mongoose';
  *           phone: 08012345678
  */
 
+const { ObjectId } = mongoose.Schema.Types;
 const UserSchema = new mongoose.Schema(
   {
     firstName: {
@@ -62,6 +63,23 @@ const UserSchema = new mongoose.Schema(
       type: String,
       unique: true,
     },
+    role: {
+      type: Number,
+      default: 1,
+    },
+    assignedProperties: [
+      {
+        propertyId: ObjectId,
+        assignedBy: ObjectId,
+        assignedDate: Date,
+        approved: {
+          type: Boolean,
+          default: false,
+        },
+        approvedBy: ObjectId,
+        approvedDate: Date,
+      },
+    ],
     activated: {
       type: Boolean,
       default: false,
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -54,6 +54,9 @@ const UserSchema = new mongoose.Schema(
       select: false,
     },
     phone: String,
+    referralCode: {
+      type: String,
+    },
     activated: { type: Boolean, default: false },
     activationDate: Date,
   },
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -32,8 +32,46 @@ export const comparePassword = async (password, hashedPassword) => {
   }
 };
 
+export const getUserByReferralCode = async (referralCode, fields = null) =>
+  User.findOne({ referralCode }).select(fields);
+
+export const generateReferralCode = async (firstName) => {
+  const name = firstName.substring(0, 3).toLowerCase();
+  let code = name;
+  let charsToAdd = 4;
+
+  if (firstName.length < 3) {
+    charsToAdd = 7 - firstName.length;
+  }
+
+  function generateCode(pre) {
+    const possible = '0123456789';
+    let fullCode = pre;
+    for (let i = 0; i < charsToAdd; i += 1) {
+      fullCode += possible.charAt(Math.floor(Math.random() * possible.length));
+    }
+    return fullCode;
+  }
+
+  code = generateCode(code);
+
+  const referralCodeIsUsed = await getUserByReferralCode(code).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (referralCodeIsUsed) {
+    code = name;
+    code = generateCode(code);
+  }
+  return code;
+};
+
 export const addUser = async (user) => {
   const existingUser = await getUserByEmail(user.email).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  const referralCode = await generateReferralCode(user.firstName).catch((error) => {
     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
   });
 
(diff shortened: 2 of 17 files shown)


server/services/user.service.js
```diff
@@ -32,18 +32,56 @@ export const comparePassword = async (password, hashedPassword) => {
...
+export const generateReferralCode = async (firstName) => {
+  const name = firstName.substring(0, 3).toLowerCase();
+  let code = name;
+  let charsToAdd = 4;
+
+  if (firstName.length < 3) {
+    charsToAdd = 7 - firstName.length;
+  }
+
+  function generateCode(pre) {
```
andela-hpopoola, line 47: Can you place this function outside the generateReferralCode function so that you can also add tests for it?

server/services/user.service.js
```diff
@@ -32,18 +32,56 @@ export const comparePassword = async (password, hashedPassword) => {
   }
 };
 
+export const getUserByReferralCode = async (referralCode, fields = null) =>
+  User.findOne({ referralCode }).select(fields);
+
+export const generateReferralCode = async (firstName) => {
+  const name = firstName.substring(0, 3).toLowerCase();
+  let code = name;
+  let charsToAdd = 4;
```
andela-hpopoola, line 41: This line and subsequent lines can be fixed into a single line
```suggestion
  const charsToAdd = firstName.length < 3 ? 7 - firstName.length : 4;
```

server/services/user.service.js
```diff
@@ -32,18 +32,56 @@ export const comparePassword = async (password, hashedPassword) => {
   }
 };
 
+export const getUserByReferralCode = async (referralCode, fields = null) =>
+  User.findOne({ referralCode }).select(fields);
+
+export const generateReferralCode = async (firstName) => {
+  const name = firstName.substring(0, 3).toLowerCase();
+  let code = name;
```
andela-hpopoola, line 40: Not sure why you are reassigning code to name? You could as well use it as `name` throughout the code (as the reassignment has not meaningful purpose)

server/services/user.service.js
andela-hpopoola, line 47: What does pre stand for in this function? Can we choose a more descriptive name like `referralCodePrefix` or any name that describes the purpose of the variable please?

server/services/user.service.js
```diff
@@ -32,18 +32,56 @@ export const comparePassword = async (password, hashedPassword) => {
...
+  function generateCode(pre) {
+    const possible = '0123456789';
+    let fullCode = pre;
+    for (let i = 0; i < charsToAdd; i += 1) {
+      fullCode += possible.charAt(Math.floor(Math.random() * possible.length));
+    }
+    return fullCode;
+  }
+
+  code = generateCode(code);
```
andela-hpopoola, line 56: This would be better as something like this depending on how you resolved the previous comments I left.
```suggestion
  const referralCode = generateCode(name);
```

server/services/user.service.js
```diff
@@ -32,18 +32,56 @@ export const comparePassword = async (password, hashedPassword) => {
...
+
+  const referralCodeIsUsed = await getUserByReferralCode(code).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (referralCodeIsUsed) {
+    code = name;
+    code = generateCode(code);
```
andela-hpopoola, line 64: The similar code above should be removed.

server/services/user.service.js
```diff
@@ -32,18 +32,56 @@ export const comparePassword = async (password, hashedPassword) => {
...
+    return fullCode;
+  }
+
+  code = generateCode(code);
+
+  const referralCodeIsUsed = await getUserByReferralCode(code).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (referralCodeIsUsed) {
```
andela-hpopoola, line 62: I think the logic here is wrong. This will only check if the referral code has been used only once. If the referral code has also been used the second time, the code will not catch it. 

The best way to resolve this is to use a loop to determine if the referral code is available or not. Here are my thought processes:
1. Create a variable that assumes the referralCode has been used. In this case, something like this will suffice.
```js
let referralCode = true
```
2. Use a while loop to check if the referral code is available from the database. If it is available, assign referralCodeIsUsed to false else the loop should continue running.
```js
while (referralCodeIsUsed) {
// generate referralCode
// const referralCode = generateReferalCode(name) ... depending on your code
// do your database call to check if the referalCodeIsUsed here
if (!databaseReturnsAValue) { // this is also a pseudocall based on the function above. it could as well be the returned user
referralCodeIsUsed = false;
} 
```

3. Return the generated referralCode

I hope this helps.

test/services/user.service.test.js
```diff
@@ -379,23 +381,75 @@ describe('User Service', () => {
...
+          expect(err.statusCode).to.eql(500);
+          expect(err.error).to.be.an('Error');
+          expect(err.message).to.be.eql('Internal Server Error');
+        }
+        User.findById.restore();
+      });
+    });
+  });
+
+  describe('#getUserByReferralCode', () => {
```
andela-hpopoola, line 422: You will need to add the following tests to ensure that the getUserByReferralCode works as expected:
1. Add a user with firstName less than 3 characters. You need to test that it the returned referral code is 7 characters and that the first letters are present in the generated referral code.
2. Add a user with firstName more than 3 characters, You need to test that it the returned referral code is 7 characters and that the first letters are present in the generated referral code and that it is the first 3 letters.

Please let me know if you need more explanation on this.

server/services/user.service.js
```diff
@@ -32,18 +32,54 @@ export const comparePassword = async (password, hashedPassword) => {
...
+  for (let i = 0; i < charsToAdd; i += 1) {
+    referralCode += possible.charAt(Math.floor(Math.random() * possible.length));
+  }
+  return referralCode;
+};
+
+export const generateReferralCode = async (firstName) => {
+  let referralCode = generateCode(firstName);
+
+  let referralCodeIsUsed = await getUserByReferralCode(referralCode).catch((error) => {
```
andela-hpopoola, line 53: This is no longer needed since you are calling the same code on line 59. The best implementation is to assign referralCodeIsUsed as true and make all calls in the loop.

server/services/user.service.js
```diff
@@ -32,18 +32,54 @@ export const comparePassword = async (password, hashedPassword) => {
...
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  while (referralCodeIsUsed) {
+    referralCode = generateCode(firstName);
+    const invalidReferralCode = getUserByReferralCode(referralCode);
```
andela-hpopoola, line 59: **Code Smell** : `getUserByReferralCode` will return a promise, you will need an async / await to get the value. Try using console.log to check the value return by `invalidReferralCode` to understand how to implement this better.

test/services/user.service.test.js
```diff
@@ -452,4 +453,30 @@ describe('User Service', () => {
       });
     });
   });
+
+  describe('#generateCode', () => {
+    context('when firstname is longer than 2 characters', () => {
+      it('returns 6 digit code starting with first two letters of name', async () => {
```
andela-hpopoola, line 459: Can you remove the `async` used before the function call since you didn't use any await in the test?

test/services/user.service.test.js
```diff
@@ -379,23 +382,101 @@ describe('User Service', () => {
...
-        firstName: 'Updated firstname',
-        lastName: 'Updated lastname',
-        phone: '08012345678',
-      };
-      const updatedUser = updateUser(updatedDetails);
-      const user = getUserById(updatedDetails.id);
-      expect(user.firstName).to.eql(updatedUser.firstName);
-      expect(user.lastName).to.eql(updatedUser.lastName);
-      expect(user.phone).to.eql(updatedUser.phone);
+    context('when generateReferralCode fails', () => {
```
andela-hpopoola, line 397: Can you update the code so that when the referral generation fails, the user account is not created? Also, this should also test that the user is not created when the referral generation fails.

server/models/user.model.js
```diff
@@ -54,6 +54,9 @@ const UserSchema = new mongoose.Schema(
       select: false,
     },
     phone: String,
+    referralCode: {
```
andela-hpopoola, line 57: Please let's make this required and unique. I believe the `referralCode` is as important as the email address.

PR #26, opened 2020-07-28
get all properties endpoint

tobieeniafe:
#### What does this PR do?
Flesh out get all properties endpoint

Break down get properties to meet specific API needs

Properties owned by user `GET /api/v1/user/my-properties`
Properties properties added by an admin(ID) `GET /api/v1/property/added-by/:id`


server/controllers/property.controllers.js
```diff
@@ -34,28 +37,32 @@ const PropertyController = {
...
       .then((property) => {
-        if (!property) {
-          res
-            .status(httpStatus.NOT_FOUND)
-            .json({ success: false, message: 'Property does not exist' });
-        } else {
+        if (property.length > 0) {
           res.status(httpStatus.OK).json({ success: true, property });
+        } else {
+          res.status(httpStatus.NOT_FOUND).json({ success: false, message: 'No property found' });
```
andela-hpopoola, line 65: Can you update this to the new format of returning `success: true` if there is no property in the database.

server/helpers/middleware.js
```diff
@@ -59,3 +60,18 @@ export const isAdmin = async (req, res, next) => {
   }
   return null;
 };
+
+export const paramIdIsValid = async (req, res, next) => {
```
andela-hpopoola, line 64: Can we update the name to `hasValidObjectId` so that it is more descriptive as we might still update the function to cater for id in other parts of the request object.

server/helpers/middleware.js
```diff
@@ -59,3 +60,18 @@ export const isAdmin = async (req, res, next) => {
...
+  const { id } = req.params;
+  const { ObjectId } = mongoose.Types;
+
+  if (ObjectId.isValid(id)) {
+    next();
+  } else {
+    return res.status(httpStatus.PRECONDITION_FAILED).json({
+      success: false,
+      message: 'ID should be a valid mongo object ID',
```
andela-hpopoola, line 73: We don't want to reveal the technology we are using to everyone consuming the API.

```suggestion
      message: 'Invalid Id supplied',
```

server/services/user.service.js
```diff
@@ -86,6 +90,27 @@ export const addUser = async (user) => {
   }
 };
 
+export const getUserInfo = async (userId) =>
```
andela-hpopoola, line 93: Nice thinking with this, however, I believe that a better way to do this is to update the  `getUserById` and `getUserByEmail` so that they can return the aggregated values. That way, we wouldn't need to make additional requests to get the user info, and we wouldn't

A more advanced way is to update the getUserInfo so that it combines the functionality of the two services (i.e. it can work with id and email but would differentiate the one to use from the function parameter). That way it can be called directly from `getUserById` and `getUserByEmail`. If you find this too complex for you, feel free to use the easiest way you know for now but ensure that we dont make double requests. Thank you.



**Note:** When implementing new features, a more important thing then the code working is to ensure that the code is optimal and we are maximizing resources.

server/services/user.service.js
```diff
@@ -102,10 +127,8 @@ export const loginUser = async (user) => {
       delete savedUser.password;
 
       if (savedUser.activated) {
-        return {
-          ...savedUser,
-          token,
-        };
+        const userInfo = await getUserInfo(savedUser._id);
```
andela-hpopoola, line 130: Making another user request here is an overkill, not needed....

server/controllers/user.controllers.js
```diff
@@ -78,11 +81,27 @@ const UserController = {
       .catch((error) => next(error));
   },
 
-  currentUser(req, res) {
+  currentUser(req, res, next) {
+    const id = req.user._id;
+    getUserInfo(id)
```
andela-hpopoola, line 86: Making another user request here is an overkill, not needed. I believe we can get everything needed from the single request in the authenticate middleware.

server/controllers/property.controllers.js
tobieeniafe: this is for when trying to get a single property by it's ID. I think its important we know the id is not found. Perhaps i should change the message to `Property not found`

andela-hpopoola: You are 100% correct. `Property not found` is a better suggestion 👍

server/services/user.service.js
tobieeniafe: the `getUserInfo` function can be made to return the aggregated values for either `getUserById` or `getUserByEmail`. But cannot replace them in the existing code workflow, as the application breaks when `getUserById ` or `getUserByEmail` are replaced with `getUserInfo`

server/controllers/user.controllers.js
tobieeniafe: this is the only way to return the aggrgate values

server/services/user.service.js
tobieeniafe: this is the only way to return the aggregate values of assigned properties

andela-hpopoola: I believe we should be able to make it one single request instead of two. However, instead of wasting time on this, please add a TODO to both places so that we can be reminded to fix them later.

PR #27, opened 2020-08-02
#173915835 visit property endpoint

tobieeniafe:
#### What does this PR do?
Creates property visitation endpoint

Create endpoint for a user to request to schedule a visit to a particular property

`POST /api/v1/visit/book`
`{
    "propertyId": "5f22f8aec790039da1242382",
    "visitorName": "John Doe",
    "visitorPhone": "08012345678"
}`

#173915835


server/controllers/visit.controllers.js
```diff
@@ -0,0 +1,23 @@
+import scheduleVisit from '../services/visit.service';
+// import { sendMail } from '../services/mailer.service';
```
andela-hpopoola, line 2: Kindly confirm from Nnamdi the email address we wish to send the visitation to.

server/controllers/visit.controllers.js
```diff
@@ -0,0 +1,23 @@
...
+// import EMAIL_CONTENT from '../../mailer';
```
andela-hpopoola, line 3: You will need to create the content in this template.

server/models/visit.model.js
```diff
@@ -0,0 +1,55 @@
...
+ *            type: string
+ *        example:
+ *           propertyId: 5f22f7f8c790039da1242381
+ *           visitorId: 5f22f7f8c790039da1223456
+ *           visitorName: John Doe
+ *           visitorPhone: 08012345678
+ */
+
+const { ObjectId } = mongoose.Schema.Types;
+const VisitSchema = new mongoose.Schema(
```
andela-hpopoola, line 31: It is preferable to use nouns as the model name. `Visitation` is preferred instead of visit.

server/models/visit.model.js
```diff
@@ -0,0 +1,55 @@
...
+  {
+    propertyId: {
+      type: ObjectId,
+      required: true,
+    },
+    visitorId: {
```
andela-hpopoola, line 37: I think userId is more descriptive in this context.
```suggestion
    userId: {
```

server/models/visit.model.js
```diff
@@ -0,0 +1,55 @@
...
+    },
+    visitorName: {
+      type: String,
+      required: true,
+    },
+    visitorPhone: {
+      type: String,
+      required: true,
+    },
+  },
```
andela-hpopoola, line 49: We need to add an optional Email too.

server/routes/visit.routes.js
```diff
@@ -0,0 +1,38 @@
+import express from 'express';
+import propertyVisitationSchema from '../schemas/visit.schema';
+import { schemaValidation, authenticate } from '../helpers/middleware';
+import VisitController from '../controllers/visit.controllers';
+
+const router = express.Router();
+
+/**
+ * @swagger
+ * /visit/book:
```
andela-hpopoola, line 10: To match the updated model name, schedule is preferable to be in sync with the design / UI context.
```suggestion
 * /visitation/schedule:
```

server/schemas/visit.schema.js
```diff
@@ -0,0 +1,9 @@
+import Joi from '@hapi/joi';
+
+const propertyVisitationSchema = Joi.object({
+  propertyId: Joi.string().label('Property id').required(),
+  visitorName: Joi.string().label('Visitor name').required(),
+  visitorPhone: Joi.string().label('Visitor phone').required(),
```
andela-hpopoola, line 6: Can you validate that the phone number is between 11 to 14 characters to mimic our mobile number. We want the mobile number to be reachable, so it makes sense to ensure that the user enters a valid information.

test/controllers/visit.controller.test.js
```diff
@@ -0,0 +1,131 @@
...
+import UserFactory from '../factories/user.factory';
+import { addUser } from '../../server/services/user.service';
+import { addProperty } from '../../server/services/property.service';
+
+useDatabase();
+
+let userToken;
+
+const userId = mongoose.Types.ObjectId();
+const systemUser = UserFactory.build({ _id: userId, role: 0, activated: true });
```
andela-hpopoola, line 15: Can you use admin instead of systemUser to be in sync with our user model.
```suggestion
const admin = UserFactory.build({ _id: userId, role: 0, activated: true });
```

test/controllers/visit.controller.test.js
```diff
@@ -0,0 +1,131 @@
...
+
+beforeEach(async () => {
+  userToken = await addUser(systemUser);
```
andela-hpopoola, line 18: This should be adminToken. It is better to properly name the different variables to ensure that they are not confusing when introducing new tests later.
```suggestion
  adminToken = await addUser(systemUser);
```

test/controllers/visit.controller.test.js
```diff
@@ -0,0 +1,131 @@
...
+
+  context('when user token is used', () => {
+    beforeEach(async () => {
+      await User.findByIdAndDelete(userId);
+    });
+    it('returns token error', (done) => {
+      const booking = VisitFactory.build({ propertyId: propId });
+      request()
+        .post('/api/v1/visit/book')
+        .set('authorization', userToken)
```
andela-hpopoola, line 54: This should be invalid Token. I believe the endpoint is for users, so a user token should be able to consume the endpoint with no error. So you need to create an invalidToken to test this.
```suggestion
        .set('authorization', invalidToken)
```

test/controllers/visit.controller.test.js
```diff
@@ -0,0 +1,131 @@
...
+          expect(res).to.have.status(404);
+          expect(res.body.success).to.be.eql(false);
+          expect(res.body.message).to.be.eql('Invalid token');
+          done();
+        });
+    });
+  });
+
+  context('when property does not exist', () => {
+    it('returns token error', (done) => {
```
andela-hpopoola, line 66: Likely typo:
```suggestion
    it('returns a property not found error', (done) => {
```

test/controllers/visit.controller.test.js
```diff
@@ -0,0 +1,131 @@
...
+        const property = VisitFactory.build({ visitorName: '' });
+        request()
+          .post('/api/v1/visit/book')
+          .set('authorization', userToken)
+          .send(property)
+          .end((err, res) => {
+            expect(res).to.have.status(412);
+            expect(res.body.success).to.be.eql(false);
+            expect(res.body.message).to.be.eql('Validation Error');
+            expect(res.body.error).to.be.eql('"Visitor name" is not allowed to be empty');
```
andela-hpopoola, line 109: "Visitor phone" is not a good response message to users (even though you used Visitor as the field name). Kindly update it and subsequent messages too.
```suggestion
            expect(res.body.error).to.be.eql('"Name" is not allowed to be empty');
```

test/controllers/visit.controller.test.js
tobieeniafe: no this endpoint is for when a used token (when the user is deleted or no longer exists). Would add the suggested one tho

server/controllers/visitation.controllers.js
```diff
@@ -1,15 +1,15 @@
...
-const VisitController = {
+const VisitationController = {
   book(req, res, next) {
     const booking = req.locals;
     const { user } = req;
-    scheduleVisit({ ...booking, visitorId: user._id })
+    scheduleVisitation({ ...booking, userId: user._id })
       .then((schedule) => {
-        // sendMail(EMAIL_CONTENT.ACTIVATE_YOUR_ACCOUNT, user, {
+        // sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, user, {
```
andela-hpopoola, line 12: For me, I would create a separate variable to build the contentBottom for the schedule mail, something like

```js
const contentBottom  = `
<strong> Name: </strong> ${schedule.visitorName}
// other variables here
...

```
Note: BALLERS_EMAIL would have been defined in a separate file it i.e constants.js

```suggestion
        // const contentBottom = ... 
        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: BALLERS_EMAIL }, { contentBottom })
```

This would be sent to mailtrap.io when you are testing on your local machine and would send to the email address in OUR_EMAIL when we have uploaded it.

test/controllers/visitation.controller.test.js
```diff
@@ -31,27 +32,30 @@ describe('Schedule Visit Route', () => {
...
-        .set('authorization', adminToken)
+        .set('authorization', userToken)
         .send(booking)
         .end((err, res) => {
           expect(res).to.have.status(201);
           expect(res.body.success).to.be.eql(true);
           expect(res.body.message).to.be.eql('Visit scheduled successfully');
           expect(res.body).to.have.property('schedule');
+          expect(res.body).to.have.property('schedule');
+          expect(res.body.schedule).to.have.property('propertyId');
```
andela-hpopoola, line 47: Instead of testing that is has `propertyId`, it is better to test that the returned propertyId matches the one you created on line 32 (propId).

You should also try to avoid using `.to.have.property` and test for the returned value instead to make your tests robust. This should only be used when the returned value cannot be gotten from the response.

test/controllers/visitation.controller.test.js
```diff
@@ -31,27 +32,30 @@ describe('Schedule Visit Route', () => {
...
         .post('/api/v1/visitation/schedule')
-        .set('authorization', adminToken)
+        .set('authorization', userToken)
         .send(booking)
         .end((err, res) => {
           expect(res).to.have.status(201);
           expect(res.body.success).to.be.eql(true);
           expect(res.body.message).to.be.eql('Visit scheduled successfully');
           expect(res.body).to.have.property('schedule');
+          expect(res.body).to.have.property('schedule');
```
andela-hpopoola, line 46: Duplicate assertion. Same test as the line above.

server/controllers/visitation.controllers.js
```diff
@@ -1,23 +1,44 @@
...
     const booking = req.locals;
     const { user } = req;
     scheduleVisitation({ ...booking, userId: user._id })
       .then((schedule) => {
-        // sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, user, {
-        //   link: `http://ballers.ng/activate?token=${token}`,
-        // });
+        const contentBottom = `
+          <strong> Name: </strong>: ${schedule.visitorName}\n, 
+          <strong> Phone: </strong> ${schedule.visitorPhone}\n.
```
andela-hpopoola, line 19: Email is missing here (we should include the email if the visitor inputs it when filling the schedule form.)

server/controllers/visitation.controllers.js
```diff
@@ -1,23 +1,44 @@
...
   book(req, res, next) {
     const booking = req.locals;
     const { user } = req;
     scheduleVisitation({ ...booking, userId: user._id })
       .then((schedule) => {
-        // sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, user, {
-        //   link: `http://ballers.ng/activate?token=${token}`,
-        // });
+        const contentBottom = `
+          <strong> Name: </strong>: ${schedule.visitorName}\n,
```
andela-hpopoola, line 18: `\n` will be displayed as a string as this is an HTML email.
```suggestion
          <strong> Name: </strong>: ${schedule.visitorName}<br />, 
```

server/controllers/visitation.controllers.js
```diff
@@ -1,23 +1,44 @@
...
     const { user } = req;
     scheduleVisitation({ ...booking, userId: user._id })
       .then((schedule) => {
-        // sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, user, {
-        //   link: `http://ballers.ng/activate?token=${token}`,
-        // });
+        const contentBottom = `
+          <strong> Name: </strong>: ${schedule.visitorName}\n, 
+          <strong> Phone: </strong> ${schedule.visitorPhone}\n.
+          Kindly check your dashboard for details
```
andela-hpopoola, line 20: I don't think this is necessary as this is an internal email.

server/controllers/visitation.controllers.js
```diff
@@ -1,23 +1,44 @@
...
-// import EMAIL_CONTENT from '../../mailer';
+import dotenv from 'dotenv';
+import { scheduleVisitation, getAllVisitations } from '../services/visitation.service';
+import EMAIL_CONTENT from '../../mailer';
+import { sendMail } from '../services/mailer.service';
 import httpStatus from '../helpers/httpStatus';
 
+dotenv.config();
+
+const { BALLERS_EMAIL } = process.env;
```
andela-hpopoola, line 9: Lets add a default value here to prevent the code from breaking if BALLERS_EMAIL is not set.
```suggestion
const { BALLERS_EMAIL } = process.env || 'dev@highrachy.com';
```

server/controllers/visitation.controllers.js
```diff
@@ -1,23 +1,44 @@
-import scheduleVisitation from '../services/visitation.service';
-// import { sendMail } from '../services/mailer.service';
-// import EMAIL_CONTENT from '../../mailer';
+import dotenv from 'dotenv';
+import { scheduleVisitation, getAllVisitations } from '../services/visitation.service';
+import EMAIL_CONTENT from '../../mailer';
+import { sendMail } from '../services/mailer.service';
 import httpStatus from '../helpers/httpStatus';
 
+dotenv.config();
```
andela-hpopoola, line 7: I think this is not needed.

server/controllers/visitation.controllers.js
```diff
@@ -1,23 +1,44 @@
...
+    getAllVisitations()
+      .then((schedules) => {
+        if (schedules.length > 0) {
+          res.status(httpStatus.OK).json({ success: true, schedules });
+        } else {
+          res
+            .status(httpStatus.NOT_FOUND)
+            .json({ success: false, message: 'No scheduled visits available' });
+        }
+      })
```
andela-hpopoola, line 39: Even if no schedules are found in the database, we can still make the query successful (success: true) as it is doesn't feel like an error to have an empty schedule in the database.

However, I am leaving this at your discretion, just ensure you are consistent with the behaviour you think its right and stick with it.
```suggestion
    res.status(httpStatus.OK).json({ success: true, schedules });
```

test/controllers/visitation.controller.test.js
```diff
@@ -0,0 +1,298 @@
...
+      const booking = VisitationFactory.build({ propertyId: propId });
+      request()
+        .post('/api/v1/visitation/schedule')
+        .set('authorization', userToken)
+        .send(booking)
+        .end((err, res) => {
+          expect(res).to.have.status(201);
+          expect(res.body.success).to.be.eql(true);
+          expect(res.body.message).to.be.eql('Visit scheduled successfully');
+          expect(res.body).to.have.property('schedule');
```
andela-hpopoola, line 45: We also need to add a test here for the email that is being sent.

tobieeniafe: I tried adding this as it was done in `user.controller.test.js` but it always broke the test with the error `"before each" hook for "returns the default mailer": TypeError: Attempted to wrap sendMail which is already wrapped` that's why I didn't add it

tobieeniafe: I've pushed the code with my current implementation commented out in `visitation.controller.test.js`. So you can see if it's my approach that's wrong.

server/controllers/visitation.controllers.js
```diff
@@ -15,9 +12,9 @@ const VisitationController = {
     scheduleVisitation({ ...booking, userId: user._id })
       .then((schedule) => {
         const contentBottom = `
-          <strong> Name: </strong>: ${schedule.visitorName}\n, 
-          <strong> Phone: </strong> ${schedule.visitorPhone}\n.
-          Kindly check your dashboard for details
+          <strong> Name: </strong>: ${schedule.visitorName}<br>, 
+          <strong> Phone: </strong> ${schedule.visitorPhone}<br>.
+          <strong> Email: </strong> ${schedule.visitorEmail}<br>.
```
andela-hpopoola, line 17: Email is optional, I think it would be better to put this in an `if statement` to prevent it from throwing an error if email is not present.

server/controllers/visitation.controllers.js
```diff
@@ -0,0 +1,43 @@
...
+    const { user } = req;
+    scheduleVisitation({ ...booking, userId: user._id })
+      .then((schedule) => {
+        const contentBottom = `
+          <strong> Name: </strong>: ${schedule.visitorName}<br>, 
+          <strong> Phone: </strong> ${schedule.visitorPhone}<br>.
+          <strong> Email: </strong> ${
+            schedule.visitorEmail ? schedule.visitorEmail : 'no email provided'
+          } <br>.
+        `;
```
andela-hpopoola, line 20: Here is how the mail template looks like if you are interested. You will also need to set up MAILTRAP API variables in your `.env` so that your mails can be sent to MAILTRAP.

Screenshot 2020-08-07 at 06 57 51


Screenshot 2020-08-07 at 06 59 07

```suggestion
        const contentBottom = `
          <strong> Name: </strong>: ${schedule.visitorName}<br /> 
          <strong> Phone: </strong> ${schedule.visitorPhone}<br />
          <strong> Email: </strong> ${
            schedule.visitorEmail ? schedule.visitorEmail : 'No email provided'
          } <br />
        `;
```

server/services/visitation.service.js
```diff
@@ -29,4 +29,9 @@ export const getAllVisitations = async () =>
         as: 'propertyInfo',
       },
     },
+    {
+      $project: {
+        'propertyInfo.assignedTo': 0,
```
andela-hpopoola, line 34: I am curious, any reasons for excluding this from the returned results?

test/controllers/visitation.controller.test.js
```diff
@@ -232,6 +232,9 @@ describe('Get all properties', () => {
             expect(res).to.have.status(200);
             expect(res.body.success).to.be.eql(true);
             expect(res.body).to.have.property('schedules');
+            expect(propId.equals(res.body.schedules[0].propertyId)).to.be.eql(true);
```
andela-hpopoola, line 235: Nice thinking with the test, however, I think this can be improved:
The test will be in 2 categories (but no demarcation is needed)
1. The first section tests that valid visitation data are returned.
2. The second section tests that a valid property information data are returned in propertyInfo.

```suggestion
           
            expect(res.body.schedules[0].propertyId).to.be.eq(propId.toString());
            expect(res.body.schedules[0].userId).to.be.eql(adminId.toString());
            expect(res.body.schedules[0].name).to.be.eql(booking.name);
            // ... add more booking tests here
            expect(res.body.schedules[0].propertyInfo[0].name).to.be.eql(demoProperty.name);
            // ... add more property tests here
```

PR #33, opened 2020-08-21
#174250315 add payment plan

tobieeniafe:
#### What does this PR do?
Add payment plan endpoints

add endpoints for admin to create, view and delete payment plans

`POST /api/v1/payment-plan/add`
`{
  "planName": "monthly",
  "planDescription": "monthly payment plan",
  "paymentFrequency": 1
}`
`GET /api/v1/payment-plan/all`
`DELETE /api/v1/payment-plan/delete/:id`

#174250315

### DIFF
--- a/server/controllers/paymentPlan.controllers.js
+++ b/server/controllers/paymentPlan.controllers.js
@@ -2,6 +2,8 @@ import {
   addPaymentPlan,
   deletePaymentPlan,
   getAllPaymentPlans,
+  updatePaymentPlan,
+  assignPaymentPlanToProperty,
 } from '../services/paymentPlan.service';
 import httpStatus from '../helpers/httpStatus';
 
@@ -29,6 +31,27 @@ const PaymentPlanController = {
     getAllPaymentPlans()
       .then((plans) => {
         res.status(httpStatus.OK).json({ success: true, plans });
+      })
+      .catch((error) => next(error));
+  },
+
+  update(req, res, next) {
+    const updatedPaymentPlan = req.locals;
+    const { user } = req;
+    updatePaymentPlan({ ...updatedPaymentPlan, updatedBy: user._id })
+      .then((plan) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Payment plan updated', plan });
+      })
+      .catch((error) => next(error));
+  },
+
+  assignPaymentPlan(req, res, next) {
+    const toBeAssigned = req.locals;
+    assignPaymentPlanToProperty(toBeAssigned)
+      .then(() => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Payment plan assigned to property' });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/paymentPlan.controllers.js
+++ b/server/controllers/paymentPlan.controllers.js
@@ -3,7 +3,6 @@ import {
   deletePaymentPlan,
   getAllPaymentPlans,
   updatePaymentPlan,
-  assignPaymentPlanToProperty,
 } from '../services/paymentPlan.service';
 import httpStatus from '../helpers/httpStatus';
 
@@ -41,17 +40,6 @@ const PaymentPlanController = {
     updatePaymentPlan({ ...updatedPaymentPlan, updatedBy: user._id })
       .then((plan) => {
         res.status(httpStatus.OK).json({ success: true, message: 'Payment plan updated', plan });
-      })
-      .catch((error) => next(error));
-  },
-
-  assignPaymentPlan(req, res, next) {
-    const toBeAssigned = req.locals;
-    assignPaymentPlanToProperty(toBeAssigned)
-      .then(() => {
-        res
-          .status(httpStatus.OK)
-          .json({ success: true, message: 'Payment plan assigned to property' });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/paymentPlan.controllers.js
+++ b/server/controllers/paymentPlan.controllers.js
@@ -1,0 +1,37 @@
+import {
+  addPaymentPlan,
+  deletePaymentPlan,
+  getAllPaymentPlans,
+} from '../services/paymentPlan.service';
+import httpStatus from '../helpers/httpStatus';
+
+const PaymentPlanController = {
+  add(req, res, next) {
+    const newPaymentPlan = req.locals;
+    const { user } = req;
+    addPaymentPlan({ ...newPaymentPlan, addedBy: user._id })
+      .then((plan) => {
+        res.status(httpStatus.CREATED).json({ success: true, message: 'Plan added', plan });
+      })
+      .catch((error) => next(error));
+  },
+
+  delete(req, res, next) {
+    const { id } = req.params;
+    deletePaymentPlan(id)
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Plan deleted' });
+      })
+      .catch((error) => next(error));
+  },
+
+  getAll(req, res, next) {
+    getAllPaymentPlans()
+      .then((plans) => {
+        res.status(httpStatus.OK).json({ success: true, plans });
+      })
+      .catch((error) => next(error));
+  },
+};
+
+export default PaymentPlanController;
--- a/server/models/paymentPlan.model.js
+++ b/server/models/paymentPlan.model.js
@@ -38,9 +38,6 @@ const PaymentPlanSchema = new mongoose.Schema(
       type: Number,
       required: true,
     },
-    propertiesAssignedTo: {
-      type: [ObjectId],
-    },
     addedBy: {
       type: ObjectId,
       required: true,
--- a/server/models/paymentPlan.model.js
+++ b/server/models/paymentPlan.model.js
@@ -1,0 +1,51 @@
+import mongoose from 'mongoose';
+
+/**
+ * @swagger
+ *  components:
+ *    schemas:
+ *      PaymentPlan:
+ *        type: object
+ *        required:
+ *          - planName
+ *          - planDescription
+ *          - paymentFrequency
+ *        properties:
+ *          planName:
+ *            type: string
+ *          planDescription:
+ *            type: string
+ *          paymentFrequency:
+ *            type: string
+ *        example:
+ *           planName: Weekly payment
+ *           planDescription: payment is made once a week
+ *           paymentFrequency: weekly
+ */
+
+const { ObjectId } = mongoose.Schema.Types;
+const PaymentPlanSchema = new mongoose.Schema(
+  {
+    planName: {
+      type: String,
+      required: true,
+    },
+    planDescription: {
+      type: String,
+      required: true,
+    },
+    paymentFrequency: {
+      type: String,
+      required: true,
+    },
+    addedBy: {
+      type: ObjectId,
+      required: true,
+    },
+  },
+  { timestamps: true },
+);
+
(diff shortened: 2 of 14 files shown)


server/controllers/paymentPlan.controllers.js
```diff
@@ -0,0 +1,37 @@
...
+    const newPaymentPlan = req.locals;
+    const { user } = req;
+    addPaymentPlan({ ...newPaymentPlan, addedBy: user._id })
+      .then((plan) => {
+        res.status(httpStatus.CREATED).json({ success: true, message: 'Plan added', plan });
+      })
+      .catch((error) => next(error));
+  },
+
+  delete(req, res, next) {
```
andela-hpopoola, line 19: 🤔 . This is a dicey situation. We should be careful with the delete feature with everything that deals with money. The delete endpoint for the payment plan should only delete the payment plan if it is not used by any property. A deleted payment plan in use will result in a bug if the property is retrieved.

server/models/paymentPlan.model.js
```diff
@@ -0,0 +1,51 @@
...
+  {
+    planName: {
+      type: String,
+      required: true,
+    },
+    planDescription: {
+      type: String,
+      required: true,
+    },
+    paymentFrequency: {
```
andela-hpopoola, line 37: This should be a number that denotes month. It will be difficult to automate calculations from payment plans if this is a string.

server/models/paymentPlan.model.js
```diff
@@ -0,0 +1,51 @@
...
+ *        example:
+ *           planName: Weekly payment
+ *           planDescription: payment is made once a week
+ *           paymentFrequency: weekly
+ */
+
+const { ObjectId } = mongoose.Schema.Types;
+const PaymentPlanSchema = new mongoose.Schema(
+  {
+    planName: {
```
andela-hpopoola, line 29: We could as well use name here (the plan prefix is not necessary, I notice that you do that for most of your naming convention). 

This will make the code more readable, something like `paymentPlan.name` seems better than `paymentPlan.planName`.

server/routes/index.js
```diff
@@ -5,6 +5,7 @@ import property from './property.routes';
 import enquiry from './enquiry.routes';
 import mailer from './mailer.routes';
 import visitation from './visitation.routes';
+import payment from './paymentPlan.routes';
```
andela-hpopoola, line 8: Lets be explicit with paymentPlan as we don't want a mix up with the payment and transaction model.

```suggestion
import paymentPlan from './paymentPlan.routes';
```

server/routes/index.js
```diff
@@ -13,6 +14,7 @@ export default (app) => {
   app.use('/api/v1/property', property);
   app.use('/api/v1/enquiry', enquiry);
   app.use('/api/v1/visitation', visitation);
+  app.use('/api/v1/payment', payment);
```
andela-hpopoola, line 17:
```suggestion
  app.use('/api/v1/payment-plan', paymentPlan);
```

server/routes/paymentPlan.routes.js
```diff
@@ -0,0 +1,85 @@
+import express from 'express';
+import { authenticate, schemaValidation, isAdmin, hasValidObjectId } from '../helpers/middleware';
+import addPlanSchema from '../schemas/paymentPlan.schema';
```
andela-hpopoola, line 3: `addPlanSchema` is not explicit enough. The codebase is getting bigger, it is easier to mix things up now.
```suggestion
import addPaymentPlanSchema from '../schemas/paymentPlan.schema';
```

server/routes/paymentPlan.routes.js
```diff
@@ -0,0 +1,85 @@
...
+import PaymentPlanController from '../controllers/paymentPlan.controllers';
+
+const router = express.Router();
+
+/**
+ * @swagger
+ * /payment/add:
```
andela-hpopoola, line 10: Please fix for other endpoints too.
```suggestion
 * /payment-plan/add:
```

server/routes/paymentPlan.routes.js
```diff
@@ -0,0 +1,85 @@
...
+ *            $ref: '#/components/schemas/PaymentPlan'
+ *      description: Get all owned enquiries
+ *     responses:
+ *      '200':
+ *        description: returns object of payment plans
+ *      '500':
+ *       description: Internal server error
+ */
+router.get('/all', authenticate, isAdmin, PaymentPlanController.getAll);
+
```
andela-hpopoola, line 61: We should have an endpoint to edit payment plans. Editing is important for fixing typos in plan name and making the description more descriptive.

However, the frequency field should not be updated since updating this will affect all the previously calculated price made with this payment plans.

andela-hpopoola (changes requested):

Can you integrate the payment plan model into the property model so that when properties are loaded via any of the property endpoint, the payment plan is also returned among the property details. This is the most important reason why this is important in the added to MVP. Thank you.

server/models/property.model.js
```diff
@@ -129,6 +129,9 @@ const PropertySchema = new mongoose.Schema(
     assignedTo: {
       type: [ObjectId],
     },
+    paymentPlan: {
```
andela-hpopoola, line 132: PaymentPlan should be an array of ObjectId. 
(If you study the mockup - you will notice that we have multiple payment plans.)

Your implementation is different from the concept in the mockup 😏  

In the mockup, you need to study all the places the payment plan is used, you will notice that it is returned with the property information.

1. Single Portfolio Page
 This means that the returned result for the `/property/:id` endpoint need to return the names of the payment plan.

Screenshot 2020-08-25 at 07 13 04


2. List of all Properties contains payment plan. We need to ensure that the payment plan is aggregated in all property endpoint so that it is returned as part of the property information.

See another instance here:
Screenshot 2020-08-25 at 07 23 28

To save time and energy, lets ensure we return the aggregated payment plan for all property endpoints.

server/controllers/paymentPlan.controllers.js
```diff
@@ -32,6 +34,27 @@ const PaymentPlanController = {
...
+    const updatedPaymentPlan = req.locals;
+    const { user } = req;
+    updatePaymentPlan({ ...updatedPaymentPlan, updatedBy: user._id })
+      .then((plan) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Payment plan updated', plan });
+      })
+      .catch((error) => next(error));
+  },
+
+  assignPaymentPlan(req, res, next) {
```
andela-hpopoola, line 48: This is a great implementation, however, it will cause the user and application bad user experience. I believe payment plans should be added when the user (developer/admin) is adding a new property (not assigned individually). This will be implemented via a multiple Select field in the frontend so that the user can add and edit payment plans at ease.

A better implementation will be to remove this method and add payment plans to add property/update property implementation. The ids of the payment plan are saved in the property model as an array. If you are curious about the multi select, here is a link to the one I used in the application https://react-select.com/home (I am referring to the 2nd example for multiple select field)

I hope you understand, please free to follow up with me if this is unclear or you have a much better implementation in mind.

andela-hpopoola (changes requested):
Thank you for working on this. I believe the implementation can be improved to make the endpoint more effective.

server/models/paymentPlan.model.js
```diff
@@ -7,41 +7,47 @@ import mongoose from 'mongoose';
...
+    description: {
       type: String,
       required: true,
     },
     paymentFrequency: {
-      type: String,
+      type: Number,
       required: true,
     },
+    propertiesAssignedTo: {
```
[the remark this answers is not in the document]
tobieeniafe: this is to kow is a payment plan is linked to a property, to prevent it from being deleted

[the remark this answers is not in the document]
andela-hpopoola: I understand your point but this is an extra overhead over the payment plan and it is not necessary(as we will need to add an extra query to add the properties to the payment plan). This is extra work as we will need to keep track of the payment plans in the property model and also keep track of the payment plan model at the same time. If the payment plan field in the property is added, we must add the property field too in the payment plan model. This also applies to when a property is updated and deleted. This will be extra code and extra work (it doesn't worth the stress, abeg)

Moreover, we can know if a payment plan is linked to a property by querying the properties since the payment plan is saved in the property model too. 🤷‍♂️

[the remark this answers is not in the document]
tobieeniafe: `hasBeenAssignedToProperty` `true` ?🤔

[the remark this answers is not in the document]
andela-hpopoola: The reason why querying the property model is the best approach is because it will only be done when the admin is deleting. The problem with this are:
1. You will need to write an additional query to update the payment plan `hasBeenAssignedToProperty` field to true every time a new property is added. 
2. We are not assigning payment plans one by one via an endpoint. You will need to remove the assignment endpoint, We are assigning payment plans is multiples -- An admin can assign 4 to 5 payment plan at the same time when he is adding a new property (remember the multi-select field in my earlier comment), that means we will make 4 to 5 queries to update the respective payment plan `hasBeenAssignedToProperty`.
3. Every time a property is updated, we need to check that the payment plan has not been updated because if it has been updated (i.e a payment plan has been removed, and another one has been added) we will need to check if the removed property is the only property with the payment plan and use that to update the value of `hasBeenAssignedToProperty`.

I am actually being considerate by telling you to remove this. However, if you think adding the `asBeenAssignedToProperty` field approach is better, my only plead is for you to write tests to at least simulate my points 2 and 3 so that we are certain that we cover all the edge cases.

server/schemas/paymentPlan.schema.js
```diff
@@ -0,0 +1,20 @@
...
+  paymentFrequency: Joi.number().label('Plan Payment Frequency').required(),
+});
+
+export const updatePaymentPlanSchema = Joi.object({
+  id: Joi.objectId().label('Plan ID').required(),
+  name: Joi.string().label('Plan Name').optional(),
+  description: Joi.string().label('Plan Description').optional(),
+});
+
+export const assignPaymentPlanSchema = Joi.object({
```
andela-hpopoola, line 17: Can we remove this, not sure we need it or what do you think?

server/services/paymentPlan.service.js
```diff
@@ -0,0 +1,46 @@
...
+
+export const deletePaymentPlan = async (id) => {
+  const plan = await getPaymentPlanById(id).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+  const propertiesAssignedTo = await getPropertiesWithPaymentPlanId(id);
+  if (propertiesAssignedTo.length > 0) {
+    throw new ErrorHandler(
+      httpStatus.PRECONDITION_FAILED,
+      'Cannot delete payment plan with properties assigned',
```
andela-hpopoola, line 25: Not too important but necessary for User Experience, I believe that the error message will be more informative if we return the name of the properties (as an array) the payment plan is currently assigned to. It will make it easier to locate the property and update it.

andela-hpopoola, line 25: It is advisable to make the error message more descriptive. Let's add the number of properties it is assigned to. Please use `property` for 1 property and plural form for more than one property in the updated error message as shown below:

```suggestion
      `Payment plan cannot be deleted as it is currently assigned to 2 properties`,
```

andela-hpopoola (approved):
Thank you for working on this. I left a few non blocking comments. Feel to free to merge it after fixing them.

server/services/paymentPlan.service.js
tobieeniafe: any suggestion on how to implement this?

PR #34, opened 2020-08-24
#174466910 streamline search endpoint 

tobieeniafe:
#### What does this PR do?
Streamline Search endpoint.

Streamline Search endpoint to get distinct states & house types

`GET /api/v1/property/available`

#174466910

### DIFF
--- a/server/routes/property.routes.js
+++ b/server/routes/property.routes.js
@@ -148,27 +148,6 @@ router.get('/all', authenticate, isAdmin, PropertyController.getAllProperties);
 
 /**
  * @swagger
- * path:
- *  /property/one/:id:
- *    get:
- *      parameters:
- *        - in: query
- *          name: token
- *          schema:
- *            type: string
- *          description: verifies user access
- *      summary: Gets a property based by its ID
- *      tags: [Property]
- *      responses:
- *        '200':
- *          description: Property found
- *        '500':
- *          description: Internal server error
- */
-router.get('/one/:id', authenticate, hasValidObjectId, PropertyController.getOneProperty);
-
-/**
- * @swagger
  * /property/search:
  *   post:
  *     tags:
@@ -197,7 +176,7 @@ router.post(
 
 /**
  * @swagger
- * /property/available:
+ * /property/available-options:
  *   get:
  *     tags:
  *       - Property
@@ -216,6 +195,27 @@ router.post(
  *      '500':
  *       description: Internal server error
  */
-router.get('/available', authenticate, PropertyController.getDistinct);
+router.get('/available-options', authenticate, PropertyController.getDistinct);
+
+/**
+ * @swagger
+ * path:
+ *  /property/:id:
+ *    get:
+ *      parameters:
+ *        - in: query
+ *          name: token
+ *          schema:
+ *            type: string
+ *          description: verifies user access
+ *      summary: Gets a property based by its ID
+ *      tags: [Property]
+ *      responses:
+ *        '200':
+ *          description: Property found
+ *        '500':
+ *          description: Internal server error
+ */
+router.get('/:id', authenticate, hasValidObjectId, PropertyController.getOneProperty);
 
 module.exports = router;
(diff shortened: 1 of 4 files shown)


server/routes/property.routes.js
```diff
@@ -149,7 +149,7 @@ router.get('/all', authenticate, isAdmin, PropertyController.getAllProperties);
 /**
  * @swagger
  * path:
- *  /property/:id:
+ *  /property/one/:id:
```
andela-hpopoola, line 152: I am curious, why are we updating this endpoint. We can move it below the new endpoint to prevent conflict instead of updating it. Can we please revert? 

Also, can we try to look for alternatives before updating any endpoint so as not to create more bugs, fixes, and work for the team? This should only be done as the last resort, not as a quick fix.

server/routes/property.routes.js
```diff
@@ -195,4 +195,27 @@ router.post(
   PropertyController.search,
 );
 
+/**
+ * @swagger
+ * /property/available:
```
andela-hpopoola, line 200: Let's use available options here so that we don't confuse it with available properties.
```suggestion
 * /property/available-options:
```

server/services/property.service.js
```diff
@@ -216,3 +216,13 @@ export const searchThroughProperties = async ({ state, city, houseType, minPrice
       },
     },
   ]);
+
+export const getDistinctPropertyStatesAndTypes = async () => {
```
andela-hpopoola, line 220: We should use a more explicit name so that the method name still makes sense if we update the fields based on future suggestions.
```suggestion
export const getAvailablePropertyOptions = async () => {
```

server/services/property.service.js
```diff
@@ -216,3 +216,13 @@ export const searchThroughProperties = async ({ state, city, houseType, minPrice
...
+  const states = await Property.distinct('address.state');
```
andela-hpopoola, line 221: This is not an ideal way of getting around this. Using two queries is resources intensive and more time-consuming. It would even be worst if Namo and Co. suggests that we add more field. 

The correct implementation is to use a single query that returns both values instead of two. MongoDB distinct is not the best for this, we should use aggregate.

Please see a similar solution:
https://stackoverflow.com/questions/40691324/performing-distinct-on-multiple-fields-in-mongodb

test/controllers/property.controller.test.js
```diff
@@ -1388,3 +1388,58 @@ describe('Search Through Properties', () => {
...
+
+    context('with a valid token & id', () => {
+      it('returns successful payload', (done) => {
+        request()
+          .get('/api/v1/property/available')
+          .set('authorization', userToken)
+          .end((err, res) => {
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body).to.have.property('availableFields');
```
andela-hpopoola, line 1410: I don't buy the idea of testing for the property of a field when we can test for the exact value. This defeats the exact purpose of writing tests. The test will not be less functional and would not catch future bugs. Testing is more than passing the test coverage alone.

The ideal test will be to 3 properties where only 2 properties have the same states, another 2 have the same houseType, then test that this returns only 2 states (and of course test that the value corresponds with the 2 unique values you are expecting).

Another test should contain 3 properties but with different states and houseType, this should return 3 values for both.

Another test should contain 3 properties with the same states and houseType, this should return only 1 value. 

With these three tests, we would have more confidence that the endpoint works as expected.

test/controllers/property.controller.test.js
```diff
@@ -1388,3 +1388,58 @@ describe('Search Through Properties', () => {
...
+            expect(res.body.availableFields).to.have.property('states');
```
andela-hpopoola, line 1411: This test will pass even if the endpoint returns no value (i.e. if it returns null for states and houseTypes).

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on the implementation and improving the tests to improve our confidence in the endpoint.

I also noticed that you have been skipping tests for services files. I totally understand that it is because of our tight deadline and I am okay with controller tests, for now, however, we would need to write the tests later to improve our confidence of the code. Thank you.

andela-hpopoola (approved):
Looks delicious to me. This works and looks great. Just a single fix, and kabow, merge it asap.

PR #35, opened 2020-08-24
#174465216 fix broken image and similar endpoint issues

tobieeniafe:
#### What does this PR do?
Fix broken image and similar endpoint issues

Fix broken image and similar endpoint issues

#174465216


server/services/property.service.js
```diff
@@ -193,18 +202,17 @@ export const searchThroughProperties = async ({ state, city, houseType, minPrice
     {
       $project: {
         _id: 1,
+        name: 1,
```
andela-hpopoola, line 205: Can we add this to the corresponding test too?

server/services/user.service.js
```diff
@@ -254,15 +254,20 @@ export const getAllUserProperties = async (userId) =>
       $project: {
         _id: 0,
         'ownedProperties._id': 1,
+        'ownedProperties.name': 1,
```
andela-hpopoola, line 257: We need to add this to the related tests too.

andela-hpopoola (approved):
Thank you for the quick turnaround. This looks good to me. Only a few comments on ensuring that the updated fields exist in the related to catch this type of bugs in future updates.

PR #36, opened 2020-08-27
#174250297 add transaction model

tobieeniafe:
#### What does this PR do?
Add transaction model and endpoint

Create transaction model and endpoint to record transactions

`GET /api/v1/transaction/all`
`POST /api/v1/transaction/add`

#174250297

### DIFF
--- a/server/controllers/transaction.controllers.js
+++ b/server/controllers/transaction.controllers.js
@@ -1,0 +1,31 @@
+import { addTransaction, getAllTransactions } from '../services/transaction.service';
+import httpStatus from '../helpers/httpStatus';
+
+const TransactionController = {
+  add(req, res, next) {
+    const newTransaction = req.locals;
+    const { user } = req;
+    if (newTransaction.userId) {
+      newTransaction.adminId = user._id.toString();
+    } else {
+      newTransaction.userId = user._id.toString();
+    }
+    addTransaction(newTransaction)
+      .then((transaction) => {
+        res
+          .status(httpStatus.CREATED)
+          .json({ success: true, message: 'Transaction added', transaction });
+      })
+      .catch((error) => next(error));
+  },
+
+  getAll(req, res, next) {
+    getAllTransactions()
+      .then((transactions) => {
+        res.status(httpStatus.OK).json({ success: true, transactions });
+      })
+      .catch((error) => next(error));
+  },
+};
+
+export default TransactionController;
--- a/server/controllers/transaction.controllers.js
+++ b/server/controllers/transaction.controllers.js
@@ -4,12 +4,6 @@ import httpStatus from '../helpers/httpStatus';
 const TransactionController = {
   add(req, res, next) {
     const newTransaction = req.locals;
-    const { user } = req;
-    if (newTransaction.userId) {
-      newTransaction.adminId = user._id.toString();
-    } else {
-      newTransaction.userId = user._id.toString();
-    }
     addTransaction(newTransaction)
       .then((transaction) => {
         res
--- a/server/controllers/transaction.controllers.js
+++ b/server/controllers/transaction.controllers.js
@@ -2,7 +2,8 @@ import {
   addTransaction,
   getAllTransactions,
   updateTransaction,
-  getUserTransactionsByPropertyAndUser,
+  getUserTransactionsByProperty,
+  getTransactionsByUser,
 } from '../services/transaction.service';
 import httpStatus from '../helpers/httpStatus';
 
@@ -38,9 +39,18 @@ const TransactionController = {
       .catch((error) => next(error));
   },
 
-  getUserTransactionsByPropertyAndUser(req, res, next) {
-    const filter = req.locals;
-    getUserTransactionsByPropertyAndUser(filter)
+  getAllPersonal(req, res, next) {
+    const id = req.user._id;
+    getTransactionsByUser(id)
+      .then((transactions) => {
+        res.status(httpStatus.OK).json({ success: true, transactions });
+      })
+      .catch((error) => next(error));
+  },
+
+  getTransactionsByProperty(req, res, next) {
+    const { propertyId } = req.locals;
+    getUserTransactionsByProperty(propertyId)
       .then((transactions) => {
         res.status(httpStatus.OK).json({ success: true, transactions });
       })
--- a/server/controllers/transaction.controllers.js
+++ b/server/controllers/transaction.controllers.js
@@ -1,10 +1,16 @@
-import { addTransaction, getAllTransactions } from '../services/transaction.service';
+import {
+  addTransaction,
+  getAllTransactions,
+  updateTransaction,
+  getUserTransactionsByPropertyAndUser,
+} from '../services/transaction.service';
 import httpStatus from '../helpers/httpStatus';
 
 const TransactionController = {
   add(req, res, next) {
     const newTransaction = req.locals;
-    addTransaction(newTransaction)
+    const { user } = req;
+    addTransaction({ ...newTransaction, adminId: user._id })
       .then((transaction) => {
         res
           .status(httpStatus.CREATED)
@@ -15,6 +21,26 @@ const TransactionController = {
 
   getAll(req, res, next) {
     getAllTransactions()
+      .then((transactions) => {
+        res.status(httpStatus.OK).json({ success: true, transactions });
+      })
+      .catch((error) => next(error));
+  },
+
+  update(req, res, next) {
+    const updatedInfo = req.locals;
+    updateTransaction(updatedInfo)
+      .then((transaction) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Payment date updated', transaction });
+      })
+      .catch((error) => next(error));
+  },
+
+  getUserTransactionsByPropertyAndUser(req, res, next) {
+    const filter = req.locals;
+    getUserTransactionsByPropertyAndUser(filter)
       .then((transactions) => {
         res.status(httpStatus.OK).json({ success: true, transactions });
       })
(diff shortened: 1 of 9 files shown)


andela-hpopoola (changes requested):
Thank you for working on this. This looks good and the implementation makes sense. I left a few comments, but I will test it properly later in the day (after work).

server/controllers/transaction.controllers.js
```diff
@@ -0,0 +1,31 @@
+import { addTransaction, getAllTransactions } from '../services/transaction.service';
+import httpStatus from '../helpers/httpStatus';
+
+const TransactionController = {
+  add(req, res, next) {
+    const newTransaction = req.locals;
+    const { user } = req;
+    if (newTransaction.userId) {
```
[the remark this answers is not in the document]
tobieeniafe: the `userId` isn't compulsory so that the endpoint could be used to create online(user initiated) and offline(admin initiated) transactions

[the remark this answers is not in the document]
andela-hpopoola: The userId should actually be compulsory. Let me explain
1. If the user makes an online payment, we will use the userId to know the user that made the payment. Since it is an online payment, we dont need the adminId.

2. For offline payments, the admin need to select the user that has made the payment into his account. Without the userId, there will be no way to query the total transaction (online and offline) made by the user. The admin Id is only compulsory because if there are discrepancies with any offline payment, the adminId can be used to trace the problem.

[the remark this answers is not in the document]
tobieeniafe: the intended idea here is, if `userId` isn't sent it automatically uses the id in the user token as the `userId` handling it as an online payment. And if `userId` is present, it uses the id in the token as the `adminid`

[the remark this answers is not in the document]
andela-hpopoola: It is not recommended to use the userId in token since this used by the admin and user. We don't want to mistakenly save the admin Id as userId. UserId is the most important field here so it should be required from Joi schema.

If the userId is not returned for whatever reason, it is safer to throw an error than saving the transaction or manipulating the user Id.

test/controllers/transaction.controller.test.js
```diff
@@ -0,0 +1,223 @@
...
+        request()
+          .get('/api/v1/transaction/all')
+          .set('authorization', adminToken)
+          .end((err, res) => {
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body).to.have.property('transactions');
+            // TODO FIX FAILING TEST HERE
+            // expect(res.body.transactions[0]._id).to.be.eql(transactionId.toString());
+            // expect(res.body.transactions[0].userInfo._id).to.be.eql(_id.toString());
```
andela-hpopoola, line 175: This is a very important test, so I was forced to look into it.

The test was failing because adminId was missing in the transaction. Adding adminId to line 139 fixes the test.

andela-hpopoola: However, we need to test that the second one with no admin Id passes too. This is important because we want the code to return the Paystack record we have saved too and these records will have no admin Id. 

Hence, we need to ensure that line 161 remains like this 
```
await addTransaction(TransactionFactory.build());
``` 

and ensure we update the transaction aggregage to return for fields with no adminId. 

I found this, not sure if it will help 
https://stackoverflow.com/questions/48354575/mongodb-mongoose-make-aggregate-lookup-as-optional

**Very Important** 
You might need to look at other models and ensure that the aggregate function used in the code doesn't omit any field when joined. Lets ensure that this bug doesn't exist please.

Thank you.

andela-hpopoola (changes requested):
Thank you for working on this. I just had the opportunity to test it and it works great. Left a few comments on improving the code and tests to reduce future bugs. Thank you.

server/models/transaction.model.js
```diff
@@ -0,0 +1,72 @@
...
+    },
+    paymentSource: {
+      type: String,
+      required: true,
+    },
+    amount: {
+      type: Number,
+      required: true,
+    },
+    paidOn: {
```
andela-hpopoola, line 60: This should be required. It is important to set the date the payment is paid when the admin inputs the payment details.

server/controllers/transaction.controllers.js
```diff
@@ -0,0 +1,51 @@
...
+    updateTransaction(updatedInfo)
+      .then((transaction) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Payment date updated', transaction });
+      })
+      .catch((error) => next(error));
+  },
+
+  getUserTransactionsByPropertyAndUser(req, res, next) {
```
andela-hpopoola, line 41: I think your current implementation of this comment is just one out of the multiple endpoint that should be created. We cannot use this to generate all transaction details for a user.

https://github.com/Highrachy/ballers-api/pull/36/files#r479644834

An easier way to get a user transaction is to create a single endpoint where the userId is gotten from the token and returns all the payments with the userId but grouped by propertyId. You can take it further by using the database to calculate the sum of returned values.

For the admin, we need another endpoint to get all the payments made for a property via the propertyId. This will get all the payments and user info to enable the admin to know all the payments made to him/her at a glance.

The last part is the one you have implemented, which enables the admin to getUserTransactions By Property Id. We need to protect it with the admin authentication middleware and ensure that the admin is the owner of the property he is viewing. We don't want to make random people see other people transactions.

Not sure if I make any sense, I will try to explain better during standup.

test/controllers/transaction.controller.test.js
```diff
@@ -0,0 +1,336 @@
...
+        .end((err, res) => {
+          expect(res).to.have.status(403);
+          expect(res.body.success).to.be.eql(false);
+          expect(res.body.message).to.be.eql('Token needed to access resources');
+          done();
+        });
+    });
+  });
+
+  context('with invalid ttransaction', () => {
```
andela-hpopoola, line 272: Typo:
```suggestion
  context('with invalid transaction', () => {
```

andela-hpopoola (changes requested):
Thank you for working on this. I added more comments on adding more endpoints to make it easier and secured to get transaction information for the admin and user.

server/routes/transaction.routes.js
```diff
@@ -3,7 +3,7 @@ import { authenticate, schemaValidation, isAdmin } from '../helpers/middleware';
 import {
   addTransactionSchema,
   updateTransactionSchema,
-  getTransactionInfoBypropertySchema,
+  getTransactionsBypropertySchema,
```
andela-hpopoola, line 6:
```suggestion
  getTransactionsByPropertySchema,
```

server/routes/transaction.routes.js
```diff
@@ -92,21 +92,44 @@ router.put(
   TransactionController.update,
 );
 
+/**
+ * @swagger
+ * /transaction/personal:
```
andela-hpopoola, line 97:
```suggestion
 * /transaction/user:
```

test/controllers/transaction.controller.test.js
```diff
@@ -334,3 +334,175 @@ describe('Update Transaction route', () => {
...
+      it('returns successful payload', (done) => {
+        request()
+          .get('/api/v1/transaction/personal')
+          .set('authorization', adminToken)
+          .end((err, res) => {
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body).to.have.property('transactions');
+            expect(res.body.transactions[0]._id).to.be.eql(transactionId.toString());
+            expect(res.body.transactions[0].propertyInfo._id).to.be.eql(propertyId.toString());
```
andela-hpopoola, line 380: I think it is a better idea to test at least the property name and property image are returned via this endpoint.

test/controllers/transaction.controller.test.js
```diff
@@ -334,3 +334,175 @@ describe('Update Transaction route', () => {
...
+    context('with a valid token & id', async () => {
+      it('returns successful payload', (done) => {
+        request()
+          .get('/api/v1/transaction/personal')
+          .set('authorization', adminToken)
+          .end((err, res) => {
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body).to.have.property('transactions');
+            expect(res.body.transactions[0]._id).to.be.eql(transactionId.toString());
```
andela-hpopoola, line 379: I think it is a better idea to test for other transaction fields here (remember that omitting this type of test introduced the main image bug)

test/controllers/transaction.controller.test.js
```diff
@@ -334,3 +334,175 @@ describe('Update Transaction route', () => {
...
+    });
+  });
+
+  describe('when transactions exist in db', () => {
+    before(async () => {
+      await addProperty(property);
+      await addTransaction(transaction);
+    });
+
+    context('with a valid token & id', async () => {
```
andela-hpopoola, line 370: Can you please add another test for an empty `adminId` so that we are sure this works when used with Paystack services.

andela-hpopoola (commented):
I did a quick look and this looks good. I will test it on my system tomorrow morning.

server/routes/transaction.routes.js
```diff
@@ -0,0 +1,147 @@
...
+ *      '200':
+ *        description: returns object of transactions
+ *      '500':
+ *       description: Internal server error
+ */
+router.get('/user', authenticate, TransactionController.getAllPersonal);
+
+/**
+ * @swagger
+ * /transaction/details:
```
andela-hpopoola, line 120: This is not getting the transaction, rather it is getting all transactions for a given property. Imho, I feel this is better as a get request and the endpoint can be updated to reflect that it is for property. 

In the nearest future, we will need to create a transaction details endpoint that gets the transaction details of a single transaction based on the transaction Id.

```suggestion
 * /transaction/property/:propertyId:
```

andela-hpopoola (approved):
Thank you for working on this. I left a  comment on replacing the transaction details endpoint with transaction properties to make it more meaningful. It is actually a blocking comment, but I trust you will fix it right.

Feel free to merge after you have made these changes. Thank you.

PR #42, opened 2020-09-10
add referral code to user registration

tobieeniafe:
#### What does this PR do?
Add referral code to user registration

Add referral code to user registration so as to reflect referrals

`POST api/v1/user/register` `referralCode: [referralCode]`
`GET api/v1/referral/:id`
`PUT api/v1/referral/registered` `referralId`
`PUT api/v1/referral/rewarded` `referralId`


server/models/user.model.js
```diff
@@ -130,6 +130,9 @@ const UserSchema = new mongoose.Schema(
       type: String,
       unique: true,
     },
+    referrals: {
```
andela-hpopoola, line 133: We can add this to the user model or create a referral model to handle all of this. Personally, I prefer creating the referral model but I am leaving the ball in your court (show the one you think will be easier for you to maintain).

I have not done a deep dive into this yet, however, here are future things that comes to mind at the moment:
1. An endpoint / there should be a way for a user to load all his referrals information (the table in the screenshot below). Here are a number of things we need to keep track of:
    - The id of the referral so that we can use it to get email address
    - The current status of the referral 
    - The date of registration
Screenshot 2020-09-10 at 19 57 17

2. An admin should be able to get all referral information. This will be tied to the payment information / property status of the user. This will enable the admin/vendor to know when to pay.

3. We need another model to keep track of referral email sent out by a user to avoid him / her spamming another user (the form above the red border).

server/controllers/referral.controllers.js
```diff
@@ -0,0 +1,37 @@
...
+  getReferraslByRefereeId(req, res, next) {
+    const refereeId = req.params.id;
+    getReferraslByRefereeId(refereeId)
+      .then((referrals) => {
+        res.status(httpStatus.OK).json({ success: true, referrals });
+      })
+      .catch((error) => next(error));
+  },
+
+  updateReferralToRegistered(req, res, next) {
```
andela-hpopoola, line 18: I believe this should be done automatically when the referee registers on the website. An additional endpoint is not needed.

server/controllers/referral.controllers.js
```diff
@@ -0,0 +1,37 @@
+import {
+  getReferraslByRefereeId,
+  updateReferralToRegistered,
+  updateReferralToRewarded,
+} from '../services/referal.service';
+import httpStatus from '../helpers/httpStatus';
+
+const ReferralController = {
+  getReferraslByRefereeId(req, res, next) {
```
andela-hpopoola, line 9: I am not sure if this really makes sense. Let me give you the workflow:

Samuel refers Bose by one of the following
1. **LINK REFERRAL**: Sharing his link on social media i.e https://ballers.ng/refcode/sa1890. Note that with this, we dont even know where Samuel has shared the code.
2. **DIRECT REFERRAL**: Invite directly by filling the form directly i.e the referral form with email and optional name. 

From the two methods of referring people, the refereeId is not involved because an account will not be created for them.

server/models/referral.model.js
```diff
@@ -0,0 +1,56 @@
...
+const { ObjectId } = mongoose.Schema.Types;
+const ReferralSchema = new mongoose.Schema(
+  {
+    userId: {
+      type: ObjectId,
+      required: true,
+    },
+    refereeId: {
+      type: ObjectId,
+      required: true,
```
andela-hpopoola, line 40: This should be optional and should only be updated when the user registers.

server/models/referral.model.js
```diff
@@ -0,0 +1,56 @@
...
+    },
+    registrationDate: {
```
andela-hpopoola, line 42: This is not needed. We can get it from the createdAt field in the user model when we join via the referee Id.

server/models/referral.model.js
```diff
@@ -0,0 +1,56 @@
...
+ *           status: John
+ */
+
+const { ObjectId } = mongoose.Schema.Types;
+const ReferralSchema = new mongoose.Schema(
+  {
+    userId: {
+      type: ObjectId,
+      required: true,
+    },
```
andela-hpopoola, line 37: I think we need to add firstName and email to the model. They should be optional though. Here is my thought process (you can actually improve on it)

1. When someone refers another person via DIRECT REFERRAL, the details should be saved in the referral model. Hence, the reason for having the email and firstName in the model to prevent spamming people

1. When a user registers with a referral code (we don't know if they got the link directly or from an external link or social media -- and we shouldn't care where they got it from), the ref code is sent with the user details to the endpoint

1. When the endpoint receives a refcode, it saves the user and get the user id (to be used as refereeId) and checks if the refCode and email exist in the referral model, 
       a. if it exists, it should update the existing referral field as deem fit
       b. if it doesn't exist, it should use the refCode to get the user that has the referral info and add the necessary information into the referral model

server/models/referral.model.js
```diff
@@ -0,0 +1,56 @@
...
+    registrationDate: {
+      type: Date,
+      required: true,
+    },
+    status: {
+      type: String,
+      default: REFERRAL_STATUS.SENT,
+    },
```
andela-hpopoola, line 49: We need more fields here for future handling of referral rewards. Maybe we should add a reward object that contains the following
  -  amount: This is the amount the referrer will be paid based on some things we need to confirm from Nnamdi. This can be done via a cron job (we will handle it later)
  -  status: This tells whether the rewards has be paid, or InProgress 
  -  paidBy: The vendor (i.e userId) that paid the reward
  -  paidOn: The date the reward is paid.

Something like this: 
```js
reward: {
amount: Number,
status: String,
... 
}
```

server/services/referal.service.js
```diff
@@ -0,0 +1,72 @@
+import mongoose from 'mongoose';
```
andela-hpopoola, line 1: There is a typo with the file name. Can you update it to `referral.service.js`, please?

server/services/referal.service.js
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  try {
+    const referralInfo = await new Referral(info).save();
+    return referralInfo;
+  } catch (error) {
+    throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error adding referral info', error);
+  }
+};
+
```
andela-hpopoola, line 23: We will need a method here to save the referral info from the DIRECTLY INVITE REFERRAL form. This should send a mail to the referee. The link attached to this mail will have the refCode and email address.

We will also need another method to getReferralByEmail so that we can get the firstName and display it in the modal shown on the frontpage. This should be a public route (without authentication)

server/services/referal.service.js
```diff
@@ -0,0 +1,72 @@
...
+export const getReferraslByRefereeId = async (refereeId) =>
```
andela-hpopoola, line 24: Please be mindful of your spellings.
```suggestion
export const getReferralsByRefereeId = async (refereeId) =>
```

server/routes/referral.routes.js
```diff
@@ -0,0 +1,93 @@
...
+ *       description: Internal server error
+ */
+router.put(
+  '/rewarded',
+  authenticate,
+  isAdmin,
+  schemaValidation(updateReferralSchema),
+  ReferralController.updateReferralToRewarded,
+);
+
```
andela-hpopoola, line 92: A route is needed to load all the referral information in a table for the user. A similar route is need for an admin so that he knows when to reward the referrer.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on my thought process. Please let me know if you dont understand any part, otherwise, feel free to improve and ask more questions.

server/controllers/referral.controllers.js
tobieeniafe: this is to get all the information of the referrals a user referred

server/models/referral.model.js
tobieeniafe: the referral code is added during registration, and it updates on creation.

server/routes/referral.routes.js
tobieeniafe: a similar method already exists `getReferralsByRefereeId`

andela-hpopoola: Awesome, however, the name is confusing. Can you update the method name to `getAllUserReferrals`. 

Here is the difference between referrer and referee:

> “Referrer” means a Candidate and voluntarily participates in the Scheme by referring eligible friends to become Referees. “Referee” means a person who is not an existing Candidate of the Company and who has given his/her approval and whom the Referrer has referred through the Scheme.

server/models/referral.model.js
andela-hpopoola: I am sorry, I don't understand your definition of registration and creation. Can you explain better?

server/controllers/referral.controllers.js
andela-hpopoola: The method name is confusing. I have explained this in another comment.

server/models/referral.model.js
tobieeniafe: during user registration, the field for `referralCode` has been added. This would allow creation of the referral model on registration

andela-hpopoola: Understood but the problem with this is that we need to keep track of INVITE SENT. This is the first stage for DIRECT REFERRAL and the user has not registered at this point. Since the referred user hasn't registered as they have just been sent an invite email, how do you propose we get the `referralId` if it is always required?

tobieeniafe: referral invites and the referrals should be stored separately tho

andela-hpopoola: Having a different model for referral invites works too and has some additional advantages. However, I don't think it is necessary as we can save the referral invites in the referral table if referrerId is not required, I will not discourage you from during it if you feel it is a better implementation; just ensure that you manage both models very well to prevent data duplication and also prevent making unnecessary requests.

server/controllers/referral.controllers.js
tobieeniafe: I believe both `LINK REFERRAL` & `DIRECT REFERRAL` should have the same flow. 
ie. for `LINK REFERRAL`, the `referralCode` is gotten from the URL then the new user is taken to the registration page with the `referralCode` filled in. While for  `DIRECT REFERRAL`, the user adds the  `referralCode`  to the form manually

andela-hpopoola: The implementation of the 2 workflows should be the same for the API, so I am not sure of your point. The API implementation should be generic, so it shouldn't matter how the referral code is gotten. (the generic web page workflow could change in the future and this should have minimal effect on the API).

Redirecting the user to the registration URL with the `referralCode` filled in is poor user experience. When people are referred to something new, they will want to do their own personal investigation before registering. Nobody wants to register for things they have no knowledge of. Even if we redirect them, we cannot force them from clicking on another page (like About Us) and the referral code would be lost.

Also, telling the user to type in the referralCode is also bad user experience. We could get this with no additional effort from the URL sent via DIRECT / LINK REFERRAL, why ask the user for it again? Asking the user to type it will increase the number of fields on the form (the more fields you add for registration, the fewer people would be interested in signing up). Even if people are not discouraged by this, how will you be sure that they will type the right referral code with no typos. As a software developer, you should always think of making things better for everyone; People type their own name wrong every time, so why bother stressing them 🤷‍♂️

server/controllers/referral.controllers.js
```diff
@@ -0,0 +1,70 @@
...
+  updateReferralToRewarded,
+  getReferralByEmail,
+  sendReferralInvite,
+} from '../services/referral.service';
+import httpStatus from '../helpers/httpStatus';
+import EMAIL_CONTENT from '../../mailer';
+import { sendMail } from '../services/mailer.service';
+
+const ReferralController = {
+  getAllUserReferrals(req, res, next) {
```
andela-hpopoola, line 12: I am not sure of the main purpose of this that is not covered by the method below. We can convert this to load all the referrals for an admin. 

An admin need to see all the referrals so that they can make necessary payments for them. Please note that you will need to load necessary the referrer and referee information to make it easier for the admin to know who to pay when necessary.

server/controllers/referral.controllers.js
```diff
@@ -0,0 +1,70 @@
...
+    const referrerId = req.params.id;
+    getAllUserReferrals(referrerId)
+      .then((referrals) => {
+        res.status(httpStatus.OK).json({ success: true, referrals });
+      })
+      .catch((error) => next(error));
+  },
+
+  getReferrals(req, res, next) {
```
andela-hpopoola, line 21: I am not sure of your thought process for this but I think it is serving the same purpose as the method above. Lets make the name more descriptive as this is a better way to get the user referral:
```suggestion
  getUserReferrals(req, res, next) {
```

server/controllers/referral.controllers.js
```diff
@@ -0,0 +1,70 @@
...
+  sendInvite(req, res, next) {
+    const { email } = req.locals;
+    const referrerId = req.user._id;
+    sendReferralInvite({ referrerId, email })
+      .then((inviteInfo) => {
+        const contentTop = `${inviteInfo.referrerName} just invited you to BALLERS.`;
+        const contentBottom = `Use their code ${inviteInfo.referralCode}, or click the link below`;
+        sendMail(
+          EMAIL_CONTENT.REFERRAL_INVITE,
+          { email: inviteInfo.email, firstName: ' ' },
```
andela-hpopoola, line 57: We can get firstName from above, if it is not present then it can be null or undefined.

server/controllers/referral.controllers.js
```diff
@@ -0,0 +1,70 @@
...
+      .then((referral) => {
+        res.status(httpStatus.OK).json({ success: true, referral });
+      })
+      .catch((error) => next(error));
+  },
+
+  sendInvite(req, res, next) {
+    const { email } = req.locals;
+    const referrerId = req.user._id;
+    sendReferralInvite({ referrerId, email })
```
andela-hpopoola, line 51: We need to add firstName here so that we can save the firstName in the referral Model. This is necessary for creating a personalize interface for the referee when they visit the ballers website.

server/controllers/referral.controllers.js
```diff
@@ -0,0 +1,70 @@
...
+      .then((inviteInfo) => {
+        const contentTop = `${inviteInfo.referrerName} just invited you to BALLERS.`;
+        const contentBottom = `Use their code ${inviteInfo.referralCode}, or click the link below`;
+        sendMail(
+          EMAIL_CONTENT.REFERRAL_INVITE,
+          { email: inviteInfo.email, firstName: ' ' },
+          {
+            contentTop,
+            contentBottom,
+            link: `http://ballers.ng/referral/invite/${inviteInfo.referralCode}`,
```
andela-hpopoola, line 61: I believe this should reference the referralId not the referral code to make it easier to load the referral details from the database. 

This means that we will need an endpoint to load a referral via the referralId. 
I believe we can remove referral from the URL to make it shorter.

```suggestion
            link: `http://ballers.ng/invite/${inviteInfo._id}`,
```
See mockup below:

Screenshot 2020-09-20 at 8 12 22 PM

The referee name is gotten from the first name and the referrer name is gotten from the user model.

andela-hpopoola: For people visiting the site via the referral code that is shared by the user, they will see something like this. We will need another endpoint to cater for them. This endpoint will load the referrer info from the refCode.

Screenshot 2020-09-20 at 8 20 45 PM

server/models/referral.model.js
```diff
@@ -0,0 +1,63 @@
...
+    referrerId: {
+      type: ObjectId,
+      required: true,
+    },
+    reward: {
+      amount: {
+        type: Number,
+      },
+      status: {
```
andela-hpopoola, line 46: I think there is a mix up here. The reward status is different from the referral status. The reward status deals with Pending(default), In Progress and Paid.

The referral status deals with Invite Send, Registered and Sent. 

Here is the structure: 

```
reward : {
   status : {
   // reward status is here
    }
},
status: {
   //this is the referral status
}
```

server/models/user.model.js
andela-hpopoola, line 133: I dont think this is needed now, lets remove it 🙏

server/services/referral.service.js
```diff
@@ -0,0 +1,86 @@
...
+
+export const sendReferralInvite = async ({ referrerId, email }) => {
+  const referral = await getReferralByEmail(email).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (referral) {
+    throw new ErrorHandler(
+      httpStatus.PRECONDITION_FAILED,
+      'Multiple invites cannot be sent to same email',
```
andela-hpopoola, line 71: Hmmmm, actually multiple invites can be sent to the same email but they can't be sent from the same user. Let me explain:

1. Haruna sends an invite to Samuel
2. Nnamdi sends an invite to Samuel (this should work too despite the fact that Haruna has sent to Samuel). Since it is from different users, we are not spamming the user.
1. Haruna sends another invite to Samuel (this should trigger an error, has Haruna has already sent an invite to Samuel)
1. Leonard sends an invite to Samuel. This should work too despite the fact that Haruna and Nnamdi have sent an invite to the Samuel.

server/services/referral.service.js
```diff
@@ -0,0 +1,86 @@
...
+  } catch (error) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Referral not found', error);
+  }
+};
+
+export const getReferralByEmail = async (email, fields = null) =>
+  Referral.findOne({ email }).select(fields);
+
+export const sendReferralInvite = async ({ referrerId, email }) => {
+  const referral = await getReferralByEmail(email).catch((error) => {
```
andela-hpopoola, line 64: I believe this should check if the email exists with the referrer id (as explained below).

Also, we need to ensure that the email address has not been registered too. We don't want to sent email to registered users asking them to register via referral again.

andela-hpopoola (changes requested):
Thank you for working on this. You have done a great job, however, I think we need to add the following endpoint to make the referral workflow robust and satisfies most of referral requirements.

We will need to add the following endpoints:

1. An endpoint to get a referral information by referral Id 
2. An endpoint to get user information by refcode
3. An endpoint for load all referral for the admin (the one for user exists but we need one for the admin too to enable the admin to see the people he will pay).

After implementing the enpoints above, in another PR, we need to work on how to get the referral amount from a fixed referral percentage, assign this amount to the referrer and also link the property to the referrer when the referrer pays for his first property. This is still sketchy, so don't worry about this for now.

server/controllers/referral.controllers.js
tobieeniafe: this method gets the referrals a user has sent. that's why the user id is gotten from the token. the method above does the same thing, but it's used by an admin to ge a user's referrals by passing their id in the url

server/controllers/referral.controllers.js
tobieeniafe: for this, should I create an `invite` route? or it should be appended to the referral route?

andela-hpopoola (changes requested):
The current changes looks good. I think you still need to conclude the pending endpoints I mentioned earlier so that we can wrap this up asap. Thank you for the hard work.

server/routes/referral.routes.js
```diff
@@ -139,4 +139,50 @@ router.post(
...
+ *          schema:
+ *            $ref: '#/components/schemas/Referral'
+ *      description: Get a user's information by refcode
+ *     responses:
+ *      '200':
+ *        description: returns referral
+ *      '500':
+ *       description: Internal server error
+ */
+router.get('/refcode/:refcode', authenticate, ReferralController.getUserByRefCode);
```
andela-hpopoola, line 163: Doesn't really matter, but I feel `ref` is shorter and convenes the same meaning.
```suggestion
router.get('/ref/:refcode', authenticate, ReferralController.getUserByRefCode);
```

server/routes/referral.routes.js
```diff
@@ -139,4 +139,50 @@ router.post(
...
+ *          schema:
+ *            $ref: '#/components/schemas/Referral'
+ *      description: Get a referral by the id
+ *     responses:
+ *      '200':
+ *        description: returns referral
+ *      '500':
+ *       description: Internal server error
+ */
+router.get('/:id', authenticate, hasValidObjectId, ReferralController.getReferralById);
```
andela-hpopoola, line 186: Same comment as above. We dont need to authenticate this route.

server/routes/referral.routes.js
andela-hpopoola: Also, get referral by refcode and id should be public endpoint. Referees are unauthenticated so  you shouldn't authenticate before accessing them.

andela-hpopoola (commented):
This looks good to me. I need more time to test all the endpoints to be sure that they work as expected. Thank you for working no this.

server/services/user.service.js
```diff
@@ -84,6 +86,13 @@ export const addUser = async (user) => {
   try {
     const password = await hashPassword(user.password);
     const savedUser = await new User({ ...user, password, referralCode }).save();
+
+    if (user.referralCode) {
+      const referee = await getUserByReferralCode(user.referralCode).catch((error) => {
+        throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+      });
+      await addReferral({ userId: savedUser._id, refereeId: referee._id });
```
andela-hpopoola, line 94: This fails when I tested it. I believe it should be `referrerId`. Also can we have have a test that ensures that the referral table is created appropriate when they register.
```suggestion
      await addReferral({ userId: savedUser._id, referrerId: referrer._id });
```

server/services/user.service.js
```diff
@@ -84,6 +86,13 @@ export const addUser = async (user) => {
   try {
     const password = await hashPassword(user.password);
     const savedUser = await new User({ ...user, password, referralCode }).save();
+
+    if (user.referralCode) {
+      const referee = await getUserByReferralCode(user.referralCode).catch((error) => {
```
andela-hpopoola, line 91: The referee is the person that is currently registering. I believe we are getting the referrer here.

server/services/user.service.js
andela-hpopoola, line 94: When adding referral during registration we need to ensure that we do the following

Check if the referral is a DIRECT REFERRAL or INVITED REFERRAL. The truth is that, we don't need to depend on the frontend / another platform to verify this. The verification should be done by querying the referral table with the referrerId and email. If it exists, we should update the existing field else we should create a new one.

Also, we need to change the referral status to Registered here.

I think it is better to create a new method to handle this instead of reusing the addReferral method.

server/services/referral.service.js
```diff
@@ -0,0 +1,167 @@
...
+        referrerId: 1,
+        reward: 1,
+        status: 1,
+        'referee._id': 1,
+        'referee.email': 1,
+      },
+    },
+  ]);
+
+export const getAllReferrals = async () =>
```
andela-hpopoola, line 128: We need to return name, email and phone number of the referrer when an admin is viewing all referrals. This will be important to contact him/her for his reward.

server/services/referral.service.js
```diff
@@ -0,0 +1,167 @@
...
+        'referee.email': 1,
+      },
+    },
+  ]);
+
+export const updateReferralToRewarded = async (referralId) => {
+  try {
+    return Referral.findByIdAndUpdate(
+      referralId,
+      { $set: { status: REFERRAL_STATUS.REWARDED } },
```
andela-hpopoola, line 54: At this point, we will also need to set REWARD_STATUS to paid.

When the referee pays for his property, we need to update REWARD_STATUS to 'Progress' to denote that their reward is in progress. [This should be handled in a different PR]

server/routes/referral.routes.js
```diff
@@ -0,0 +1,188 @@
...
+ *      description: Get a referral by email
+ *     responses:
+ *      '200':
+ *        description: Referral found
+ *      '404':
+ *        description: Referral not found
+ *      '500':
+ *       description: Internal server error
+ */
+router.post('/email', schemaValidation(emailReferralSchema), ReferralController.getByEmail);
```
andela-hpopoola, line 112: Do you recall the use case for this? I am sorry, I can't think of the reason why we would want to get a referral by email (this can be problematic as we allow multiple email invites from different registered member).

If you can remember, can you state the reason here else can we remove it the enpoint please.

andela-hpopoola: Also, if we are getting by email, the method should be get not post.

```suggestion
router.get('/email', schemaValidation(emailReferralSchema), ReferralController.getByEmail);
```

server/routes/referral.routes.js
```diff
@@ -0,0 +1,188 @@
...
+ *      '200':
+ *        description: returns referral
+ *      '500':
+ *       description: Internal server error
+ */
+router.get('/ref/:refcode', ReferralController.getUserByRefCode);
+
+/**
+ * @swagger
+ * /referral/:id:
```
andela-hpopoola, line 167: The referrer id, firstName and lastName should be returned here too.  

Since this is a public endpoint, we need to be careful with the number of information we return.

andela-hpopoola (changes requested):
Phew, the review was mouthful 😫.

I have reviewed all the endpoints for referrals (Nice work). Please see comments and fix accordingly. Thank you and Nice work.

server/routes/referral.routes.js
tobieeniafe: I can't recall either, but i think you asked for it for some reason, i'd could remove it tho

andela-hpopoola: Please let's remove it. We can also come to this PR to copy the code if we later need it.

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+describe('Referral Controller', () => {
+  beforeEach(async () => {
+    adminToken = await addUser(adminUser);
+    userToken = await addUser(regularUser);
+  });
+
+  describe('Referral Invite Route', () => {
+    context('with valid data', () => {
+      it('returns successful invite', (done) => {
+        const invite = ReferralFactory.build({ email: 'invite-1@mail.com' });
```
andela-hpopoola, line 38: Referral Invite requires email and firstName, not sure why you didn't test for First Name. Also, we need to add a test to confirm that the first name is truly optional.

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+              expect(res).to.have.status(412);
+              expect(res.body.success).to.be.eql(false);
+              expect(res.body.message).to.be.eql('Validation Error');
+              expect(res.body.error).to.be.eql('"Email Address" is not allowed to be empty');
+              done();
+            });
+        });
+      });
+
+      context('when invite email is has been registered', () => {
```
andela-hpopoola, line 89:
```suggestion
      context('when invite email is a registered user', () => {
```

andela-hpopoola, line 89: Some other necessary tests here are
1. Test that a user cannot invite the same email more than once.
2. Test that another user can invite the same email that has been previously been invited by another user but hasn't registered

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+      context('with a valid token & id', () => {
+        it('returns successful payload', (done) => {
+          request()
+            .get('/api/v1/referral/all')
+            .set('authorization', adminToken)
+            .end((err, res) => {
+              expect(res).to.have.status(200);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body).to.have.property('referrals');
+              expect(res.body.referrals[0]).to.have.property('reward');
```
andela-hpopoola, line 140: Since you added just one referral above, I am surprised why you are using `have.property`. We have said it several times that this doesn't reduces the integrity of the test and should only be used when you cannot confirm that data. 

Test is more than just the coverage, there is no purpose on wasting time on tests if we are not testing for the right things. Can you test that the exact values are returned instead of testing if the property exists.

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+            .end((err, res) => {
+              expect(res).to.have.status(403);
+              expect(res.body.success).to.be.eql(false);
+              expect(res.body.message).to.be.eql('Token needed to access resources');
+              done();
+            });
+        });
+      });
+
+      context('when token is used', () => {
```
andela-hpopoola, line 166:
```suggestion
      context('when an invalid token is used', () => {
```

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+      });
+
+      context('when getAllReferrals service fails', () => {
+        it('returns the error', (done) => {
+          sinon.stub(Referral, 'aggregate').throws(new Error('Type Error'));
+          request()
+            .get('/api/v1/referral/all')
+            .set('authorization', adminToken)
+            .end((err, res) => {
+              expect(res).to.have.status(500);
```
andela-hpopoola, line 190: You should be able to test for the returned message too here.

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+        it('returns successful payload', (done) => {
+          request()
+            .get('/api/v1/referral/')
+            .set('authorization', userToken)
+            .end((err, res) => {
+              expect(res).to.have.status(200);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body.referrals.length).to.be.eql(2);
+              expect(res.body).to.have.property('referrals');
+              expect(res.body.referrals[0]).to.have.property('reward');
```
andela-hpopoola, line 235: Same property comment as above.

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+            .end((err, res) => {
+              expect(res).to.have.status(403);
+              expect(res.body.success).to.be.eql(false);
+              expect(res.body.message).to.be.eql('Token needed to access resources');
+              done();
+            });
+        });
+      });
+
+      context('when token is used', () => {
```
andela-hpopoola, line 259:
```suggestion
      context('when an invalid token is used', () => {
```

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+  describe('Get user information by referral code', () => {
+    const testId = mongoose.Types.ObjectId();
+    const referralCode = 'RC1234';
+    const referrer = UserFactory.build({ _id: testId, referralCode });
+
+    beforeEach(async () => {
+      await User.create(referrer);
+    });
+
+    context('with a valid referrl code', () => {
```
andela-hpopoola, line 301:
```suggestion
    context('with a valid referral code', () => {
```

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+      it('returns successful payload', (done) => {
+        request()
+          .get(`/api/v1/referral/ref/${referralCode}`)
+          .end((err, res) => {
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body).to.have.property('user');
+            expect(res.body.user._id).to.be.eql(testId.toString());
```
andela-hpopoola, line 309: Can you test that the user name and important user information is returned here too.

test/controllers/referral.controller.test.js
```diff
@@ -0,0 +1,390 @@
...
+            done();
+          });
+      });
+    });
+
+    context('with an invalid offer id', () => {
```
andela-hpopoola, line 315:
```suggestion
    context('with an invalid referral id', () => {
```

server/services/user.service.js
```diff
@@ -84,6 +86,17 @@ export const addUser = async (user) => {
   try {
     const password = await hashPassword(user.password);
     const savedUser = await new User({ ...user, password, referralCode }).save();
+
+    if (user.referralCode) {
```
andela-hpopoola, line 90: You need to add tests for this too.

test/services/referral.service.test.js
```diff
@@ -0,0 +1,278 @@
...
+          const currentcountedReferrals = await Referral.countDocuments({});
+          expect(err.statusCode).to.eql(500);
+          expect(err.message).to.be.eql('Internal Server Error');
+          expect(currentcountedReferrals).to.eql(countedReferrals);
+        }
+      });
+    });
+
+    // TODO fix tests for invite referral
+    // context('when a valid invite referral is entered', () => {
```
andela-hpopoola, line 58: You need to fix this too before we merge.

andela-hpopoola (commented):
Thank you for working on this, we are almost there. 

I focused more on the tests. Will do a final review later to be sure we didn't omit any important stuff.

andela-hpopoola (commented):
Thank you for making the changes. Everything works as expected. Can you wrap up the pending tests so that I can approve this asap 🙏

test/controllers/referral.controller.test.js
tobieeniafe: `aggregate` doesn't return an error message when stubbed, just a value

test/controllers/referral.controller.test.js
```diff
@@ -87,7 +103,64 @@ describe('Referral Controller', () => {
         });
       });
 
-      context('when invite email is has been registered', () => {
+      context('when invite email is not sent', () => {
```
andela-hpopoola, line 106:
```suggestion
      context('when no email is not given', () => {
```

test/controllers/referral.controller.test.js
```diff
@@ -87,7 +103,64 @@ describe('Referral Controller', () => {
...
+      context('when email that has been invited by another user but has not registered', () => {
+        const email = 'invite-1@mail.com';
+        beforeEach(async () => {
+          await sendReferralInvite({ email, referrerId: userId });
+        });
+        it('returns successful invite', (done) => {
+          const invite = { email, firstName: 'John' };
+          request()
+            .post('/api/v1/referral/invite')
+            .set('authorization', adminToken)
```
andela-hpopoola, line 132: Why are we inviting with an admin here? Can we invite with another user instead of an admin to replicate the exact case, please?

andela-hpopoola (approved):
Thank you for working on this. Can you add the referral test for someone registering with referral in `user.controller.test.js`. This is to ensure that registering with referral code doesn't break the normal registration flow and also confirm that the referral registration gives the same out as normal registration.

Thank you. APPROVED 🥇

PR #44, opened 2020-09-17
generate offer reference number

tobieeniafe:
#### What does this PR do?
Generate offer letter reference number

Generate unique reference number when offer letter is created 

`POST /api/v1/offer/create`

#174250401

### DIFF
--- a/server/helpers/dates.js
+++ b/server/helpers/dates.js
@@ -1,0 +1,5 @@
+import format from 'date-fns/format';
+
+export const TODAY_WORDS = format(new Date(), 'd MMM yyyy');
+
+export const TODAY_DDMMYY = format(new Date(), 'ddMMyyyy');
--- a/server/helpers/dates.js
+++ b/server/helpers/dates.js
@@ -3,3 +3,5 @@ import format from 'date-fns/format';
 export const getTodaysDateInWords = () => format(new Date(), 'd MMM yyyy');
 
 export const getTodaysDateShortCode = () => format(new Date(), 'ddMMyyyy');
+
+export const getTodaysDateStandard = () => format(new Date(), 'yyyy-MM-dd');
--- a/server/helpers/dates.js
+++ b/server/helpers/dates.js
@@ -1,5 +1,5 @@
 import format from 'date-fns/format';
 
-export const TODAY_WORDS = format(new Date(), 'd MMM yyyy');
+export const getTodaysDateInWords = () => format(new Date(), 'd MMM yyyy');
 
-export const TODAY_DDMMYY = format(new Date(), 'ddMMyyyy');
+export const getTodaysDateShortCode = () => format(new Date(), 'ddMMyyyy');
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -159,10 +159,6 @@ const UserSchema = new mongoose.Schema(
       default: false,
     },
     activationDate: Date,
-    vendorCode: {
-      type: String,
-      unique: true,
-    },
   },
   { timestamps: true },
 );
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -159,6 +159,10 @@ const UserSchema = new mongoose.Schema(
       default: false,
     },
     activationDate: Date,
+    vendorCode: {
+      type: String,
+      unique: true,
+    },
   },
   { timestamps: true },
 );
--- a/server/schemas/enquiry.schema.js
+++ b/server/schemas/enquiry.schema.js
@@ -1,5 +1,5 @@
 import Joi from '@hapi/joi';
-import { TODAY_WORDS } from '../helpers/dates';
+import { getTodaysDateInWords } from '../helpers/dates';
 
 Joi.objectId = require('joi-objectid')(Joi);
 
@@ -27,11 +27,11 @@ export const addEnquirySchema = Joi.object({
   initialInvestmentAmount: Joi.number().label('Initial Investment Amount').required(),
   periodicInvestmentAmount: Joi.number().label('Periodic Investment Amount').required(),
   investmentStartDate: Joi.date()
-    .greater(TODAY_WORDS)
+    .greater(getTodaysDateInWords())
     .label('Investment Start Date')
     .required()
     .messages({
-      'date.greater': `"Investment Start Date" should a date later than ${TODAY_WORDS}`,
+      'date.greater': `"Investment Start Date" should a date later than ${getTodaysDateInWords()}`,
     }),
 });
 
--- a/server/schemas/enquiry.schema.js
+++ b/server/schemas/enquiry.schema.js
@@ -1,9 +1,7 @@
 import Joi from '@hapi/joi';
-import format from 'date-fns/format';
+import { TODAY_WORDS } from '../helpers/dates';
 
 Joi.objectId = require('joi-objectid')(Joi);
-
-const today = format(new Date(), 'd MMM yyyy');
 
 const address = Joi.object().keys({
   street1: Joi.string().label('Street 1').required(),
@@ -29,11 +27,11 @@ export const addEnquirySchema = Joi.object({
   initialInvestmentAmount: Joi.number().label('Initial Investment Amount').required(),
   periodicInvestmentAmount: Joi.number().label('Periodic Investment Amount').required(),
   investmentStartDate: Joi.date()
-    .greater(today)
+    .greater(TODAY_WORDS)
     .label('Investment Start Date')
     .required()
     .messages({
-      'date.greater': `"Investment Start Date" should a date later than ${today}`,
+      'date.greater': `"Investment Start Date" should a date later than ${TODAY_WORDS}`,
     }),
 });
 
--- a/server/schemas/enquiry.schema.js
+++ b/server/schemas/enquiry.schema.js
@@ -1,5 +1,5 @@
 import Joi from '@hapi/joi';
-import { getTodaysDateInWords } from '../helpers/dates';
+import { getTodaysDateInWords, getTodaysDateStandard } from '../helpers/dates';
 
 Joi.objectId = require('joi-objectid')(Joi);
 
@@ -27,7 +27,7 @@ export const addEnquirySchema = Joi.object({
   initialInvestmentAmount: Joi.number().label('Initial Investment Amount').required(),
   periodicInvestmentAmount: Joi.number().label('Periodic Investment Amount').required(),
   investmentStartDate: Joi.date()
-    .greater(getTodaysDateInWords())
+    .greater(getTodaysDateStandard())
     .label('Investment Start Date')
     .required()
     .messages({
--- a/server/schemas/offer.schema.js
+++ b/server/schemas/offer.schema.js
@@ -1,27 +1,27 @@
 import Joi from '@hapi/joi';
-import { TODAY_WORDS } from '../helpers/dates';
+import { getTodaysDateInWords } from '../helpers/dates';
 
 Joi.objectId = require('joi-objectid')(Joi);
 
 export const createOfferSchema = Joi.object({
   enquiryId: Joi.objectId().label('Enquiry ID').required(),
   handOverDate: Joi.date()
-    .greater(TODAY_WORDS)
+    .greater(getTodaysDateInWords())
     .label('Handover Date')
     .required()
     .messages({
-      'date.greater': `"Handover Date" should a date later than ${TODAY_WORDS}`,
+      'date.greater': `"Handover Date" should a date later than ${getTodaysDateInWords()}`,
     }),
   deliveryState: Joi.string().label('Delivery State').required(),
   totalAmountPayable: Joi.number().label('Total Amount Payable').required(),
   allocationInPercentage: Joi.number().min(1).max(100).label('Allocation In Percentage').required(),
   title: Joi.string().label('Title').required(),
   expires: Joi.date()
-    .greater(TODAY_WORDS)
+    .greater(getTodaysDateInWords())
     .label('Expiry Date')
     .required()
     .messages({
-      'date.greater': `"Expiry Date" should a date later than ${TODAY_WORDS}`,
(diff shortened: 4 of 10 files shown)


test/controllers/offer.controller.test.js
```diff
@@ -52,7 +65,8 @@ describe('Offer Controller', () => {
             expect(res.body.offer).to.have.property('userInfo');
             expect(res.body.offer.vendorInfo._id).to.be.eql(adminId.toString());
             expect(res.body.offer.vendorInfo._id).to.be.eql(adminId.toString());
-            expect(res.body.offer.propertyInfo._id).to.be.eql(propertyId.toString());
+            expect(res.body.offer.propertyInfo._id).to.be.eql(createPropertyId.toString());
+            expect(res.body.offer.referenceCode.substr(0, 9)).to.be.eql('LVE/OLM/1');
```
andela-hpopoola, line 69: Can you search for a way to stub Date.now so that you can use it in this test. We need to ensure that we test for everything not a substring.

andela-hpopoola, line 69: From the mail sent by Nnamdi, the number sold should be a minimum of 2 digits. Here is an excerpt from the mail. Simply write a function in the helper file to fix this.

```
If its the third maisonette sold that day - 03
```
```suggestion
            expect(res.body.offer.referenceCode.substr(0, 9)).to.be.eql('LVE/OLM/01XXXXXX');
```

server/services/offer.service.js
```diff
@@ -1,15 +1,36 @@
...
 export const getOfferById = async (id) => Offer.findById(id).select();
 
+export const generateReferenceCode = async (propertyId) => {
+  const property = await getPropertyById(propertyId);
+  const initials = property.name
+    .match(/\b(\w)/g)
+    .join('')
+    .toUpperCase();
+  const numberSold = await Offer.countDocuments({ propertyId })
+    .then((count) => {
```
andela-hpopoola, line 22: This can be converted into 1 line.

```suggestion
    .then((count) => count + 1;
```

server/services/offer.service.js
```diff
@@ -1,15 +1,36 @@
...
 import { getEnquiryById, approveEnquiry } from './enquiry.service';
+import { getPropertyById } from './property.service';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
 export const getOfferById = async (id) => Offer.findById(id).select();
 
+export const generateReferenceCode = async (propertyId) => {
+  const property = await getPropertyById(propertyId);
+  const initials = property.name
```
andela-hpopoola, line 17: This should be extracted into its own function so that it is easier to know what it does, reuse later if needed and if possible make it easier for us to write test for it.

andela-hpopoola (changes requested):
Thank you for working on this. My major concern is ensuring that we test that the expected date is returned with the reference code and not testing for the substring.

server/services/offer.service.js
tobieeniafe: extracting this to its own function would make us query the database twice for the same thing due to line 28

andela-hpopoola: I am sorry, I dont understand how extracting this as a function will affect the database. It has nothing to do with the database.

 ```suggestion
  const initials = getPropertyInitial(property.name)
```

```
const getPropertyInitial = (name) => name 
     .match(/\b(\w)/g)
    .join('')
    .toUpperCase();
```

File: server/helpers/dates.js [hunk 4]
```diff
@@ -0,0 +1,5 @@
+import format from 'date-fns/format';
+
+export const TODAY_WORDS = format(new Date(), 'd MMM yyyy');
```
andela-hpopoola, line 3: TODAY_WORDS doesn't make any sense to someone reading the code for the first time as it has multiple meanings.

Also, the date will change everyday so it is a bad idea to make it a constant. A function is more preferable.

```suggestion
export const getTodaysDateInWord = () => format(new Date(), 'd MMM yyyy');
```

server/services/offer.service.js
tobieeniafe: ohhh. I was thought process was very different. Would fix it

server/helpers/dates.js
```diff
@@ -0,0 +1,5 @@
...
+
+export const TODAY_DDMMYY = format(new Date(), 'ddMMyyyy');
```
andela-hpopoola, line 5: It is a bad idea to put a format type in the variable name. This is not advisable because the name will be incorrect if we decide to change the format tomorrow. Also, no one wants to memorize the code type when they are using the function.

This is better as a function than a constant. The date changes everyday, so it is not a constant. 

```suggestion
export const getTodaysDateShortCode = () => format(new Date(), 'ddMMyyyy');
```

server/services/offer.service.js
andela-hpopoola: Extracting it in a different function makes the code easier to read for anyone that wants to work on the code later in the future. All of this will not make sense to you now since you are working on the code from scratch, if you are working on legacy codes (i.e. codes that have been written by different people), all of this will make more sense to you.

server/services/offer.service.js
```diff
@@ -5,11 +5,29 @@ import httpStatus from '../helpers/httpStatus';
...
+    .match(/\b(\w)/g)
+    .join('')
+    .toUpperCase();
+  const numberSold = await Offer.countDocuments({ propertyId })
+    .then((count) => count + 1)
+    .catch((error) => {
+      throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+    });
+  const type = `OL${property.houseType.charAt(0)}`;
+  const referenceCode = `${initials}/${type}/0${numberSold}${TODAY_DDMMYY}`;
```
andela-hpopoola, line 27: This is invalid. This will cause a bug when the number sold is greater or equal to 10 😞 . You need to always think for edge cases before writing your code.

You need to write a function to check if the number is less than 10, append 0 else just output the number. This should be done before you concatenate everything together here.

```suggestion
  const referenceCode = `${initials}/${type}/${numberSold}${TODAY_DDMMYY}`;
```

test/services/offer.service.test.js
```diff
@@ -47,6 +49,62 @@ describe('Offer Service', () => {
...
 
+  describe('#generateReferenceCode', () => {
+    const userId = mongoose.Types.ObjectId();
+    const propertyId = mongoose.Types.ObjectId();
+    const enquiryId1 = mongoose.Types.ObjectId();
+    const enquiryId2 = mongoose.Types.ObjectId();
+    const offerId = mongoose.Types.ObjectId();
+
+    const user = UserFactory.build({ _id: userId });
+    const property = PropertyFactory.build({
```
andela-hpopoola, line 60: Can you write more tests to be 100% certain that your code covers all the edge cases we raised in the email trail:

Please go through the mail again for better context, When in doubt, you can reply to the email trail or ask Nnamdi:

Here are some 

- How do we cater for titles like '3 Bedroom Flat Lekkiville Estates'? Will the generated code be 3BFL or BFLE? [LE for 3 bedroom flat Lekkiville Estate, so with Offer letter will be OL3BLE] 

- [Edge case] Do we have a maximum length of generated code here or the generation will be from all the words in the property name.
E.g If a developer writes something like '4 Bedroom Flat with maid room in Lekkiville Estates', what will be the generated code? Without a maximum length, the generated code will be '4BFWMRILE'.
If there is a maximum, what should be the generated code? [Kindly disregard the maids room, please leave as 3 above. 

- Lastly, the digital REF code doesn't have any reference to the developer that owns the property. I believe we should add the developer's SHORTCODE to differentiate each REF code. e.g. HIG for Highrachy, DEN for Denim Properties, DEN2 for Deno Properties (2 is added to the ref here since we already have DEN for Denim properties). If we are adding this, please let us know how it affects the current implementation. [Pls add the developer before the estate shortened name as suggested, thanks.]

andela-hpopoola (changes requested):
Thank you for working on this. I left more comments on writing tests to ensure your code satisfies all the edge cases raised in the email trail. Thank you.

test/services/offer.service.test.js
tobieeniafe: 1. With the current implementation, `PropertyName/PropertyType/NumberSoldDate` `3 Bedroom Flat Lekkiville Estates` would be generated as `LE/OL3/...`.
2. Property name should be different from description. `4 Bedroom Flat with maid room in Lekkiville Estates` seems more like the description than the name.
3. We'd have to pick a way to generate a developer's shortcode first. Appending it to the reference code shouldn't be an issue.

andela-hpopoola: Thank you for the comments!

1. Please confirm this with Nnamdi to be 100% sure and report his response here. His response will affect the implementation of 2 below:

2. Unfortunately, we are not in control of that. Even if it looks like a description, we cannot dictate what a vendor will enter as a title. Moreover, using `4 Bedroom Flat with maid room in Lekkiville Estates` as the title name will give no error when you create a property, so it is better it.

The main purpose of the question is to verify how the property name shortcode will be generated if the property shortcode has over 6 letters. You need to be sure if we would truncate it (if yes, what should be the maximum length -- 3, 4 or 5) or should we use everything. This is a little 

This example might look like a description, here are more examples that have the same effect but don't look like a description
1. Jide Taiwo & Co Lekki Estate
2. Le Moon Song Chinese Penthouse
3. The Lord is my Shepherd Houses (Annex)

3. I think you are overthinking this (we have discussed issues like this several times). As long as we dont have the vendor information yet, you should use a default code for Highrachy / Ballers (confirm from Nnamdi and report his response here). Developer's shortcode will be generated when the vendor account is created, so you need to use a constant with this. When this is done, the shortcode will be replaced accordingly. This should be added before the estate shortened name.

server/schemas/enquiry.schema.js
```diff
@@ -27,11 +27,11 @@ export const addEnquirySchema = Joi.object({
   initialInvestmentAmount: Joi.number().label('Initial Investment Amount').required(),
   periodicInvestmentAmount: Joi.number().label('Periodic Investment Amount').required(),
   investmentStartDate: Joi.date()
-    .greater(TODAY_WORDS)
+    .greater(getTodaysDateInWords())
```
andela-hpopoola, line 30: I think this will require a proper date not the date in words. Can you verify?

andela-hpopoola (commented):
Still waiting for Nnamdi's response on some of the comments. I have you have contacted him.

server/models/user.model.js
```diff
@@ -159,6 +159,10 @@ const UserSchema = new mongoose.Schema(
       default: false,
     },
     activationDate: Date,
+    vendorCode: {
```
andela-hpopoola, line 162: Not sure what you are thinking but I don't think we should have a vendor code yet. It will make our work complicated later when we are ready to add the vendor information to the user model. 

We could use Highrachy/Ballers vendorcode instead of the vendor code. This will be easier to manage since we will just remove the reference only.

server/services/offer.service.js
```diff
@@ -5,30 +5,31 @@ import httpStatus from '../helpers/httpStatus';
...
-import { getPropertyById } from './property.service';
+import { getOneProperty } from './property.service';
 import { getTodaysDateShortCode } from '../helpers/dates';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
 export const getOfferById = async (id) => Offer.findById(id).select();
 
-export const getPropertyInitials = (propertyName) =>
+export const getInitials = (propertyName) =>
```
andela-hpopoola, line 15:
```suggestion
export const getPropertyInitials = (propertyName) =>
```

andela-hpopoola (changes requested):

server/models/user.model.js
tobieeniafe: I don't understand the 2nd comment

andela-hpopoola: Confirm Highrachy shortcode from Nnamdi (you used HIG in the test) and use it as the vendorcode since we have not implemented the vendor features yet.

```js
const VENDOR_CODE = 'HIG';
...
const referenceCode = `${VENDOR_CODE}/...`;

```

server/schemas/offer.schema.js
```diff
@@ -1,29 +1,27 @@
...
 
 Joi.objectId = require('joi-objectid')(Joi);
 
-const today = format(new Date(), 'd MMM yyyy');
-
 export const createOfferSchema = Joi.object({
   enquiryId: Joi.objectId().label('Enquiry ID').required(),
   handOverDate: Joi.date()
-    .greater(today)
+    .greater(getTodaysDateInWords())
```
andela-hpopoola, line 9: Can you add a test to be sure that this works as expected? I overlooked some of your tests because of time but in the ideal world, your test should cover all cases.

You added only one test to confirm if it is empty, we need another one to test that it disallows today's date.
https://github.com/Highrachy/ballers-api/blob/e86cf126f1d7b05b8bcb9135b898e41470b0dcd4/test/controllers/offer.controller.test.js#L130

server/services/offer.service.js
```diff
@@ -21,14 +21,14 @@ export const getInitials = (propertyName) =>
 export const generateReferenceCode = async (propertyId) => {
   const property = await getOneProperty(propertyId);
   const { vendorCode } = property[0].adminInfo;
```
andela-hpopoola, line 23: Still expecting you to update this? Why are you finding it difficult to update -- you are wasting more time implementing feedback without giving any tangible reason. This is making you spend more than enough time on simple features.

Make the changes and lets move on to other stories, if you don't agree with the comment, state your concerns and lets move on.

andela-hpopoola (changes requested):
Can you pay more attention to the comments and let's proceed with other stories. You are spending too much time on implementing feedback, let get this done and proceed please 🙏

andela-hpopoola (approved):
Thank you for working on this. This works great when I tested it. 

I noticed another edge case that a user can make multiple enquiries for the same property. We should only allow a user to make JUST ONE enquiry for EACH PROPERTY. This should be a quick fix in another PR 🙏 ?

Feel free to merge when ready. Thank you.

PR #46, opened 2020-09-23
prevent user from making more than one enquiry per property

tobieeniafe:
#### What does this PR do?
Prevent user from making more than one enquiry per property

Add functionality to add enquiry endpoint to prevent user from making more than one enquiry per property

`POST /api/v1/enquiry/add`

### DIFF
--- a/test/controllers/enquiry.controller.test.js
+++ b/test/controllers/enquiry.controller.test.js
@@ -59,6 +59,35 @@ describe('Add Enquiry Route', () => {
     });
   });
 
+  context('when an enquiry by the user for the property exists', () => {
+    const propertyId = mongoose.Types.ObjectId();
+    const enquiry = EnquiryFactory.build({
+      propertyId,
+    });
+    beforeEach(async () => {
+      await addEnquiry(
+        EnquiryFactory.build({
+          propertyId,
+          userId: _id,
+          addedBy: _id,
+          updatedBy: _id,
+        }),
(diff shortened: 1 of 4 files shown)


test/controllers/enquiry.controller.test.js
```diff
@@ -59,6 +59,35 @@ describe('Add Enquiry Route', () => {
...
   });
 
+  context('when an enquiry by the user for the property exists', () => {
+    const propertyId = mongoose.Types.ObjectId();
+    const enquiry = EnquiryFactory.build({
+      propertyId,
+    });
+    beforeEach(async () => {
+      await addEnquiry(
+        EnquiryFactory.build({
```
andela-hpopoola, line 69: I am curious to why you don't want to add the same enquiry above so that we don't need to create another one here.

```
 const enquiry = EnquiryFactory.build({
      propertyId,
      userId: _id,
  });

beforeEach(async () => {
      await addEnquiry(enquiry)
   ...
```

andela-hpopoola (approved):
This looks good to me. Just left a single non blocking nit. Feel free to merge after fixing. Great job 👍

test/controllers/enquiry.controller.test.js
tobieeniafe: this is using the service directly, while the other is using the route. So it would basically throw a validation error

PR #53, opened 2020-10-06
#175070809 remove assigned properties when searching

tobieeniafe:
#### What does this PR do?
Remove assigned properties when searching

Remove properties that have been assigned to a user from results when searching for properties

`POST /api/v1/property/search`

#175070809


server/services/property.service.js
```diff
@@ -190,8 +190,16 @@ export const getOneProperty = async (propertId) =>
...
+export const searchThroughProperties = async ({
+  userId,
+  state,
+  city,
+  houseType,
+  minPrice,
+  maxPrice,
+}) =>
   Property.aggregate([
+    { $match: { assignedTo: { $nin: [ObjectId(userId)] } } },
```
andela-hpopoola, line 202: Nice work here. Currently, the assignedTo field is not populated by any of the endpoint so this might not work as expected. We might need to update an existing endpoint to make this work accordingly.

Property assignment should happen when the user has signed the offer letter form, so we might want to add the user id to the property when they have signed the offer letter and remove the id when the offer letter is rejected or neglected (not sure if we have implemented this yet, just adding it as a food for thought).

test/controllers/property.controller.test.js
```diff
@@ -1293,6 +1294,7 @@ describe('Property Controller', () => {
               expect(res.body.properties[0]).to.have.property('houseType');
               expect(res.body.properties[0]).to.have.property('description');
               expect(res.body.properties[0]._id).to.be.eql(property._id.toString());
+              expect(res.body.properties[0].assignedTo).to.not.include(regularUserId.toString());
               done();
             });
         });
```
andela-hpopoola, line 1300: The test above doesn't actually test for assigned properties, it will work for assigned and unassigned properties. We need to add a valid test to be sure that this works accordingly. 

You can add multiple properties and test that it returns the following with the same search parameter
- all the properties if the user is not assigned any property
- only unassigned properties if the user is assigned to one
- no property if the user is assigned to all the properties.

andela-hpopoola (changes requested):
Thank you for working on this. I love the implementation you used even though I had a different thing in mind. This will work too if we ensure we manage it well.

Please don't forget to write tests to verify that properties are properly assigned when a user accepts their offer letter. Thank you.

test/controllers/property.controller.test.js
```diff
@@ -1411,27 +1415,48 @@ describe('Property Controller', () => {
...
       });
+
+      context('when all properties have been assigned', () => {
```
andela-hpopoola, line 1419: https://github.com/Highrachy/ballers-api/pull/53/commits/dea670427c94941ae24afb20a5edc89e933072ab#r500568732

From my previous comment, you will need to add three tests to properly test for assigned properties and ensure we covered all the bases. The three tests should use the same search parameter to ensure that they are properly implemented.

This is just one out of the 3 tests, can you add the remaining 2 tests to improve our confidence.

This one is testing for when all the properties are assigned to the user. The remaining 2 tests are:

- return all the properties if the user is not assigned any property
- return only unassigned properties if the user is assigned to one

andela-hpopoola (changes requested):
Thank you for working on this. Most of the tests I requested from my last comments are still missing, can you add them so that we can close this PR, please?

> Please don't forget to write tests to verify that properties are properly assigned when a user accepts their offer letter.

The current implementation will not work as expected if we are not adding the userId to the property. That is why the above test is very important. Without this, the assignedTo array will always be empty and there is be no way to know if a user has been assigned to a property or not.

andela-hpopoola (changes requested):

andela-hpopoola:
Any updates on this. Can you fix the pending comments so that we can merge this, we are wasting too much time 🙏 

> Looking at the implementation of assignPropertyToUser, I think it is the wrong thing to use here. There is a section where it reduces the units of the property, this is not the intended behaviour, as properties are not TOTALLY ASSIGNED until payment has been made to a certain threshold.
> 
> In the current implementation, we are assigning people to the properly even if they are INTERESTED (i.e they have signed the offer letter). In the real world, INTERESTED doesn't relate to assigning the property to them as they could change their mind later.
> 
> Can you create another service to do the assignment without the bell and whistles of assignPropertyToUser`. Also, can we add tests to ensure that the property has been properly assigned.

server/services/user.service.js
```diff
@@ -222,6 +221,45 @@ export const assignPropertyToUser = async (toBeAssigned) => {
   }
 };
 
+// TODO this is commented out as user model would need an `allocatedTo` to cater to allocated properties and separate them from assigned properties
```
andela-hpopoola, line 224: Thinking of it again, I think we can use the above service for this too but by passing an additional variable to determine if the units will be removed or not. We should be able to get rid of this. However, we will need to write tests to ensure that the services work for the 2 scenarios as intended.

andela-hpopoola (changes requested):

andela-hpopoola (changes requested):
I think it is safer to remove the code instead of commenting them out. We can always revisit the commit that removed the code and copy and paste it back. I don't want us to feel pressured in implementing the property assignment now and we might also need to update the implementation so don't feel bad or unproductive about removing the entire code.

server/services/user.service.js
andela-hpopoola, line 224: Please remove this too entirely.

server/routes/user.routes.js
```diff
@@ -194,43 +194,43 @@ router.post(
...
- *       description: Internal server error
- */
-router.post(
-  '/assign-property',
-  authenticate,
-  isAdmin,
-  schemaValidation(assignPropertySchema),
-  UserController.assignProperty,
-);
+// /**
```
andela-hpopoola, line 197: Can we remove this too, please? We can always come to this PR and copy it back.

test/controllers/user.controller.test.js
```diff
@@ -888,100 +888,100 @@ describe('User Controller', () => {
...
-          .end((err, res) => {
-            expect(res).to.have.status(400);
-            expect(res.body.success).to.be.eql(false);
-            done();
-            User.findByIdAndUpdate.restore();
-          });
-      });
-    });
-  });
+  // describe('Assign property route', () => {
```
andela-hpopoola, line 891: Can we remove this one too 🙏

andela-hpopoola (changes requested):
Thank you for working on this. I will test this locally later but in the meantime, you can remove all the commented codes so that we can add them later. The implementation might even be different when we want to add them later so don't feel bad about removing them now.

andela-hpopoola (changes requested):
This works as expected. Thank you for working on this, however, I still don't feel confidence about the assignPropertyToUser service. I have added my comments to the code:

```js

export const assignPropertyToUser = async (toBeAssigned) => {
  const assignedProperty = {
    propertyId: toBeAssigned.propertyId,
    assignedBy: toBeAssigned.assignedBy, // this is not needed as properties are assigned automatically
    assignedDate: Date.now(), // this is not needed as it can be gotten from offer
  };
  const property = await getPropertyById(toBeAssigned.propertyId).catch((error) => {
  });

  if (property.units < 1) { // this should be added when an enquiry or offer letter is made
    throw new ErrorHandler(httpStatus.NOT_FOUND, 'No available units');
  }

  const owner = await getUserById(toBeAssigned.userId).catch((error) => {
  });

  try {
    const updatedProperty = {
      id: property.id,
      $push: { assignedTo: toBeAssigned.userId },
    };

    const updatePropertyUnit = await updateProperty(updatedProperty).catch((error) => {
    });

     // this is no longer needed as we can load assigned properties from offers
    if (updatePropertyUnit) {
      return User.findByIdAndUpdate(owner.id, { $push: { assignedProperties: assignedProperty } });
    }

    return new ErrorHandler(httpStatus.BAD_REQUEST, 'Error assigning property');
  } catch (error) {
    throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error assigning property', error);
  }
};

```

tobieeniafe:
> This works as expected. Thank you for working on this, however, I still don't feel confidence about the assignPropertyToUser service. I have added my comments to the code:
> 
> ```js
> export const assignPropertyToUser = async (toBeAssigned) => {
>   const assignedProperty = {
>     propertyId: toBeAssigned.propertyId,
>     assignedBy: toBeAssigned.assignedBy, // this is not needed as properties are assigned automatically
>     assignedDate: Date.now(), // this is not needed as it can be gotten from offer
>   };
>   const property = await getPropertyById(toBeAssigned.propertyId).catch((error) => {
>     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
>   });
> 
>   if (property.units < 1) { // this should be added when an enquiry or offer letter is made
>     throw new ErrorHandler(httpStatus.NOT_FOUND, 'No available units');
>   }
> 
>   const owner = await getUserById(toBeAssigned.userId).catch((error) => {
>     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
>   });
> 
>   try {
>     const updatedProperty = {
>       id: property.id,
>       $push: { assignedTo: toBeAssigned.userId },
>     };
> 
>     const updatePropertyUnit = await updateProperty(updatedProperty).catch((error) => {
>       throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
>     });
> 
>      // this is no longer needed as we can load assigned properties from offers
>     if (updatePropertyUnit) {
>       return User.findByIdAndUpdate(owner.id, { $push: { assignedProperties: assignedProperty } });
>     }
> 
>     return new ErrorHandler(httpStatus.BAD_REQUEST, 'Error assigning property');
>   } catch (error) {
>     throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error assigning property', error);
>   }
> };
> ```

could you write the service in the presumed way, because removing lines where you highlighted renders the service useless more or less

server/services/user.service.js
```diff
@@ -202,24 +194,14 @@ export const assignPropertyToUser = async (toBeAssigned) => {
...
-    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
-  });
-
   try {
     const updatedProperty = {
       id: property.id,
       $push: { assignedTo: toBeAssigned.userId },
     };
-    const updatePropertyUnit = await updateProperty(updatedProperty).catch((error) => {
+    return await updateProperty(updatedProperty).catch((error) => {
```
andela-hpopoola, line 202: I don't think it is a good idea to return await. Lets return the value of the await.
```suggestion
    const updatePropertyUnit = await updateProperty(updatedProperty).catch((error) => {
    ...
    
    return updatePropertyUnit;
```

andela-hpopoola (approved):
I left a non blocking comment. Feel free to merge this later. Thank you and nice work.

PR #54, opened 2020-10-08
#175160638 add offer id to transaction model

tobieeniafe:
#### What does this PR do?
Add offer id to transaction model

Add offer id to transaction model

#175160638

### DIFF
--- a/test/controllers/transaction.controller.test.js
+++ b/test/controllers/transaction.controller.test.js
@@ -77,6 +77,22 @@ describe('Transaction Controller', () => {
             });
         });
       });
+      context('when user id is empty', () => {
+        it('returns an error', (done) => {
+          const transaction = TransactionFactory.build({ offerId: '' });
+          request()
+            .post('/api/v1/transaction/add')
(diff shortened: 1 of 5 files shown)


test/controllers/transaction.controller.test.js
```diff
@@ -77,6 +77,22 @@ describe('Transaction Controller', () => {
             });
         });
       });
+      context('when user id is empty', () => {
```
andela-hpopoola, line 80:
```suggestion
      context('when offer id is empty', () => {
```

andela-hpopoola (approved):

From a closer observation, this would cause so unintended updates to some endpoints that is consumed in the frontend as we would need to include the offer details of a property (as the offerId is now compulsory) because a transaction can be made. This is easier to test via the endpoint as you would just copy and paste the offerId, however, in the frontend, all of this information must be readily available before a valid transaction can be made.

Let's watch out for the affected endpoints and fix them in another PR. Feel free to merge this.

PR #56, opened 2020-10-09
#175070906 load assigned properties endpoint

tobieeniafe:
#### What does this PR do?
Add endpoints to load assigned properties which include additional information such as transactions and sums

Add endpoints to load assigned properties which include additional information such as transactions and sums

`GET /property/assigned`
`GET /property/assigned/:id`

#175070906

### DIFF
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -1,9 +1,9 @@
+/* eslint-disable import/no-cycle */
 import mongoose from 'mongoose';
 import Offer from '../models/offer.model';
 import { ErrorHandler } from '../helpers/errorHandler';
 import httpStatus from '../helpers/httpStatus';
 import { OFFER_STATUS } from '../helpers/constants';
-// eslint-disable-next-line import/no-cycle
 import { getUserById } from './user.service';
 import { getEnquiryById, approveEnquiry } from './enquiry.service';
 import { getOneProperty } from './property.service';
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -1,11 +1,12 @@
-/* eslint-disable import/no-cycle */
 import mongoose from 'mongoose';
 import Offer from '../models/offer.model';
 import { ErrorHandler } from '../helpers/errorHandler';
 import httpStatus from '../helpers/httpStatus';
 import { OFFER_STATUS } from '../helpers/constants';
+// eslint-disable-next-line import/no-cycle
 import { getUserById } from './user.service';
 import { getEnquiryById, approveEnquiry } from './enquiry.service';
+// eslint-disable-next-line import/no-cycle
 import { getOneProperty } from './property.service';
 import { getTodaysDateShortCode } from '../helpers/dates';
 
@@ -207,6 +208,14 @@ export const getOffer = async (offerId) =>
         localField: 'propertyId',
         foreignField: '_id',
         as: 'propertyInfo',
+      },
+    },
+    {
+      $lookup: {
+        from: 'transactions',
+        localField: '_id',
+        foreignField: 'offerId',
+        as: 'transactionInfo',
       },
     },
     {
--- a/server/services/property.service.js
+++ b/server/services/property.service.js
@@ -307,9 +307,10 @@ export const getAssignedPropertyByOfferId = async (offerId) => {
 
 export const getAssignedProperties = async (userId) =>
   Offer.aggregate([
+    { $match: { userId: ObjectId(userId) } },
     {
       $match: {
-        $and: [{ userId: ObjectId(userId) }, { status: OFFER_STATUS.ASSIGNED }],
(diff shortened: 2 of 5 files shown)


server/services/offer.service.js
```diff
@@ -1,9 +1,9 @@
+/* eslint-disable import/no-cycle */
```
andela-hpopoola, line 1: It is more preferable to add the eslint to the affected line instead of the full page. Adding it to the affected line makes us know the line we are disabling eslint for and makes it easier to fix later. Disabling for the whole page can easily introduce unintended bug.

server/services/property.service.js
```diff
@@ -265,3 +268,40 @@ export const getAvailablePropertyOptions = async () =>
...
+export const getAssignedPropertyById = async (offerId) => {
+  const total = await getTotalAmountPaidForProperty(offerId);
+  const totalPaid = total.length > 0 ? total[0].totalAmountContributed : 0;
+
+  const offer = await getOfferById(offerId);
+
+  const property = await getPropertyById(offer.propertyId);
+
+  return {
+    totalAmountPayable: offer.totalAmountPayable,
```
andela-hpopoola, line 303: Looking at this, I think it is more preferable to return the offer to cater for future UI changes instead of the `totalAmountPayable` only. For example, offerId is needed in the frontend to make the links work appropriately but it is missing.

```suggestion
    offer,
```

server/services/property.service.js
```diff
@@ -265,3 +268,40 @@ export const getAvailablePropertyOptions = async () =>
...
+    },
+  ]);
+
+export const getAssignedPropertyById = async (offerId) => {
+  const total = await getTotalAmountPaidForProperty(offerId);
+  const totalPaid = total.length > 0 ? total[0].totalAmountContributed : 0;
+
+  const offer = await getOfferById(offerId);
+
+  const property = await getPropertyById(offer.propertyId);
```
andela-hpopoola, line 300: I think this can be merged into the query above.

server/services/property.service.js
```diff
@@ -265,3 +268,40 @@ export const getAvailablePropertyOptions = async () =>
...
+        totalAmountContributed: { $sum: '$amount' },
+      },
+    },
+  ]);
+
+export const getAssignedPropertyById = async (offerId) => {
+  const total = await getTotalAmountPaidForProperty(offerId);
+  const totalPaid = total.length > 0 ? total[0].totalAmountContributed : 0;
+
+  const offer = await getOfferById(offerId);
```
andela-hpopoola, line 298: I believe this is more preferable

https://github.com/Highrachy/ballers-api/blob/09e6938539f585b1d825fc10f28b2aa46e25af99/server/services/offer.service.js#L185

Also, we should load the offer transactions into the above service to cater for make it more robust and reusable.

Here is a part I noticed in the UI which is supposed to load the transactions for a specific offer:

Screenshot 2020-10-12 at 8 35 47 PM 1

andela-hpopoola (changes requested):
Thank you for working on this. For the assigned multiple endpoints, let's not load the sum of transactions, to make the query easier and save time. We can always revisit this later. Thank you.

andela-hpopoola (changes requested):
Thank you for working on this. This is taking shape. Let's start work on getting multiple assigned properties.

server/services/property.service.js
```diff
@@ -304,4 +306,26 @@ export const getAssignedPropertyByOfferId = async (offerId) => {
 };
 
 export const getAssignedProperties = async (userId) =>
-  Property.aggregate([{ $match: { assignedTo: ObjectId(userId) } }]);
+  Offer.aggregate([
+    {
+      $match: {
+        $and: [{ userId: ObjectId(userId) }, { status: OFFER_STATUS.ASSIGNED }],
```
andela-hpopoola, line 312: I believe this endpoint should cover both ASSIGNED and ALLOCATED properties. ALLOCATED properties are assigned properties that the payment has reached a certain threshold to give the vendor 100% confidence that the user is buying the property.

andela-hpopoola: Lets add additional tests for ALLOCATED too 🙏

andela-hpopoola (changes requested):
This is taking shape, thank you for the hard work. I will test this later today just to ensure that we have covered all the edge cases and would approve once you have fixed the last comments.

andela-hpopoola (commented):
The code looks good. I will test it locally on my computer and approve later. Thank you for the hard work.

andela-hpopoola (approved):
Thank you for working on this. This works as expected.

Can you add a test for when no transaction has been made for an assigned property. This is important because we need to always test that this works as expected even if we make future changes to the transaction or any other part of the code.

We should test that transaction returns an empty array and that totalAmountPaid is 0.

Feel free to merge after you have written the test.

PR #60, opened 2020-10-16
#175261521 fix referral mail link

tobieeniafe:
#### What does this PR do?
Fix referral mail link

Fix bug of undefined id in referral mail link

`POST /api/v1/referral/invite`

#175261521

### DIFF
--- a/server/services/referral.service.js
+++ b/server/services/referral.service.js
@@ -99,8 +99,9 @@ export const sendReferralInvite = async (invite) => {
 
   const referrer = await getUserById(invite.referrerId);
   try {
-    await addReferral(invite);
+    const referralMessage = await addReferral(invite);
     return {
+      _id: referralMessage._id,
       referrerName: referrer.firstName,
       email: invite.email,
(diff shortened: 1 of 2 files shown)


andela-hpopoola (changes requested):
Thank you for working on this. I noticed that the test that the email is sent is missing from referral.controller.test.js. Can we add this 🙏 ? 

[Optional] Without waiting time, can you see if you can write test that the right link is sent via the mail.

andela-hpopoola (changes requested):
Thank you for the updates. However, like commented in previous pull requests, you need to add tests to ensure that the mail is not sent when the invite is not valid.

Something like adding this to tests where invites are not sent.
```
expect(sendMailSpy.callCount).to.eq(0);
```

Can you please try to ensure you take previous comments to heart so that we don't need to keep repeating ourselves over simple fixes and also save time?

andela-hpopoola (approved):
Thank you for working on this. This looks good to go. 👍

PR #61, opened 2020-10-16
#175261626 load properties in account overview

tobieeniafe:
#### What does this PR do?
Return assigned properties in account overview

Return assigned properties in account overview

`GET api/v1/user/account-overview`

#175261626


andela-hpopoola (changes requested):

Unfortunately, this is not handling the most important reason for loading properties which is the transactions. Here is the expected result I provided in the story:

```
accountOverview: {
           properties: [
              {
                  name : "Property A",
                  paid: "150000",
              },
              {
                 ...
              }
           ],
          totalAmountPaid: 1500000,
          referralRewards: 10000,
          contributionReward: 500000,
}
```

Until we find a way to get the transaction per property, we cannot resolve this story.

PR #65, opened 2020-10-23
#175261571 endpoint to raise concern

tobieeniafe:
#### What does this PR do?
Add endpoint to raise and resolve concern for offer

Add endpoint to raise and resolve concern for offer

`PUT api/v1/offer/raise-concern`
`PUT api/v1/offer/resolve-concern`

#175261571


server/models/offer.model.js
```diff
@@ -151,6 +151,18 @@ const OfferSchema = new mongoose.Schema(
       type: Number,
       required: true,
     },
+    concern: {
```
andela-hpopoola, line 154: This allows only 1 concern, we should allow the model to allow multiple concerns as most users will always raise more concern irrespective of whether their concern has been resolved or not.

server/controllers/offer.controllers.js
```diff
@@ -125,6 +127,35 @@ const OfferController = {
...
+          .json({ success: true, message: 'Concern raised', offer: offer[0] });
+      })
+      .catch((error) => next(error));
+  },
+
+  resolveConcern(req, res, next) {
+    const concern = req.locals;
+    const vendorId = req.user._id;
+    resolveConcern({ ...concern, vendorId })
+      .then((offer) => {
```
andela-hpopoola, line 152: We might need to send another mail when the concern has been resolved too.

mailer/index.js
```diff
@@ -57,6 +57,11 @@ const EMAIL_CONTENT = {
     subject: 'Welcome to BALLers',
     contentTop: `Congratulations you're officially a BALLer.`,
   },
+  RAISE_CONCERN: {
+    subject: 'Concern Raised',
+    contentTop: `A concern has been raised on your offer to [User name]. The question states [question]. `,
```
andela-hpopoola, line 62: Kindly inform and remind Namo on getting or reviewing the contents for this mail.

andela-hpopoola (changes requested):
Thank you for working on this. My major concern is ensuring that a user can raise multiple concerns on an offer irrespective of whether the first concern has been resolved or not.

server/models/offer.model.js
```diff
@@ -151,18 +151,20 @@ const OfferSchema = new mongoose.Schema(
...
-        type: String,
+    concern: [
+      {
+        question: {
+          type: String,
+        },
+        response: {
+          type: String,
+        },
+        status: {
```
andela-hpopoola, line 162: Let's add a date to know when the question was asked and another date for when the response is given.

test/controllers/offer.controller.test.js
```diff
@@ -1359,4 +1359,260 @@ describe('Offer Controller', () => {
...
+      ],
+    });
+
+    const concern = {
+      offerId,
+      concernId,
+      response: 'Yes all rooms are',
+    };
+
+    describe('when offers exist in db', () => {
```
andela-hpopoola, line 1504:
```suggestion
    describe('when an offer exist in db', () => {
```

test/controllers/offer.controller.test.js
```diff
@@ -1359,4 +1359,260 @@ describe('Offer Controller', () => {
...
+      beforeEach(async () => {
+        await addEnquiry(enquiry);
+        await createOffer(offer);
+      });
+
+      context('with a valid token & id', async () => {
+        it('returns successful payload', (done) => {
```
andela-hpopoola, line 1511:
```suggestion
        it('resolves the right concern', (done) => {
```

mailer/index.js
```diff
@@ -59,9 +59,13 @@ const EMAIL_CONTENT = {
   },
   RAISE_CONCERN: {
     subject: 'Concern Raised',
-    contentTop: `A concern has been raised on your offer to [User name]. The question states [question]. `,
+    contentTop: `A concern has been raised on your offer to [User name]. The question states [question].`,
     contentBottom: `Check your dashboard to resolve this concern.`,
   },
+  RESOLVE_CONCERN: {
+    subject: 'Concern Resolved',
+    contentTop: `Your raised concern has been resolved. Check your dashboard for more information.`,
```
andela-hpopoola, line 67: Not really bothered about the implementation here, but I think it would make more sense if the question and the response is given in the mail so that the user can read them without logging into the dashboard.

```suggestion
contentTop: `Your raised concern has been resolved. <br /> <strong>[Question]</strong> <br /> <br /> [Response]`,
```

andela-hpopoola (changes requested):
Thank you for working on this. You did a great work in the comment implementation. I have more comments on test 🙏 

1. Can we write tests to ensure that the user can raise another concern after raising the first concern. This is just to test that the implementation supports multiple questions (I know it supports it but the test ensure we don't break this in the future)

2. Can we also write tests to ensure that the concerns are loaded every time an offer is loaded. You might need to update existing tests to ensure that they are loaded as expected since we just implement concerns. This should be done in all important places where `Offers are loaded` in the controller tests. This ensures that the concerns are available for the users and the admin always.

I will test this on my local machine and approve once you have implemented this above fixes. Thank you and Nice work.

andela-hpopoola:

> Can we also write tests to ensure that the concerns are loaded every time an offer is loaded. You might need to update existing tests to ensure that they are loaded as expected since we just implement concerns. This should be done in all important places where Offers are loaded in the controller tests. This ensures that the concerns are available for the users and the admin always.

I believe you need to add more expectations to ensure that the concerns loads for the admin and users. The trick is to look at all the offer routes and add an expectation to test for concerns for all the `GET` routes. 

Just a single line to test that concerns are returned is enough for each of these routes.

test/controllers/offer.controller.test.js
```diff
@@ -946,6 +946,8 @@ describe('Offer Controller', () => {
               expect(res.body.success).to.be.eql(true);
               expect(res.body).to.have.property('offers');
               expect(res.body.offers[0]).to.have.property('concern');
+              expect(res.body.offers[0].concern.length).to.be.eql(0);
+              expect(res.body.offers[1]).to.have.property('concern');
```
andela-hpopoola, line 950: Not sure why you used offers[1] instead of 0 here. I believe that if we test for the index of 0 only, it will cover for 1 too. 

```suggestion
              expect(res.body.offers[0]).to.have.property('concern');
```

andela-hpopoola: OR this will work too
```suggestion
              expect(res.body.offers[1].concern.length).to.be.eql(0);
```

andela-hpopoola (commented):
Thank you for working on this. I will approve when I fix the test.

test/controllers/offer.controller.test.js
tobieeniafe: exists already in line `948`

andela-hpopoola (approved):
Thank you for updating this. Feel free to merge this when ready.

PR #69, opened 2020-11-02
#175261557 add routes to reactivate offer

tobieeniafe:
#### What does this PR do?
Adds route to reactivate offer letter

Adds route to reactivate offer letter

`PUT /api/v1/offer/reactivate`

#175261557

### DIFF
--- a/server/controllers/offer.controllers.js
+++ b/server/controllers/offer.controllers.js
@@ -9,6 +9,7 @@ import {
   getActiveOffers,
   raiseConcern,
   resolveConcern,
+  reactivateOffer,
 } from '../services/offer.service';
 import httpStatus from '../helpers/httpStatus';
 import EMAIL_CONTENT from '../../mailer';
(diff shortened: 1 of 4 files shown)


server/controllers/offer.controllers.js
```diff
@@ -164,6 +165,16 @@ const OfferController = {
       })
       .catch((error) => next(error));
   },
+
+  reactivate(req, res, next) {
+    const offer = req.locals;
+    const vendorId = req.user._id;
+    reactivateOffer({ ...offer, vendorId })
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Offer reactivated' });
```
andela-hpopoola, line 174: Can you confirm if we need to send a mail on the offer reactivation? I currently think we should send a mail to the user, but you should ask Namo to prevent us from working on something that is not necessary.

andela-hpopoola (changes requested):
Thank you for working on this. The workflow looks good to me. Can you work on the test so that we can approve this asap 🙏 ?

PR #71, opened 2020-11-05
#175179054 add pagination to all users endpoint

tobieeniafe:
#### What does this PR do?
Add pagination to `/api/v1/user/all` endpoint

Add pagination to `/api/v1/user/all` endpoint

`GET /api/v1/user/all?page=1&limit=1`

`page` is the page number. Default is `1`
`limit` is the amount of documents per page. Default is `10`

#175179054

### DIFF
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -128,11 +128,11 @@ const UserController = {
         res.status(httpStatus.OK).json({
           success: true,
           pagination: {
-            currentPage: page,
-            limit,
-            offset: (page - 1) * limit,
+            currentPage: users[0].metadata[0].page,
+            limit: users[0].metadata[0].limit,
+            offset: (page - 1) * users[0].metadata[0].limit,
             total: users[0].metadata[0].total,
-            totalPage: Math.ceil(users[0].metadata[0].total / limit),
+            totalPage: Math.ceil(users[0].metadata[0].total / users[0].metadata[0].limit),
           },
           result: users[0].data,
         });
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -121,8 +121,7 @@ const UserController = {
   },
 
   getAllRegisteredUsers(req, res, next) {
-    const { page } = req.query;
-    const { limit } = req.query;
+    const { page, limit } = req.query;
     getAllRegisteredUsers(page, limit)
       .then((response) => {
         res.status(httpStatus.OK).json({
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -121,14 +121,20 @@ const UserController = {
   },
 
   getAllRegisteredUsers(req, res, next) {
-    const skip = req.params.count;
-    const { limit } = req.params;
-    getAllRegisteredUsers(skip, limit)
+    const page = req.query.page || 1;
+    const limit = req.query.limit || 10;
+    getAllRegisteredUsers(page, limit)
       .then((users) => {
         res.status(httpStatus.OK).json({
           success: true,
-          totalDocumentCount: users[0].metadata[0].total,
-          users: users[0].data,
+          pagination: {
+            currentPage: page,
+            limit,
+            offset: (page - 1) * limit,
+            total: users[0].metadata[0].total,
+            totalPage: Math.ceil(users[0].metadata[0].total / limit),
+          },
+          result: users[0].data,
         });
       })
       .catch((error) => next(error));
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -121,9 +121,15 @@ const UserController = {
   },
 
   getAllRegisteredUsers(req, res, next) {
-    getAllRegisteredUsers()
+    const skip = req.params.count;
+    const { limit } = req.params;
+    getAllRegisteredUsers(skip, limit)
       .then((users) => {
-        res.status(httpStatus.OK).json({ success: true, users });
+        res.status(httpStatus.OK).json({
+          success: true,
+          totalDocumentCount: users[0].metadata[0].total,
+          users: users[0].data,
+        });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -123,11 +123,11 @@ const UserController = {
   getAllRegisteredUsers(req, res, next) {
     const { page, limit } = req.query;
     getAllRegisteredUsers(page, limit)
-      .then((response) => {
+      .then(({ pagination, result }) => {
         res.status(httpStatus.OK).json({
           success: true,
-          pagination: response.pagination,
-          result: response.result,
+          pagination,
+          result,
         });
       })
       .catch((error) => next(error));
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -121,20 +121,14 @@ const UserController = {
   },
 
   getAllRegisteredUsers(req, res, next) {
-    const page = req.query.page || 1;
-    const limit = req.query.limit || 10;
+    const { page } = req.query;
+    const { limit } = req.query;
     getAllRegisteredUsers(page, limit)
-      .then((users) => {
+      .then((response) => {
         res.status(httpStatus.OK).json({
           success: true,
-          pagination: {
-            currentPage: users[0].metadata[0].page,
-            limit: users[0].metadata[0].limit,
-            offset: (page - 1) * users[0].metadata[0].limit,
-            total: users[0].metadata[0].total,
-            totalPage: Math.ceil(users[0].metadata[0].total / users[0].metadata[0].limit),
-          },
-          result: users[0].data,
+          pagination: response.pagination,
+          result: response.result,
         });
       })
       .catch((error) => next(error));
--- a/server/helpers/pagination.js
+++ b/server/helpers/pagination.js
@@ -1,9 +1,12 @@
-const generatePagination = (page, limit, total) => ({
-  currentPage: parseInt(page, 10),
-  limit: parseInt(limit, 10),
-  offset: (page - 1) * limit,
-  total,
-  totalPage: Math.ceil(total / limit),
-});
+const generatePagination = (page, limit, total) => {
+  const currentPage = parseInt(page, 10) < 1 ? 1 : parseInt(page, 10);
+  return {
+    currentPage,
+    limit: parseInt(limit, 10),
+    offset: (currentPage - 1) * limit,
+    total,
+    totalPage: Math.ceil(total / limit),
+  };
+};
 
 export default generatePagination;
--- a/server/helpers/pagination.js
+++ b/server/helpers/pagination.js
@@ -1,0 +1,9 @@
+const generatePagination = (page, limit, total) => ({
+  currentPage: parseInt(page, 10),
+  limit: parseInt(limit, 10),
+  offset: (page - 1) * limit,
+  total,
+  totalPage: Math.ceil(total / limit),
+});
+
+export default generatePagination;
--- a/server/routes/user.routes.js
+++ b/server/routes/user.routes.js
@@ -278,7 +278,7 @@ router.post('/upload-image', authenticate, Upload.uploadImage, UploadController.
 
 /**
  * @swagger
- * /user/all:
+ * /user/all/:limit/:count:
  *   get:
  *     tags:
  *       - User
@@ -297,7 +297,7 @@ router.post('/upload-image', authenticate, Upload.uploadImage, UploadController.
  *      '500':
  *       description: Internal server error
  */
-router.get('/all', authenticate, isAdmin, UserController.getAllRegisteredUsers);
+router.get('/all/:limit/:count', authenticate, isAdmin, UserController.getAllRegisteredUsers);
 
 /**
  * @swagger
--- a/server/routes/user.routes.js
+++ b/server/routes/user.routes.js
@@ -278,7 +278,7 @@ router.post('/upload-image', authenticate, Upload.uploadImage, UploadController.
 
 /**
  * @swagger
- * /user/all/:limit/:count:
+ * /user/all:
  *   get:
  *     tags:
  *       - User
@@ -297,7 +297,7 @@ router.post('/upload-image', authenticate, Upload.uploadImage, UploadController.
  *      '500':
  *       description: Internal server error
  */
-router.get('/all/:limit/:count', authenticate, isAdmin, UserController.getAllRegisteredUsers);
+router.get('/all', authenticate, isAdmin, UserController.getAllRegisteredUsers);
 
 /**
  * @swagger
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -10,6 +10,7 @@ import { calculateContributionReward } from './offer.service';
 import { addReferral, calculateReferralRewards } from './referral.service';
 import { getTotalAmountPaidByUser } from './transaction.service';
 import { REFERRAL_STATUS, USER_ROLE } from '../helpers/constants';
+import generatePagination from '../helpers/pagination';
 
 export const getUserByEmail = async (email, fields = null) =>
   User.findOne({ email }).select(fields);
@@ -248,7 +249,7 @@ export const updateUser = async (updatedUser) => {
 export const getAllRegisteredUsers = async (pageNumber, resultLimit) => {
   const page = pageNumber || 1;
   const limit = resultLimit || 10;
-  return User.aggregate([
+  const users = await User.aggregate([
     {
       $lookup: {
         from: 'properties',
@@ -260,12 +261,15 @@ export const getAllRegisteredUsers = async (pageNumber, resultLimit) => {
     {
       $facet: {
         metadata: [{ $count: 'total' }, { $addFields: { page, limit } }],
-        // eslint-disable-next-line radix
-        data: [{ $skip: parseInt((page - 1) * limit) }, { $limit: parseInt(limit) }],
+        data: [{ $skip: parseInt((page - 1) * limit, 10) }, { $limit: parseInt(limit, 10) }],
       },
     },
     { $project: { preferences: 0, password: 0, notifications: 0 } },
   ]);
+  const { total } = users[0].metadata[0];
+  const pagination = generatePagination(page, limit, total);
+  const result = users[0].data;
+  return { pagination, result };
 };
 
 export const addPropertyToFavorites = async (favorite) => {
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -259,7 +259,7 @@ export const getAllRegisteredUsers = async (pageNumber, resultLimit) => {
     },
     {
       $facet: {
-        metadata: [{ $count: 'total' }, { $addFields: { page } }],
+        metadata: [{ $count: 'total' }, { $addFields: { page, limit } }],
         // eslint-disable-next-line radix
         data: [{ $skip: parseInt((page - 1) * limit) }, { $limit: parseInt(limit) }],
       },
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -245,8 +245,10 @@ export const updateUser = async (updatedUser) => {
   }
 };
 
-export const getAllRegisteredUsers = async (skip, limit) =>
-  User.aggregate([
+export const getAllRegisteredUsers = async (pageNumber, resultLimit) => {
+  const page = pageNumber || 1;
+  const limit = resultLimit || 10;
+  return User.aggregate([
     {
       $lookup: {
         from: 'properties',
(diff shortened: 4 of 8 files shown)


server/controllers/user.controllers.js
```diff
@@ -121,9 +121,15 @@ const UserController = {
   },
 
   getAllRegisteredUsers(req, res, next) {
-    getAllRegisteredUsers()
+    const skip = req.params.count;
```
andela-hpopoola, line 124: This should be optional, so there should be a default value of 0 here. Also, can we make this offset instead of count so that it can be similar to the output defined.
```suggestion
    const skip = req.params.offset || 0;
```

server/controllers/user.controllers.js
```diff
@@ -121,9 +121,15 @@ const UserController = {
...
+    const { limit } = req.params;
```
andela-hpopoola, line 125: This also should be optional:

```suggestion
    const { limit } = req.params || 10;
```

server/controllers/user.controllers.js
```diff
@@ -121,9 +121,15 @@ const UserController = {
...
+    getAllRegisteredUsers(skip, limit)
       .then((users) => {
-        res.status(httpStatus.OK).json({ success: true, users });
+        res.status(httpStatus.OK).json({
+          success: true,
+          totalDocumentCount: users[0].metadata[0].total,
```
andela-hpopoola, line 130: This is not required in the output given in the story 🤷‍♂️ . The total should be inside the pagination object.

server/routes/user.routes.js
```diff
@@ -278,7 +278,7 @@ router.post('/upload-image', authenticate, Upload.uploadImage, UploadController.
 
 /**
  * @swagger
- * /user/all:
+ * /user/all/:limit/:count:
```
andela-hpopoola, line 281: Pagination should not affect the current endpoint. Let's make this query params instead of including it in the endpoint.

```suggestion
 * /user/all
```

server/routes/user.routes.js
```diff
@@ -297,7 +297,7 @@ router.post('/upload-image', authenticate, Upload.uploadImage, UploadController.
  *      '500':
  *       description: Internal server error
  */
-router.get('/all', authenticate, isAdmin, UserController.getAllRegisteredUsers);
+router.get('/all/:limit/:count', authenticate, isAdmin, UserController.getAllRegisteredUsers);
```
andela-hpopoola, line 300: Same comment above.

server/services/user.service.js
```diff
@@ -245,16 +245,23 @@ export const updateUser = async (updatedUser) => {
...
+        localField: '_id',
+        foreignField: 'assignedTo',
         as: 'assignedProperties',
       },
     },
+    {
+      $facet: {
+        metadata: [{ $count: 'total' }], //  { $addFields: { page: 1 } }
+        // eslint-disable-next-line radix
+        data: [{ $skip: parseInt(skip) }, { $limit: parseInt(limit) }],
```
andela-hpopoola, line 262:
```suggestion
        data: [{ $skip: parseInt(skip, 10) }, { $limit: parseInt(limit) }],
```

server/services/user.service.js
```diff
@@ -245,16 +245,23 @@ export const updateUser = async (updatedUser) => {
...
-        foreignField: '_id',
+        localField: '_id',
+        foreignField: 'assignedTo',
         as: 'assignedProperties',
       },
     },
+    {
+      $facet: {
+        metadata: [{ $count: 'total' }], //  { $addFields: { page: 1 } }
+        // eslint-disable-next-line radix
```
andela-hpopoola, line 261: This is not necessary, ParseInt requires a radix

server/services/user.service.js
```diff
@@ -245,16 +245,23 @@ export const updateUser = async (updatedUser) => {
...
-        localField: 'assignedProperties.propertyId',
-        foreignField: '_id',
+        localField: '_id',
+        foreignField: 'assignedTo',
         as: 'assignedProperties',
       },
     },
+    {
+      $facet: {
+        metadata: [{ $count: 'total' }], //  { $addFields: { page: 1 } }
```
andela-hpopoola, line 260: Ensure you clear this up to prevent leaving unnecessary comments.

andela-hpopoola (changes requested):
Thank you for working on this. You didn't request my review on this so I am not sure whether this is ready for review or not. 

You already promised that you would finish your 7 point story (with no excuse before each Monday -- that is all your stories will have to be merged before Monday and not just having a PR raised -- I decided to review this to make you reach your deadline faster). 

Please fix and make this work as expected asap so that we can merge before Monday. Thank you.

server/controllers/user.controllers.js
```diff
@@ -121,9 +121,21 @@ const UserController = {
   },
 
   getAllRegisteredUsers(req, res, next) {
-    getAllRegisteredUsers()
+    const page = req.query.page || 1;
```
andela-hpopoola, line 124: Lets remove the default values here and use the one in the services. Having the default values defined in multiple places can cause errors for us in the future.

```suggestion
    const page = req.query.page;
```

server/controllers/user.controllers.js
```diff
@@ -121,9 +121,21 @@ const UserController = {
...
+    const limit = req.query.limit || 10;
+    getAllRegisteredUsers(page, limit)
       .then((users) => {
-        res.status(httpStatus.OK).json({ success: true, users });
+        res.status(httpStatus.OK).json({
+          success: true,
+          pagination: {
```
andela-hpopoola, line 130: The pagination object should be constructed and returned from the `get all users` service. You should create another service to generate this. This ensures that we can always reuse this for other services. Something like this:

```js
const generatePagination = (currentPage, limit, total) => ({
         currentPage,
         limit,
         offset: (page - 1) * limit,
         total,
         totalPage: Math.ceil(total / limit),
})
```

andela-hpopoola (changes requested):
Thank you for working on this. I believe we need to add more tests to ensure that this works as expected:

1. Add tests to ensure that the limit is updated from the default value of 10 when it is given in in the query
2. Add lots of users (create the users via factory's buildList and bulk insert), and test to be sure that the page parameter gets a new page when the page is updated
3. Add lots of users again and test that when the number of returned user exceeds the total number of limit, the service returns only the limit and ignores the other ones.

test/controllers/user.controller.test.js
```diff
@@ -811,21 +811,83 @@ describe('User Controller', () => {
...
               expect(res).to.have.status(200);
               expect(res.body.success).to.be.eql(true);
-              expect(res.body.result.length).to.be.eql(2);
+              expect(res.body.result.length).to.be.eql(10);
               expect(res.body.pagination.currentPage).to.be.eql(1);
               expect(res.body.pagination.limit).to.be.eql(10);
               expect(res.body.pagination.offset).to.be.eql(0);
-              expect(res.body.pagination.total).to.be.eql(2);
-              expect(res.body.pagination.totalPage).to.be.eql(1);
+              expect(res.body.pagination.total).to.be.eql(20);
```
andela-hpopoola, line 832: Hmmm, you added 18 dummy users, I am not sure why this is 20, I would expect it to be 18.

test/controllers/user.controller.test.js
```diff
@@ -811,21 +811,83 @@ describe('User Controller', () => {
...
+      context('when parameters page and limit are passed', () => {
+        it('returns second page with five items', (done) => {
+          request()
+            .get('/api/v1/user/all?page=2&limit=5')
+            .set('authorization', adminToken)
+            .end((err, res) => {
+              expect(res).to.have.status(200);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body.result.length).to.be.eql(5);
+              expect(res.body.pagination.currentPage).to.be.eql('2');
```
andela-hpopoola, line 848: Can we make this a number so that it can be the same with the other parameters: Please fix for other assertions too.
```suggestion
              expect(res.body.pagination.currentPage).to.be.eql(2);
```

test/controllers/user.controller.test.js
```diff
@@ -811,21 +811,83 @@ describe('User Controller', () => {
...
+              expect(res.body.pagination.limit).to.be.eql('5');
+              expect(res.body.pagination.offset).to.be.eql(5);
+              expect(res.body.pagination.total).to.be.eql(20);
```
andela-hpopoola, line 851: Same comment as above, I presume this to be 18. Please fix for others too.

server/controllers/user.controllers.js
```diff
@@ -128,11 +128,11 @@ const UserController = {
         res.status(httpStatus.OK).json({
           success: true,
           pagination: {
-            currentPage: page,
-            limit,
-            offset: (page - 1) * limit,
+            currentPage: users[0].metadata[0].page,
```
andela-hpopoola, line 131: I am not sure you understand the feedback on this, I believe it is better to generate the pagination in the `getAllRegisteredUsers` service and just retrieve in the controller. 

This will enable us to use a custom method to generate the pagination instead of rewriting this same object over and over again for other services when we convert them into pagination.

test/controllers/user.controller.test.js
```diff
@@ -811,21 +811,83 @@ describe('User Controller', () => {
...
+
     describe('when users exist in db', () => {
-      context('with a valid token & id', () => {
-        it('returns successful payload', (done) => {
+      context('when no parameters are passed', () => {
+        it('returns first page with ten items', (done) => {
           request()
             .get('/api/v1/user/all')
             .set('authorization', adminToken)
             .end((err, res) => {
```
andela-hpopoola, line 825: Instead of repeating the same pagination tests over and over again, we can write a custom function for the tests and just pass in the values. Something like this

https://github.com/Highrachy/ballers-api/blob/84770a19988df9b8da88ce61034f5b87491c7573/test/controllers/mailer.controller.test.js#L6


For me, I would create a custom  helper test file and add the above function (I believe we will still use it to test other controllers). The function would look like this:

```js
const expectsPaginationToReturnTheRightValues = (res, { result, currentPage, limit, offset, total, totalPage }) => {
  expect(res.body.result.length).to.be.eql(10);
  expect(res.body.pagination.currentPage).to.be.eql(1);
  expect(res.body.pagination.limit).to.be.eql(10);
  expect(res.body.pagination.offset).to.be.eql(0);
  expect(res.body.pagination.total).to.be.eql(20);
  expect(res.body.pagination.totalPage).to.be.eql(2);
};
```

The test expection will look like this:

```js
context('when only parameter page is passed', () => {
        it('returns second page with limit of 10', (done) => {
          request()
            .get('/api/v1/user/all?page=2')
            .end((err, res) => {
             expectsPaginationToReturnTheRightValues(res, { result: 10, currentPage: 1, limit: 10, offset: 0, total: 20, totalPage: 2});
              done();
            });
        });

```

To make it even more cool, you can define a default expected. value for the pagination

```js
const expectedPaginationResult =  { result: 10, currentPage: 1, limit: 10, offset: 0, total: 20, totalPage: 2};
````

Then in the test, you will have something like this so that you will overwrite only the one you believe would change :

```js
   expectsPaginationToReturnTheRightValues(res, { ...expectedPaginationResult, limit: 5 });
```

server/controllers/user.controllers.js
andela-hpopoola: You didn't resolve this and didn't add a comment on why you think it is better this way. Can we remove this and use the default in the service?

server/services/user.service.js
andela-hpopoola: You didn't fix this too and you resolved it 😏 . Can we remove the eslint comment please?

test/controllers/user.controller.test.js
```diff
@@ -811,17 +811,83 @@ describe('User Controller', () => {
...
+    const dummyUsers = UserFactory.buildList(18);
+    beforeEach(async () => {
+      await User.insertMany(dummyUsers);
+    });
+
     describe('when users exist in db', () => {
-      context('with a valid token & id', () => {
-        it('returns successful payload', (done) => {
+      context('when no parameters are passed', () => {
+        it('returns first page with ten items', (done) => {
```
andela-hpopoola, line 821:
```suggestion
        it('returns the default values', (done) => {
```

test/controllers/user.controller.test.js
```diff
@@ -811,17 +811,83 @@ describe('User Controller', () => {
...
+              expect(res.body.pagination.offset).to.be.eql(0);
+              expect(res.body.pagination.total).to.be.eql(20);
+              expect(res.body.pagination.totalPage).to.be.eql(2);
+              done();
+            });
+        });
+      });
+
+      context('when parameters page and limit are passed', () => {
+        it('returns second page with five items', (done) => {
```
andela-hpopoola, line 840:
```suggestion
        it('returns the given page and limit', (done) => {
```

test/controllers/user.controller.test.js
```diff
@@ -811,17 +811,83 @@ describe('User Controller', () => {
...
+              expect(res.body.pagination.offset).to.be.eql(5);
+              expect(res.body.pagination.total).to.be.eql(20);
+              expect(res.body.pagination.totalPage).to.be.eql(4);
+              done();
+            });
+        });
+      });
+
+      context('when only parameter page is passed', () => {
+        it('returns second page with limit of 10', (done) => {
```
andela-hpopoola, line 859:
```suggestion
        it('returns the second page', (done) => {
```

test/controllers/user.controller.test.js
```diff
@@ -811,17 +811,83 @@ describe('User Controller', () => {
...
+              expect(res.body.pagination.offset).to.be.eql(10);
+              expect(res.body.pagination.total).to.be.eql(20);
+              expect(res.body.pagination.totalPage).to.be.eql(2);
+              done();
+            });
+        });
+      });
+
+      context('when only parameter limit is passed', () => {
+        it('returns first page with four items', (done) => {
```
andela-hpopoola, line 878:
```suggestion
        it('returns 4 users', (done) => {
```

test/controllers/user.controller.test.js
```diff
@@ -811,17 +811,83 @@ describe('User Controller', () => {
...
+              expect(res.body.pagination.limit).to.be.eql(10);
+              expect(res.body.pagination.offset).to.be.eql(10);
+              expect(res.body.pagination.total).to.be.eql(20);
+              expect(res.body.pagination.totalPage).to.be.eql(2);
+              done();
+            });
+        });
+      });
+
+      context('when only parameter limit is passed', () => {
```
andela-hpopoola, line 877:
```suggestion
      context('when limit is set to 4', () => {
```

test/controllers/user.controller.test.js
```diff
@@ -811,17 +811,83 @@ describe('User Controller', () => {
...
+              expect(res.body.pagination.limit).to.be.eql('5');
+              expect(res.body.pagination.offset).to.be.eql(5);
+              expect(res.body.pagination.total).to.be.eql(20);
+              expect(res.body.pagination.totalPage).to.be.eql(4);
+              done();
+            });
+        });
+      });
+
+      context('when only parameter page is passed', () => {
```
andela-hpopoola, line 858:
```suggestion
      context('when page is set to 2', () => {
```

test/controllers/user.controller.test.js
```diff
@@ -811,17 +811,83 @@ describe('User Controller', () => {
...
+              expect(res.body.pagination.limit).to.be.eql(10);
+              expect(res.body.pagination.offset).to.be.eql(0);
+              expect(res.body.pagination.total).to.be.eql(20);
+              expect(res.body.pagination.totalPage).to.be.eql(2);
+              done();
+            });
+        });
+      });
+
+      context('when parameters page and limit are passed', () => {
```
andela-hpopoola, line 839:
```suggestion
      context('when both page and limit parameters are set', () => {
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments but I am particularly concerned about resolving conversations without fixing them. Can we ensure that we drop a comment if we feel there is no need to fix a particular section of the code to serve as an awareness to the reviewer on the reason why the code is not fixed to save all of our time? 

Thank you.

server/controllers/user.controllers.js
```diff
@@ -121,9 +121,15 @@ const UserController = {
   },
 
   getAllRegisteredUsers(req, res, next) {
-    getAllRegisteredUsers()
-      .then((users) => {
-        res.status(httpStatus.OK).json({ success: true, users });
+    const { page } = req.query;
```
andela-hpopoola, line 124: This can be combined with the one below:

```suggestion
    const { page, limit } = req.query;
```

server/services/user.service.js
```diff
@@ -245,18 +246,31 @@ export const updateUser = async (updatedUser) => {
   }
 };
 
-export const getAllRegisteredUsers = async () =>
-  User.aggregate([
+export const getAllRegisteredUsers = async (pageNumber, resultLimit) => {
```
andela-hpopoola, line 249: I think we can use the page and limit as default parameters directly and remove the two lines below
```suggestion
export const getAllRegisteredUsers = async (page = 1, limit=10) => {
```

test/controllers/user.controller.test.js
```diff
@@ -39,156 +40,488 @@ describe('User Controller', () => {
     sandbox.restore();
   });
 
-  beforeEach(async () => {
-    adminToken = await addUser(adminUser);
-    userToken = await addUser(regularUser);
-  });
+  describe('', () => {
```
andela-hpopoola, line 43:
```suggestion
  describe('User Controller', () => {
```

test/controllers/user.controller.test.js
```diff
@@ -1152,183 +1213,199 @@ describe('User Controller', () => {
...
-            User.findByIdAndUpdate.restore();
-          });
+    describe('when users exist in db', () => {
+      context('when no parameters are passed', () => {
+        it('returns the default values', (done) => {
+          request()
+            .get('/api/v1/user/all')
+            .set('authorization', adminToken)
+            .end((err, res) => {
+              expectsPaginationToReturnTheRightValues(res, {
```
andela-hpopoola, line 1282: Lets use a default value for this and other tests to make the test assertion more readable:

```suggestion
              expectsPaginationToReturnTheRightValues(res, expectedPaginationResult);
```

test/controllers/user.controller.test.js
```diff
@@ -1152,183 +1213,199 @@ describe('User Controller', () => {
...
-            .put('/api/v1/user/editor/upgrade')
+            .get('/api/v1/user/all?page=2&limit=5')
             .set('authorization', adminToken)
-            .send({ userId: '' })
             .end((err, res) => {
-              expect(res).to.have.status(412);
-              expect(res.body.success).to.be.eql(false);
-              expect(res.body.message).to.be.eql('Validation Error');
-              expect(res.body.error).to.be.eql('"User id" is not allowed to be empty');
+              expectsPaginationToReturnTheRightValues(res, {
```
andela-hpopoola, line 1301: Using a default value will make this more readable:

```suggestion
               expectsPaginationToReturnTheRightValues(res, { ...expectedPaginationResult, limit:5, page: 2 });
```

andela-hpopoola (changes requested):
Thank you for fixing this. I left a few comments to ensure that the code and tests are of high quality. This is to ensure that we save time when we are implementing the same functionality for similar services.

server/helpers/pagination.js
```diff
@@ -0,0 +1,9 @@
+const generatePagination = (page, limit, total) => ({
```
andela-hpopoola, line 1: Even though it not too important, it is best practises to ensure that the page parameter is never less than 0. This value will be received from the frontend and it is best practises never to trust anything from the user.

Can we add a conditional statement to ensure this. Something like this should work:

```js
const currentPage = parseInt(page, 10) < 1 ? 1 : parseInt(page, 10);
```

However, you will also need to change the page value used for offset to currentPage:

```js
...
offset: (currentPage - 1) * limit;
...
```

test/controllers/user.controller.test.js
```diff
@@ -1152,180 +1213,193 @@ describe('User Controller', () => {
...
-          });
-      });
+  describe('Get all users', () => {
+    const dummyUsers = UserFactory.buildList(16);
+    beforeEach(async () => {
+      adminToken = await addUser(adminUser);
+      userToken = await addUser(regularUser);
+      await User.insertMany(dummyUsers);
     });
+    const defaultPaginationResult = {
```
andela-hpopoola, line 1274: Not important and can be ignored, however, arranging the values of an object alphabetically makes people see you as a World Class developer.

```js

const defaultPaginationResult = {
  currentPage: 1,
  limit: 10,
  offset: 0,
  result: 10,
  total: 18,
  totalPage: 2,
};


```

andela-hpopoola (approved):
Phew: 



Thank you for making the fixes and Nice work. This is a good step forward in implementing pagination for other endpoints, so it will only get easier. 

I left a few comments but they are most nits and non-blocking. Feel free to merge when ready.

PR #72, opened 2020-11-09
#175634783 create content property model

tobieeniafe:
#### What does this PR do?
Create content property model

Create content property model for properties used in property search and price estimation

NA

#175634783

### DIFF
--- a/server/models/contentProperty.model.js
+++ b/server/models/contentProperty.model.js
@@ -7,24 +7,15 @@ import mongoose from 'mongoose';
  *      ContentProperty:
  *        type: object
  *        required:
- *          - state
- *          - area
- *          - longitude
- *          - latitude
+ *          - areaId
  *          - category
  *          - houseType
  *          - price
  *          - website
  *          - link
  *        properties:
- *          state:
- *            type: string
- *          area:
- *            type: string
- *          longitude:
- *            type: number
- *          latitude:
- *            type: number
+ *          areaId:
+ *            type: objectid
  *          category:
  *            type: string
  *          houseType:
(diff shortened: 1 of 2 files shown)


server/models/contentProperty.model.js
```diff
@@ -0,0 +1,90 @@
...
+  {
+    state: {
+      type: String,
+      required: true,
+    },
+    area: {
+      type: String,
+      required: true,
+    },
+    longitude: {
```
andela-hpopoola, line 60: The longtitude and latitude belongs to the area and not the housetype. This structure will be problematic and difficult to maintain as there is no way to get a single longitude and latitude for many types of properties with the same area.

The point is that there should only be one source of truth for the longitude and latitude. 

We will need to split this into multiple models. One to load the content property area which would contain the state, area, longitude, and latitude. The content property schema will contain category, houseType, price, website, link and the reference to the  to the content property area schema (via an id).

With this, there will be a source of truth for the area information and this can be updated separately without affecting the house Type added.

andela-hpopoola (changes requested):
Thank you for working on this. I am afraid your current implementation is not optimal and would cause lots of problems for us in the future.

The longitude and latitude belong to the area and not the housetype. This structure will be problematic and difficult to maintain as there is no way to get a single longitude and latitude for many types of properties with the same area.

The point is that there should only be one source of truth for the longitude and latitude.

We will need to split this into multiple models. One to load the content property area which would contain the state, area, longitude, and latitude. The content property schema will contain category, houseType, price, website, link, and the reference to the content property area schema (via an id).

With this, there will be a source of truth for the area information and this can be updated separately without affecting the house Type added.

andela-hpopoola (approved):
Looks good to me. Thank you for making the changes.

PR #74, opened 2020-11-14
#175634788 edit and delete property routes

tobieeniafe:
#### What does this PR do?
Add routes to edit and delete properties

Add routes to edit and delete properties

`PUT /api/v1/content-property/update`
`DELETE /api/v1/content-property/delete/:id`

#175634788

### DIFF
--- a/server/controllers/contentProperty.controllers.js
+++ b/server/controllers/contentProperty.controllers.js
@@ -1,4 +1,8 @@
-import addContentProperty from '../services/contentProperty.service';
+import {
+  addContentProperty,
+  updateContentProperty,
+  deleteContentProperty,
+} from '../services/contentProperty.service';
 import httpStatus from '../helpers/httpStatus';
 
 const ContentPropertyController = {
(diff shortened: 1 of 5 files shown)


server/controllers/contentProperty.controllers.js
```diff
@@ -12,6 +16,24 @@ const ContentPropertyController = {
       })
       .catch((error) => next(error));
   },
+
+  update(req, res, next) {
+    const updatedPaymentPlan = req.locals;
+    updateContentProperty(updatedPaymentPlan)
+      .then((plan) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Property updated', plan });
```
andela-hpopoola, line 24:
```suggestion
        res.status(httpStatus.OK).json({ success: true, message: 'Content Property updated', plan });
```

andela-hpopoola (changes requested):
This looks good, Thank you for the hardwork. Can we add tests for this, please? 

Also, we would need similar endpoints to update and delete the area, maybe a different PR 🤷‍♂️

andela-hpopoola (approved):
Thank you for working on this. Even though I prefer we should also add services test, I am not too concerned about it (and it can be skipped to be done later). I am approving this, so feel free to merge when ready. Nice work.

PR #75, opened 2020-11-14
#175634790 search for content property

tobieeniafe:
#### What does this PR do?
Search for content property

Search for content property based on state, area & house type

`GET api/v1/area/`
`GET api/v1/area/:state`
`GET /api/v1/content-property/area/:id`

#175634790

### DIFF
--- a/server/controllers/area.controllers.js
+++ b/server/controllers/area.controllers.js
@@ -1,4 +1,4 @@
-import { addArea, getState, getArea } from '../services/area.service';
+import { addArea, getStates, getAreas } from '../services/area.service';
 import httpStatus from '../helpers/httpStatus';
 
 const AreaController = {
@@ -13,17 +13,17 @@ const AreaController = {
       .catch((error) => next(error));
   },
 
-  getState(req, res, next) {
-    getState()
+  getStates(req, res, next) {
+    getStates()
       .then((states) => {
         res.status(httpStatus.OK).json({ success: true, states: states.states });
       })
       .catch((error) => next(error));
   },
 
-  getArea(req, res, next) {
+  getAreas(req, res, next) {
     const { state } = req.params;
-    getArea(state)
+    getAreas(state)
       .then((areas) => {
         res.status(httpStatus.OK).json({ success: true, areas: areas.areas });
       })
--- a/server/controllers/area.controllers.js
+++ b/server/controllers/area.controllers.js
@@ -1,4 +1,4 @@
-import { addArea, getStateAndArea } from '../services/area.service';
+import { addArea, getState, getArea } from '../services/area.service';
 import httpStatus from '../helpers/httpStatus';
 
 const AreaController = {
@@ -14,7 +14,7 @@ const AreaController = {
   },
 
   getState(req, res, next) {
-    getStateAndArea()
+    getState()
       .then((states) => {
         res.status(httpStatus.OK).json({ success: true, states: states.states });
       })
@@ -23,7 +23,7 @@ const AreaController = {
 
   getArea(req, res, next) {
     const { state } = req.params;
-    getStateAndArea(state)
+    getArea(state)
       .then((areas) => {
         res.status(httpStatus.OK).json({ success: true, areas: areas.areas });
       })
--- a/server/controllers/area.controllers.js
+++ b/server/controllers/area.controllers.js
@@ -1,4 +1,4 @@
-import { addArea } from '../services/area.service';
+import { addArea, getStateAndArea } from '../services/area.service';
 import httpStatus from '../helpers/httpStatus';
 
 const AreaController = {
@@ -9,6 +9,15 @@ const AreaController = {
         res
           .status(httpStatus.CREATED)
           .json({ success: true, message: 'Area added successfully', area });
+      })
+      .catch((error) => next(error));
+  },
+
+  search(req, res, next) {
+    const { state } = req.query;
+    getStateAndArea(state)
+      .then((locations) => {
+        res.status(httpStatus.OK).json({ success: true, locations });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/area.controllers.js
+++ b/server/controllers/area.controllers.js
@@ -13,11 +13,19 @@ const AreaController = {
       .catch((error) => next(error));
   },
 
-  search(req, res, next) {
-    const { state } = req.query;
+  getState(req, res, next) {
+    getStateAndArea()
+      .then((states) => {
+        res.status(httpStatus.OK).json({ success: true, states: states.states });
+      })
+      .catch((error) => next(error));
+  },
+
+  getArea(req, res, next) {
+    const { state } = req.params;
     getStateAndArea(state)
-      .then((locations) => {
-        res.status(httpStatus.OK).json({ success: true, locations });
+      .then((areas) => {
+        res.status(httpStatus.OK).json({ success: true, areas: areas.areas });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/contentProperty.controllers.js
+++ b/server/controllers/contentProperty.controllers.js
@@ -38,7 +38,7 @@ const ContentPropertyController = {
       .catch((error) => next(error));
   },
 
-  search(req, res, next) {
+  getHouseTypesByAreaId(req, res, next) {
     const areaId = req.params.id;
     getHouseTypesByAreaId(areaId)
       .then((houseTypes) => {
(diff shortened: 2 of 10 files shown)


server/controllers/area.controllers.js
```diff
@@ -12,6 +12,15 @@ const AreaController = {
       })
       .catch((error) => next(error));
   },
+
+  search(req, res, next) {
```
andela-hpopoola, line 16: Lets use another name for this instead of search. This is not actually searching but getting the states and areas.

server/controllers/area.controllers.js
```diff
@@ -12,6 +12,15 @@ const AreaController = {
...
+    const { state } = req.query;
+    getStateAndArea(state)
```
andela-hpopoola, line 18: Let's separate this into 2 different endpoints (one to get the states and another to get the areas), please 🙏 ? The idea of returning location for both states and areas doesn't follow the best practises.

server/controllers/contentProperty.controllers.js
```diff
@@ -36,6 +37,15 @@ const ContentPropertyController = {
       })
       .catch((error) => next(error));
   },
+
+  search(req, res, next) {
```
andela-hpopoola, line 41: This is getting the house types, so using search here is not explanatory enough.

andela-hpopoola (changes requested):
Great work on this. This looks good to me. Apart from splitting the states and areas into different endpoints to make it easier to maintain and follow best practices, the implementation looks great. 

Waiting for the tests so that we can merge this asap. Great work and thank you for the lovely implementation.

server/services/area.service.js
```diff
@@ -12,3 +12,16 @@ export const addArea = async (area) => {
     throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error adding area', error);
   }
 };
+
+export const getStateAndArea = async (state) => {
+  if (!state) {
+    const states = await Area.distinct('state');
+    return { states };
+  }
+  const areas = await Area.find({ state });
```
andela-hpopoola, line 21: I believe the below lines are not needed since we don't know the format the frontend will need. It is better to just return the id and area as an array and let the frontend convert it to the needed format. 

```suggestion
  const areas = await Area.find({ state }).select('area');
  return { areas };
```

server/services/area.service.js
```diff
@@ -12,3 +12,16 @@ export const addArea = async (area) => {
     throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error adding area', error);
   }
 };
+
+export const getStateAndArea = async (state) => {
```
andela-hpopoola, line 16: Lets separate this service and have different services for state and another one for the area 🙏 . Combining them together doesn't give us any advantage, we will still call the controller 2 times to retrieve them so it is better to leave them separated for now.

test/controllers/area.controller.test.js
```diff
@@ -148,4 +144,159 @@ describe('Area Controller', () => {
...
+  describe('Get all states', () => {
+    const lagosState = AreaFactory.build({ state: 'lagos' });
+    const oyoState = AreaFactory.build({ state: 'oyo' });
+    beforeEach(async () => {
+      await addArea(lagosState);
+      await addArea(oyoState);
+    });
+
+    context('when editor token is used', () => {
+      it('returns array of five states', (done) => {
```
andela-hpopoola, line 157: Not sure why you used 5 states here:
```suggestion
      it('returns array of two states', (done) => {
```

test/controllers/area.controller.test.js
```diff
@@ -148,4 +144,159 @@ describe('Area Controller', () => {
...
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body.states.length).to.be.eql(2);
+            done();
+          });
+      });
+    });
+
+    context('when admin token is used', () => {
+      it('returns array of five states', (done) => {
```
andela-hpopoola, line 171: Same comment as above

test/controllers/area.controller.test.js
```diff
@@ -148,4 +144,159 @@ describe('Area Controller', () => {
...
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body.areas.length).to.be.eql(5);
+            done();
+          });
+      });
+    });
+
+    context('when admin token is used', () => {
+      it('returns array of five states', (done) => {
```
andela-hpopoola, line 248:
```suggestion
      it('returns array of five areas', (done) => {
```

test/controllers/area.controller.test.js
```diff
@@ -148,4 +144,159 @@ describe('Area Controller', () => {
...
+
+  describe('Get all states', () => {
+    const lagosState = AreaFactory.build({ state: 'lagos' });
+    const oyoState = AreaFactory.build({ state: 'oyo' });
+    beforeEach(async () => {
+      await addArea(lagosState);
+      await addArea(oyoState);
+    });
+
+    context('when editor token is used', () => {
```
andela-hpopoola, line 156: We need to add tests that the endpoint returns distinct states.

We will also need to add a test that adds 5 areas for 1 state and try to get the state so that we are sure that the state endpoint returns just 1 state (to be sure that it returns distinct state).

test/controllers/area.controller.test.js
```diff
@@ -148,4 +144,159 @@ describe('Area Controller', () => {
...
+          .end((err, res) => {
+            expect(res).to.have.status(500);
+            done();
+            Area.distinct.restore();
+          });
+      });
+    });
+  });
+
+  describe('Get all areas', () => {
```
andela-hpopoola, line 226: **Bug Alert:** Not sure how I missed this, we need to ensure that we add tests to ensure that the editor or admin cannot add an existing area in the same state.

This will cause lots of confusion in the application if we dont have the check. Sorry I missed it when reviewing your previous code 😞

test/controllers/contentProperty.controller.test.js
```diff
@@ -467,4 +467,80 @@ describe('Content Property Controller', () => {
...
+          .set('authorization', editorToken)
+          .end((err, res) => {
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body.houseTypes.length).to.be.eql(5);
+            done();
+          });
+      });
+    });
+
```
andela-hpopoola, line 490: Same comment with returning distinct states above.

We need to add tests that the endpoint returns distinct house Types.

andela-hpopoola (changes requested):
Thank you for your work on this. The implementation looks solid but I left a few comments to ensure to improve the code quality and ensure we don't run into future endpoint issues.

test/factories/area.factory.js
```diff
@@ -1,7 +1,7 @@
 import { Factory } from 'rosie';
 
 export default new Factory()
-  .sequence('area', (i) => `Lekki Phase ${i}`)
+  .sequence('area', (i) => `lekki phase ${i}`)
```
andela-hpopoola, line 4: Can we leave this as the normal case format since we are already converting the saved area and state to lowercase:

```suggestion
  .sequence('area', (i) => `Lekki Phase ${i}`)
```

server/services/area.service.js
```diff
@@ -4,24 +4,33 @@ import httpStatus from '../helpers/httpStatus';
...
-    const states = await Area.distinct('state');
-    return { states };
-  }
-  const areas = await Area.find({ state });
-  const area = [];
-  areas.forEach((a) => {
-    area.push({ [a.area]: a._id });
-  });
-  return { areas: area };
+export const getState = async () => {
```
andela-hpopoola, line 28: One last thing please, to make this method semantically correct, let's update it to the plural form; getStates not getState. 

```suggestion
export const getStates = async () => {
```

server/controllers/area.controllers.js
```diff
@@ -23,7 +23,7 @@ const AreaController = {
 
   getArea(req, res, next) {
     const { state } = req.params;
-    getStateAndArea(state)
+    getArea(state)
```
andela-hpopoola, line 26: Just to point out that we are retrieving multiple areas: The same applies to `getState` too 🙏 

```suggestion
    getAreas(state)
```

andela-hpopoola (approved):
Awesome job @tobieeniafe. This looks good to me, just a few nits to ensure that the code is correct grammatically. Feel free to merge when you fix the non-blocking comments. Nice work and Thank you.

test/factories/area.factory.js
tobieeniafe: this makes the test fail because the value exists in uppercase but the returned value is in lowercase

PR #76, opened 2020-11-17
#175634791 return properties info

tobieeniafe:
#### What does this PR do?
Retrieve searched content property for calculation

Retrieve searched content property for calculation

`GET /api/v1/content-property?areaId={areaId}& houseType={houseType}`

#175634791

### DIFF
--- a/test/controllers/contentProperty.controller.test.js
+++ b/test/controllers/contentProperty.controller.test.js
@@ -593,7 +593,7 @@ describe('Content Property Controller', () => {
       });
     });
 
-    context('when parameters areaId is sent', () => {
+    context('when the areaId parameter is given only', () => {
       it('returns evaluation of 8 properties', (done) => {
         request()
           .get(`/api/v1/content-property?areaId=${areaId}`)
@@ -612,7 +612,7 @@ describe('Content Property Controller', () => {
       });
     });
 
-    context('when parameters house type is sent', () => {
+    context('when the houseType parameter is given only', () => {
       it('returns not found', (done) => {
         request()
           .get(`/api/v1/content-property?houseType=${houseType}`)
--- a/test/controllers/contentProperty.controller.test.js
+++ b/test/controllers/contentProperty.controller.test.js
@@ -548,4 +548,78 @@ describe('Content Property Controller', () => {
       });
     });
   });
+
+  describe('Get Seach content properties calculation', () => {
+    const houseType = 'Terrace duplex';
+    const randomProperties = ContentPropertyFactory.buildList(5, { areaId });
+    const duplexProperties = ContentPropertyFactory.buildList(5, {
+      areaId,
+      houseType,
+      price: 10000000,
+    });
+    beforeEach(async () => {
+      await ContentProperty.insertMany(randomProperties);
+      await ContentProperty.insertMany(duplexProperties);
+    });
+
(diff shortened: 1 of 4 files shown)


andela-hpopoola (changes requested):
Thanks for working on this. I left a single comment on consistency, otherwise, you need to write the tests to ensure that this is good to go.

test/controllers/contentProperty.controller.test.js
```diff
@@ -548,4 +548,78 @@ describe('Content Property Controller', () => {
...
+      areaId,
+      houseType,
+      price: 10000000,
+    });
+    beforeEach(async () => {
+      await ContentProperty.insertMany(randomProperties);
+      await ContentProperty.insertMany(duplexProperties);
+    });
+
+    context('when parameters are sent', () => {
```
andela-hpopoola, line 565: Awesome. Howeever, it would be helpful if we can another test with different prices for the properties and test for that the calculation of minimumPrice, maximumPrice, and averagePrice is correct. 

Tip: You can add like three properties with prices of 100k, 200k, 300k to make the calculation easier to predict.

test/controllers/contentProperty.controller.test.js
```diff
@@ -548,4 +548,78 @@ describe('Content Property Controller', () => {
...
+            expect(res.body.evaluation.averagePrice).to.be.eql(10000000);
+            expect(res.body.evaluation.type).to.be.eql(houseType);
+            expect(res.body.evaluation.areaName).to.be.eql(area.area);
+            expect(res.body.evaluation.stateName).to.be.eql(area.state);
+            done();
+          });
+      });
+    });
+
+    context('without parameters', () => {
```
andela-hpopoola, line 584: I am curious, what is the expected behaviour when a valid areaId is given but the houseType is not in the parameters. Can we have a test for this to be sure of expected behaviour?

andela-hpopoola (changes requested):
Nice work. I left 2 comments on adding additional tests (since the services tests are missing). Once this has been resolved, we should be ready to merge this.

test/controllers/contentProperty.controller.test.js
```diff
@@ -581,6 +593,39 @@ describe('Content Property Controller', () => {
       });
     });
 
+    context('when parameters areaId is sent', () => {
```
andela-hpopoola, line 596:
```suggestion
    context('when the areaId parameter is given only', () => {
```

test/controllers/contentProperty.controller.test.js
```diff
@@ -581,6 +593,39 @@ describe('Content Property Controller', () => {
...
+            expect(res.body.evaluation.averagePrice).to.be.eql(3875000);
+            expect(res.body.evaluation).to.not.have.property('type');
+            expect(res.body.evaluation.areaName).to.be.eql(area.area);
+            expect(res.body.evaluation.stateName).to.be.eql(area.state);
+            done();
+          });
+      });
+    });
+
+    context('when parameters house type is sent', () => {
```
andela-hpopoola, line 615:
```suggestion
    context('when the houseType parameter is given only', () => {
```

andela-hpopoola (approved):
Thank you for your work on this. I left a few non blocking comments on ensuring the test assertions are more readable. Feel free to merge when you resolve them.

PR #78, opened 2020-11-19
#175635136 list all properties

tobieeniafe:
#### What does this PR do?
List all properties & add pagination to `/api/v1/content-property/all` endpoint

List all properties & add pagination to `/api/v1/content-property/all` endpoint

`GET /api/v1/content-property/all?page=1&limit=1`

`page` is the page number. Default is `1`
`limit` is the amount of documents per page. Default is `10`

#175635136

### DIFF
--- a/server/controllers/contentProperty.controllers.js
+++ b/server/controllers/contentProperty.controllers.js
@@ -3,6 +3,7 @@ import {
   updateContentProperty,
   deleteContentProperty,
   getHouseTypesByAreaId,
+  getPropertiesByParameters,
   getAllContentProperties,
 } from '../services/contentProperty.service';
 import httpStatus from '../helpers/httpStatus';
@@ -44,6 +45,15 @@ const ContentPropertyController = {
     getHouseTypesByAreaId(areaId)
       .then((houseTypes) => {
         res.status(httpStatus.OK).json({ success: true, houseTypes });
+      })
+      .catch((error) => next(error));
+  },
+
+  search(req, res, next) {
+    const { areaId, houseType } = req.query;
+    getPropertiesByParameters({ areaId, houseType })
+      .then((evaluation) => {
+        res.status(httpStatus.OK).json({ success: true, evaluation });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/contentProperty.controllers.js
+++ b/server/controllers/contentProperty.controllers.js
@@ -3,7 +3,7 @@ import {
   updateContentProperty,
   deleteContentProperty,
   getHouseTypesByAreaId,
-  getAllProperties,
+  getAllContentProperties,
 } from '../services/contentProperty.service';
 import httpStatus from '../helpers/httpStatus';
 
@@ -48,9 +48,9 @@ const ContentPropertyController = {
       .catch((error) => next(error));
   },
 
-  getAllProperties(req, res, next) {
+  getAllContentProperties(req, res, next) {
     const { page, limit } = req.query;
-    getAllProperties(page, limit)
+    getAllContentProperties(page, limit)
       .then(({ result, pagination }) => {
         res.status(httpStatus.OK).json({ success: true, pagination, result });
       })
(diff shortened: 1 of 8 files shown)


server/controllers/contentProperty.controllers.js
```diff
@@ -3,6 +3,7 @@ import {
   updateContentProperty,
   deleteContentProperty,
   getHouseTypesByAreaId,
+  getAllProperties,
```
andela-hpopoola, line 6: Not sure if `getAllContentProperties` will be more appropriate to differentiate this from the normal Properties 🤷‍♂️

server/services/contentProperty.service.js
```diff
@@ -74,3 +75,29 @@ export const getHouseTypesByAreaId = async (areaId) => {
...
+        as: 'areaInfo',
+      },
+    },
+    {
+      $unwind: '$areaInfo',
+    },
+    {
+      $facet: {
+        metadata: [{ $count: 'total' }, { $addFields: { page, limit } }],
+        data: [{ $skip: parseInt((page - 1) * limit, 10) }, { $limit: parseInt(limit, 10) }],
```
andela-hpopoola, line 95: Let's create a custom method that returns the array for data too to avoid repetition:

```suggestion
        data: generateFacetData(page, limit),
```

In `../helpers/pagination`, you will have something similar to:

```js
const generateFacetData = (page, limit) => [
    { $skip: parseInt((page - 1) * limit, 10) }, 
    { $limit: parseInt(limit, 10) }
];
```

test/controllers/contentProperty.controller.test.js
```diff
@@ -548,4 +549,147 @@ describe('Content Property Controller', () => {
...
+              expectsPaginationToReturnTheRightValues(res, defaultPaginationResult);
+              done();
+            });
+        });
+      });
+
+      context('when editor token is used', () => {
+        it('returns the given page and limit', (done) => {
+          request()
+            .get('/api/v1/content-property/all?page=2&limit=5')
```
andela-hpopoola, line 584: Since you have tested this above, you can use the one without page and limit to make the test simpler:

```suggestion
            .get('/api/v1/content-property/all')
```

test/controllers/contentProperty.controller.test.js
```diff
@@ -548,4 +549,147 @@ describe('Content Property Controller', () => {
...
+              });
+              done();
+            });
+        });
+      });
+
+      context('when admin token is used', () => {
+        it('returns the given page and limit', (done) => {
+          request()
+            .get('/api/v1/content-property/all?page=2&limit=5')
```
andela-hpopoola, line 603: Same comment above.

andela-hpopoola (changes requested):
This looks almost completed to me, thank you for working on this. I noticed that the pagination stuff and tests are easier to write based on the first one we did, Awesome! 

I left a few comments, this should be ready to merge when you resolve.

test/controllers/contentProperty.controller.test.js
```diff
@@ -667,4 +668,140 @@ describe('Content Property Controller', () => {
...
+    const defaultPaginationResult = {
+      currentPage: 1,
+      limit: 10,
+      offset: 0,
+      result: 10,
+      total: 18,
+      totalPage: 2,
+    };
+
+    describe('when properties exist in db', () => {
```
andela-hpopoola, line 687: [Nice to have] - We can encapsulate all the pagination tests in a function and call them with a different endpoint for different cases. This will save lots of time when implementing all pagination. Maybe a different PR to encapsulate so common tests you write to save more time.

test/controllers/contentProperty.controller.test.js
```diff
@@ -667,4 +668,140 @@ describe('Content Property Controller', () => {
       });
     });
   });
+
+  describe('Get all content properties', () => {
```
andela-hpopoola, line 672: Your tests only cover pagination, can you add tests to ensure that the right values are returned for content properties.

andela-hpopoola (approved):
Thank you for your hard work on this. Can you add tests to verify that the right content properties (that is the values of the returned objects) are valid to what is expected?  

Feel free to merge after adding the tests and fixing other comments.

PR #79, opened 2020-11-20
#175735113 add route to list all areas

tobieeniafe:
#### What does this PR do?
Add route to list all areas

Add route to list all areas with number of properties under them with corresponding minimum, maximum & average values of properties 

`GET /api/v1/area/all`

#175735113

### DIFF
--- a/server/routes/area.routes.js
+++ b/server/routes/area.routes.js
@@ -147,4 +147,27 @@ router.delete(
   AreaController.delete,
 );
 
+/**
+ * @swagger
+ * /area/:
+ *   get:
+ *     tags:
+ *       - Area
+ *     description: Get all areas in db with their properties info
+ *     produces:
+ *       - application/json
+ *     requestBody:
+ *      content:
+ *        application/json:
+ *          schema:
+ *            $ref: '#/components/schemas/Area'
+ *      description: Get all areas in db with their properties info
+ *     responses:
+ *      '200':
+ *        description: returns areas
+ *      '500':
+ *       description: Internal server error
+ */
+router.get('/', authenticate, isEditorOrAdmin, AreaController.getAllAreas);
+
 module.exports = router;
--- a/server/routes/area.routes.js
+++ b/server/routes/area.routes.js
@@ -66,29 +66,6 @@ router.get('/states', authenticate, isEditorOrAdmin, AreaController.getStates);
 
 /**
  * @swagger
- * /area/:state:
- *   get:
- *     tags:
- *       - Area
- *     description: Get all areas based on query parameters
- *     produces:
- *       - application/json
- *     requestBody:
- *      content:
- *        application/json:
- *          schema:
- *            $ref: '#/components/schemas/Area'
- *      description: Get all areas based on query parameters
- *     responses:
- *      '200':
- *        description: returns areas
- *      '500':
- *       description: Internal server error
- */
-router.get('/:state', authenticate, isEditorOrAdmin, AreaController.getAreas);
-
-/**
- * @swagger
  * /area/update:
  *   put:
  *     tags:
@@ -149,7 +126,7 @@ router.delete(
 
 /**
  * @swagger
- * /area/:
+ * /area/all:
  *   get:
  *     tags:
  *       - Area
@@ -168,6 +145,29 @@ router.delete(
  *      '500':
  *       description: Internal server error
  */
-router.get('/', authenticate, isEditorOrAdmin, AreaController.getAllAreas);
+router.get('/all', authenticate, isEditorOrAdmin, AreaController.getAllAreas);
+
+/**
+ * @swagger
+ * /area/:state:
+ *   get:
+ *     tags:
+ *       - Area
+ *     description: Get all areas based on query parameters
+ *     produces:
+ *       - application/json
+ *     requestBody:
+ *      content:
+ *        application/json:
+ *          schema:
+ *            $ref: '#/components/schemas/Area'
+ *      description: Get all areas based on query parameters
+ *     responses:
+ *      '200':
+ *        description: returns areas
+ *      '500':
+ *       description: Internal server error
+ */
+router.get('/:state', authenticate, isEditorOrAdmin, AreaController.getAreas);
 
 module.exports = router;
--- a/test/controllers/area.controller.test.js
+++ b/test/controllers/area.controller.test.js
@@ -662,11 +662,12 @@ describe('Area Controller', () => {
     });
   });
 
-  describe('Get all areas and correponding properties', () => {
-    const ajah = AreaFactory.build({ area: 'ajah' });
+  describe('Get all areas and corresponding properties', () => {
+    const ajah = AreaFactory.build({ area: 'Ajah' });
     const lekkiId = mongoose.Types.ObjectId();
-    const lekki = AreaFactory.build({ _id: lekkiId, area: 'lekki' });
-    const lekkiProperties = ContentPropertyFactory.buildList(5, {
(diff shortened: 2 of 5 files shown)


server/routes/area.routes.js
```diff
@@ -147,4 +147,27 @@ router.delete(
...
+ *          schema:
+ *            $ref: '#/components/schemas/Area'
+ *      description: Get all areas in db with their properties info
+ *     responses:
+ *      '200':
+ *        description: returns areas
+ *      '500':
+ *       description: Internal server error
+ */
+router.get('/', authenticate, isEditorOrAdmin, AreaController.getAllAreas);
```
andela-hpopoola, line 171: I believe this should follow the `all` convention for other endpoints
```suggestion
router.get('/all', authenticate, isEditorOrAdmin, AreaController.getAllAreas);
```

andela-hpopoola (changes requested):
This is looking good. Thank you for working on this. I have not tested this, but can you add tests while I test later in the day.

test/controllers/area.controller.test.js
```diff
@@ -661,4 +661,111 @@ describe('Area Controller', () => {
       });
     });
   });
+
+  describe('Get all areas and correponding properties', () => {
```
andela-hpopoola, line 665:
```suggestion
  describe('Get all areas and corresponding properties', () => {
```

test/controllers/area.controller.test.js
```diff
@@ -661,4 +661,111 @@ describe('Area Controller', () => {
...
+    const ajah = AreaFactory.build({ area: 'ajah' });
```
andela-hpopoola, line 666: Can we add this as a normal case since that is how it will be submitted in the form 🙏?  It looks weird writing the name of an area in lowercase.

```suggestion
    const ajah = AreaFactory.build({ area: 'Ajah' });
```

test/controllers/area.controller.test.js
```diff
@@ -661,4 +661,111 @@ describe('Area Controller', () => {
...
+    const lekkiId = mongoose.Types.ObjectId();
+    const lekki = AreaFactory.build({ _id: lekkiId, area: 'lekki' });
```
andela-hpopoola, line 668: Same here.

test/controllers/area.controller.test.js
```diff
@@ -661,4 +661,111 @@ describe('Area Controller', () => {
...
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body.areas.length).to.be.eql(2);
+            expect(res.body.areas[0].area).to.be.eql(ajah.area);
+            expect(res.body.areas[0].numOfProperties).to.be.eql(0);
+            expect(res.body.areas[0].minimumPrice).to.be.eql(null);
+            expect(res.body.areas[0].maximumPrice).to.be.eql(null);
+            expect(res.body.areas[0].averagePrice).to.be.eql(null);
+            expect(res.body.areas[1]._id).to.be.eql(lekkiId.toString());
+            expect(res.body.areas[1].area).to.be.eql(lekki.area);
```
andela-hpopoola, line 695: This should fix the changes
```suggestion
            expect(res.body.areas[1].area).to.be.eql(lekki.area.toLowercase());
```

test/controllers/area.controller.test.js
```diff
@@ -661,4 +661,111 @@ describe('Area Controller', () => {
...
+            expect(res.body.areas[1].numOfProperties).to.be.eql(5);
+            expect(res.body.areas[1].minimumPrice).to.be.eql(100000);
```
andela-hpopoola, line 697: I am not comfortable using the same value for the minimumPrice, maximumPrice and averagePrice. Can we use a varying price to be more certain that there is no bug with the implementation.

test/controllers/area.controller.test.js
```diff
@@ -661,4 +661,111 @@ describe('Area Controller', () => {
...
+            expect(res.body.areas[1].maximumPrice).to.be.eql(100000);
+            expect(res.body.areas[1].averagePrice).to.be.eql(100000);
+            done();
+          });
+      });
+    });
+
+    context('when admin token is used', () => {
```
andela-hpopoola, line 705: In the future, it is better to use just one test and use a function or array mapping since this test is the same as the one above. Something like this:

```js
[editorToken, adminToken].map((token) => {
it('returns all areas and properties', (done) => {
        request()
          .get('/api/v1/area/all')
          .set('authorization', token) // <---- the token is replaced here
          .end((err, res) => {
              ...
          }
});
```

Remember, you are a programmer, you need to be extremely creative 😉

test/controllers/area.controller.test.js
```diff
@@ -661,4 +661,111 @@ describe('Area Controller', () => {
...
+            expect(res.body.areas[1].numOfProperties).to.be.eql(5);
+            expect(res.body.areas[1].minimumPrice).to.be.eql(100000);
+            expect(res.body.areas[1].maximumPrice).to.be.eql(100000);
+            expect(res.body.areas[1].averagePrice).to.be.eql(100000);
+            done();
+          });
+      });
+    });
+
+    context('when user token is is used', () => {
```
andela-hpopoola, line 730: We need to right a function test for this test (and the 2 tests below) and replace all instances to save time. Maybe a later PR.

andela-hpopoola (changes requested):
Thank you for your work on this. I left a few comments but I believe this should be good to go after you fix. Feel free to ignore the ones for later PR (please don't close them as I am using them as a reminder for myself). Thank you.

test/controllers/area.controller.test.js
tobieeniafe: `context('when admin or editor token is used', () => {
      [editorToken, adminToken].map((token) => {
        it('returns all areas and properties', (done) => {
          request()
            .get('/api/v1/area/all')
            .set('authorization', token)
            .end((err, res) => {
              ...
              done();
            });
        });
      });
    });
`

returns undefined for authorization tho 😬

andela-hpopoola: I think I understand why that wasn't working, it's because the tokens are defined in the beforeEach block, so they are undefined before the `it block`. This can be resolved by defining the token in the`it block`.

Enough grammar, this should work though:

```js
    context('when a valid token is used', () => {
      [...new Array(2)].map((_, index) =>
        it('returns all areas and properties', (done) => {
          request()
            .get('/api/v1/area/all')
            .set('authorization', [editorToken, adminToken][index])
            .end((err, res) => {
              expect(res.body.areas.length).to.be.eql(2);
              expect(res.body.areas[0].area).to.be.eql(ajah.area.toLowerCase());
              expect(res.body.areas[0].numOfProperties).to.be.eql(0);
              expect(res.body.areas[0].minimumPrice).to.be.eql(null);
              expect(res.body.areas[0].maximumPrice).to.be.eql(null);
              expect(res.body.areas[0].averagePrice).to.be.eql(null);
              expect(res.body.areas[1]._id).to.be.eql(lekkiId.toString());
              expect(res.body.areas[1].area).to.be.eql(lekki.area.toLowerCase());
              expect(res.body.areas[1].numOfProperties).to.be.eql(5);
              expect(res.body.areas[1].minimumPrice).to.be.eql(100000);
              expect(res.body.areas[1].maximumPrice).to.be.eql(500000);
              expect(res.body.areas[1].averagePrice).to.be.eql(180000);
              done();
            });
        }),
      );
    });
```

andela-hpopoola (approved):
Thank you for your hard work. This should be good to go. Feel feel to fix the test comment so that you can use the same technique in similar tests.

PR #80, opened 2020-11-20
#175634768 add vendor object to user model

tobieeniafe:
#### What does this PR do?
Add vendor object to user model

Add vendor object to user model

NA

[Slides](https://docs.google.com/presentation/d/1z9UGMG_RK7okhGxkhpyeam36cZcZeXzMg0291QYxXLs/edit?ts=5fb76306#slide=id.ga26e5542d0_0_50)

#175634768

### DIFF
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -56,63 +56,25 @@ import { USER_ROLE } from '../helpers/constants';
 const { ObjectId } = mongoose.Schema.Types;
 const UserSchema = new mongoose.Schema(
   {
-    firstName: {
-      type: String,
-      required: true,
+    activated: {
+      type: Boolean,
+      default: false,
     },
-    lastName: {
-      type: String,
-      required: true,
-    },
-    vendor: {
-      companyName: {
+    activationDate: Date,
+    address: {
+      city: {
         type: String,
       },
-      companyAddress: {
+      country: {
         type: String,
       },
-      companyLogo: {
+      state: {
         type: String,
       },
-      identification: {
+      street1: {
         type: String,
       },
-      website: {
-        type: String,
-      },
-      accountNumber: {
-        type: String,
-      },
-      redanNumber: {
-        type: String,
-      },
-      directors: [
-        {
-          name: {
-            type: String,
-          },
-          isSignatory: {
-            type: Boolean,
-          },
-          signature: {
-            type: String,
-          },
-          phone: {
-            type: String,
-          },
-        },
-      ],
-      socialMedia: [
-        {
-          name: {
-            type: String,
-          },
-          url: {
-            type: String,
-          },
-        },
-      ],
-      vendorCode: {
+      street2: {
         type: String,
       },
     },
@@ -121,46 +83,16 @@ const UserSchema = new mongoose.Schema(
       unique: true,
       required: true,
     },
-    password: {
+    favorites: {
+      type: [ObjectId],
+    },
+    firstName: {
       type: String,
       required: true,
-      select: false,
     },
-    phone: String,
-    phone2: String,
-    address: {
-      street1: {
-        type: String,
-      },
-      street2: {
-        type: String,
-      },
-      city: {
-        type: String,
-      },
-      state: {
-        type: String,
-      },
-      country: {
-        type: String,
-      },
-    },
-    preferences: {
-      houseType: {
-        type: String,
-      },
-      location: {
-        type: String,
-      },
-      maxPrice: {
-        type: Number,
-      },
-      minPrice: {
-        type: Number,
-      },
-      paymentPlan: {
-        String,
-      },
+    lastName: {
+      type: String,
+      required: true,
     },
     notifications: [
       {
@@ -179,6 +111,34 @@ const UserSchema = new mongoose.Schema(
         dateAdded: Date,
       },
     ],
+    password: {
+      type: String,
+      required: true,
+      select: false,
+    },
+    phone: String,
+    phone2: String,
+    preferences: {
+      houseType: {
+        type: String,
+      },
+      location: {
+        type: String,
+      },
+      maxPrice: {
+        type: Number,
+      },
+      minPrice: {
+        type: Number,
+      },
+      paymentPlan: {
+        String,
+      },
+    },
+    profileImage: {
+      id: String,
+      url: String,
+    },
     referralCode: {
       type: String,
       unique: true,
@@ -187,18 +147,58 @@ const UserSchema = new mongoose.Schema(
       type: Number,
       default: USER_ROLE.USER,
     },
-    profileImage: {
-      id: String,
-      url: String,
+    vendor: {
+      accountNumber: {
+        type: String,
+      },
+      companyAddress: {
+        type: String,
+      },
+      companyLogo: {
+        type: String,
+      },
+      companyName: {
+        type: String,
+      },
+      directors: [
+        {
+          name: {
+            type: String,
+          },
+          isSignatory: {
+            type: Boolean,
+          },
+          signature: {
+            type: String,
+          },
+          phone: {
+            type: String,
+          },
+        },
+      ],
+      identification: {
+        type: String,
+      },
+      redanNumber: {
+        type: String,
+      },
+      socialMedia: [
+        {
+          name: {
+            type: String,
+          },
+          url: {
+            type: String,
+          },
+        },
+      ],
+      vendorCode: {
+        type: String,
+      },
+      website: {
+        type: String,
+      },
     },
-    favorites: {
-      type: [ObjectId],
-    },
-    activated: {
-      type: Boolean,
-      default: false,
-    },
-    activationDate: Date,
   },
   { timestamps: true },
 );
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -71,6 +71,9 @@ const UserSchema = new mongoose.Schema(
       companyAddress: {
         type: String,
       },
+      companyLogo: {
+        type: String,
+      },
       identification: {
         type: String,
       },
@@ -85,10 +88,16 @@ const UserSchema = new mongoose.Schema(
       },
       directors: [
         {
-          firstName: {
+          name: {
             type: String,
           },
-          lastName: {
+          isSignatory: {
+            type: Boolean,
+          },
+          signature: {
+            type: String,
+          },
+          phone: {
             type: String,
           },
         },
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -63,6 +63,14 @@ const UserSchema = new mongoose.Schema(
       type: String,
       required: true,
     },
+    vendor: {
+      companyName: {
+        type: String,
+      },
+      companyAddress: {
+        type: String,
+      },
+    },
     email: {
       type: String,
       unique: true,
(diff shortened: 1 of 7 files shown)


andela-hpopoola (changes requested):
To make this useful, can you update the registration endpoint to support vendor registration. I think the logic is to check if the logic is to check if the payload contains an optional companyName, then save the payload as a vendor.

For the required firstName and lastName, I believe they should still be required, however, from the frontend the firstName will be the supplied company name and the last name will be 'Company'. This is nothing you should be bothered about in the endpoint, just explaining to clear your doubts.

server/schemas/user.schema.js
```diff
@@ -9,10 +9,39 @@ const confirmPassword = Joi.string().valid(Joi.ref('password')).required().stric
...
+  accountNumber: optionalString('Account Number'),
+  redanNumber: optionalString('Redan Number'),
+  directors,
+  socialMedia,
+});
 
 export const registerSchema = Joi.object({
   firstName: requiredString('First Name'),
   lastName: requiredString('Last Name'),
+  vendor,
```
andela-hpopoola, line 44: Awesome work. Unfortunately, this should contain only the `companyName`, the vendor is adding everything after registration (not before). 

Notwithstanding, the `vendor` schema might be useful for user updates. So it is not a wasted effort.

server/models/user.model.js
```diff
@@ -63,6 +63,55 @@ const UserSchema = new mongoose.Schema(
...
+      companyName: {
+        type: String,
+      },
+      companyAddress: {
+        type: String,
+      },
+      identification: {
+        type: String,
+      },
+      phone: {
```
andela-hpopoola, line 76: The phone number and email address are not needed here. The existing one in the user model will be used to avoid confusion.

andela-hpopoola (changes requested):
Thank you for your updates. I left a few comments. Can you work on the logic for vendor registration as stated here:

> To make this useful, can you update the registration endpoint to support vendor registration. I think the logic is to check if the payload contains an optional `companyName`, then save the payload as a vendor with the company name.

To add more context, when anyone wants to register as a vendor, instead of asking them for the firstName and lastName in the frontend, we will ask them for the company name. Every other thing is the same as the normal registration.

server/models/user.model.js
```diff
@@ -63,6 +64,49 @@ const UserSchema = new mongoose.Schema(
...
+      website: {
+        type: String,
+      },
+      accountNumber: {
+        type: String,
+      },
+      redanNumber: {
+        type: String,
+      },
+      directors: [
```
andela-hpopoola, line 86: This should be improved based on the latest Figma design:

We need the following IMHO,
```
Name - required string
Phone Number - optional string
isSignatory - true / false
signature - optional string
```

server/schemas/user.schema.js
```diff
@@ -9,10 +9,41 @@ const confirmPassword = Joi.string().valid(Joi.ref('password')).required().stric
 });
 const phone = Joi.string().allow(null, '').optional().default('');
 const requiredString = (label) => Joi.string().label(label).required();
+const optionalString = (label) => Joi.string().label(label).optional();
+
+// const directors = [
```
andela-hpopoola, line 14: Can we remove the comments here, you can always come back to this commit here on GitHub to copy them if you need them later in the future.

server/services/user.service.js
```diff
@@ -71,7 +71,8 @@ export const generateReferralCode = async (firstName) => {
   return referralCode;
 };
 
-export const addUser = async (user) => {
+export const addUser = async (newUser) => {
+  let user = newUser;
```
andela-hpopoola, line 75: I don't think this is needed. We can check if the vendor is present and update roles instead of updating the user object.

server/services/user.service.js
```diff
@@ -94,6 +95,10 @@ export const addUser = async (user) => {
     }
   }
 
+  if (user.vendor && user.vendor.companyName) {
```
andela-hpopoola, line 98: I think it is better to update the role:

```js
const role = user.vendor && user.vendor.companyName ? USER_ROLE.VENDOR : USER_ROLE.USER;

**Review by andela-hpopoola — CHANGES_REQUESTED** [review]
Thank you for your hard work here. I left a couple of comments. Next is the test to wrap this up asap.

File: server/models/user.model.js [hunk 7]
```diff
@@ -63,6 +64,49 @@ const UserSchema = new mongoose.Schema(
       type: String,
       required: true,
     },
+    vendor: {
+      companyName: {
```
andela-hpopoola, line 68: You will also need to add the company logo to the vendor schema.

server/services/user.service.js
tobieeniafe: adding this line breaks the test across all models. can't seem to find the reason

andela-hpopoola: Thank you for pointing that out. It is kind of obvious that the error is from the fact that admins are created from the test and this is setting the default to User. So the tests for admins will fail (I believe you should have figured it out since the error messages from the test clearly point to this). 

This should work rather:

```js
  const role = user.vendor && user.vendor.companyName ? USER_ROLE.VENDOR : newUser.role;
```

server/models/user.model.js
```diff
@@ -63,6 +64,58 @@ const UserSchema = new mongoose.Schema(
       type: String,
       required: true,
     },
+    vendor: {
```
andela-hpopoola, line 67: Can we move this down a little bit so that the newly added `vendor` object is arranged alphabetically with the existing schema. Let it come before the timestamp. It is a good practise to arrange object values alphabetically (makes it easier to detect a value quickly and easily detect duplicate / existing values).

andela-hpopoola (changes requested):

server/services/user.service.js
tobieeniafe: `newUser` is undefined tho. for some reason, it only works when the `user` object is modified

andela-hpopoola: Of course, `newUser` will be undefined now since you have removed `newUser` from the code. Previously you had something like this:

```js
const user = newUser; // <----- where newUser was defined in the method parameter 
```

Since you have removed this and decided to use `user` directly, so `user.role` will be appropriate. The purpose is just to get the role that is sent over when the user is created.

Please note that the codes I write in PRs are for illustrative purposes only, I believe you should be able to figure the exact code once you understand the intention behind the change. Most times I don't test them, while at other times they would be affected by the changes you have made (always keep that in mind please) 

So this will work:

```js
  const role = user.vendor && user.vendor.companyName ? USER_ROLE.VENDOR : newUser.role;
```

server/models/user.model.js
```diff
@@ -109,43 +135,70 @@ const UserSchema = new mongoose.Schema(
...
+        {
+          name: {
+            type: String,
+          },
+          url: {
+            type: String,
+          },
+        },
+      ],
+      vendorCode: {
```
andela-hpopoola, line 195: If we are adding the vendorCode, I think it is a good idea to generate it during registration. What do you think?

andela-hpopoola (changes requested):
Thank you for the updates and arranging the vendor object alphabetically. Still expecting the tests to wrap this up.

server/models/user.model.js
tobieeniafe: remember we mentioned something about vendor code being unique to each vendor during offer letter generation? So I just left it till we made a decision

andela-hpopoola: Yes, we can add a number in front of the vendor code to differentiate them as they would be unique. Let's defer this to a later PR then and ask the content team the best way we would generate the vendor code.

server/services/user.service.js
```diff
@@ -71,8 +71,10 @@ export const generateReferralCode = async (firstName) => {
   return referralCode;
 };
 
-export const addUser = async (user) => {
+export const addUser = async (newUser) => {
   let referrer;
+  const user = newUser;
```
andela-hpopoola, line 76: I think you got my comment here slightly wrong, this is no longer needed. I was referring to why I used newUser instead of `user`. Let's revert line 74 to `user` and remove this line. It is not doing anything.

test/controllers/user.controller.test.js
```diff
@@ -70,6 +70,31 @@ describe('User Controller', () => {
...
+      context('when user registers as a vendor ', () => {
+        it('returns successful token', (done) => {
+          const vendor = UserFactory.build({ vendor: { companyName: 'Highrachy Investment' } });
+          request()
+            .post('/api/v1/user/register')
+            .send(vendor)
+            .end((err, res) => {
+              expect(res).to.have.status(201);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body.message).to.be.eql('User registered');
```
andela-hpopoola, line 83: I think we should update this:

```suggestion
              expect(res.body.message).to.be.eql('Company registered');
```

test/controllers/user.controller.test.js
```diff
@@ -70,6 +70,31 @@ describe('User Controller', () => {
...
+              expect(res.body).to.have.property('token');
+              expect(sendMailStub.callCount).to.eq(1);
+              expect(sendMailStub).to.have.be.calledWith(
+                EMAIL_CONTENT.ACTIVATE_YOUR_ACCOUNT,
+                vendor,
```
andela-hpopoola, line 88: Vendor uses `companyName` instead of `firstName` so the logic here can cause unexpected bug; A random firstName is used here but it is supposed to use the companyName (as FirstName), We have 2 options in the code though.

1. We can assign companyName to firstName when saving and ensure we update the firstName when the companyName has been updated. This makes the vendor object to work flawlessly with the mail.
2. Create a logic in the mail service to cater for companyName if it is present. The advantage of this is that once it detects that the passed-in object has a `vendor.companyName` value, it uses that instead of firstName.

I prefer the second one though since it is more explicit and handles the vendor directly instead of changing the concept of the vendor.

andela-hpopoola (changes requested):
Thank you for the hardwork here. I left a few comments but I am more concerned on ensuring that we pull out the company name when we are sending a mail to the vendor. Once that is fixed, I believe this PR is good to go.

test/controllers/user.controller.test.js
tobieeniafe: my approach to this causes an error tho

server/services/mailer.service.js
```diff
@@ -61,7 +61,7 @@ export const sendMail = (defaultMailContent, user, additionalOptions = {}) => {
   const options = {
     ...defaultMailContent,
     ...additionalOptions,
-    firstName: user.firstName || null,
+    firstName: user.firstName || user.vendor.companyName ? user.vendor.companyName : null,
```
andela-hpopoola, line 64: I am not sure this will work as expected since `firstName` is a required field so it will always be present and the `companyName` will never be used here:

Not 100% sure too, but this looks better:
```suggestion
    firstName: user.vendor && user.vendor.companyName ? user.vendor.companyName : user.firstName || null,
```

server/services/user.service.js
```diff
@@ -111,6 +111,7 @@ export const addUser = async (newUser) => {
       });
     }
     return generateToken(savedUser._id);
+    // return { token: generateToken(savedUser._id), userType: isVendor ? 'Company' : 'User' };
```
andela-hpopoola, line 114: Not sure why you commented this out though

andela-hpopoola (changes requested):
We are almost there with this PR. Thank you for the hard work.

andela-hpopoola (approved):
Thank you for the hard work here. This looks good to me.

PR #81, opened 2020-11-23
#175846239 add routes to return single content property and area

tobieeniafe:
#### What does this PR do?
Add routes to return single content property and area

Add routes to return single content property and area

`GET /api/v1/content-property/:id`
`GET /api/v1/area/:id`

#175846239

### DIFF
--- a/server/controllers/area.controllers.js
+++ b/server/controllers/area.controllers.js
@@ -4,7 +4,7 @@ import {
   getAreas,
   updateArea,
   deleteArea,
-  getAreaById,
+  getAreaAndContentPropertiesByAreaId,
   getAllAreas,
 } from '../services/area.service';
 import httpStatus from '../helpers/httpStatus';
@@ -66,12 +66,12 @@ const AreaController = {
 
   getAreaById(req, res, next) {
     const { id } = req.params;
-    getAreaById(id)
+    getAreaAndContentPropertiesByAreaId(id)
       .then((area) => {
-        if (!area) {
+        if (area.length < 1) {
           res.status(httpStatus.NOT_FOUND).json({ success: false, message: 'Area not found' });
         } else {
-          res.status(httpStatus.OK).json({ success: true, area });
+          res.status(httpStatus.OK).json({ success: true, area: area[0] });
         }
       })
       .catch((error) => next(error));
--- a/server/controllers/area.controllers.js
+++ b/server/controllers/area.controllers.js
@@ -1,4 +1,11 @@
-import { addArea, getStates, getAreas, updateArea, deleteArea } from '../services/area.service';
+import {
+  addArea,
+  getStates,
+  getAreas,
+  updateArea,
+  deleteArea,
+  getAreaById,
+} from '../services/area.service';
 import httpStatus from '../helpers/httpStatus';
 
 const AreaController = {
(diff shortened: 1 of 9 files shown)


server/controllers/area.controllers.js
```diff
@@ -62,6 +63,19 @@ const AreaController = {
       })
       .catch((error) => next(error));
   },
+
+  getAreaById(req, res, next) {
+    const { id } = req.params;
+    getAreaById(id)
+      .then((area) => {
+        if (area === null) {
```
andela-hpopoola, line 71:
```suggestion
        if (!area) {
```

server/services/property.service.js
```diff
@@ -148,9 +148,9 @@ export const getAllProperties = async () =>
   ]);
 
 // get a property by its id and admin details
-export const getOneProperty = async (propertId) =>
+export const getOneProperty = async (propertyId) =>
```
andela-hpopoola, line 151: 🙌 Thank you for fixing this.

andela-hpopoola (changes requested):
Thank you for working on this. This is looking good. Can you add the tests while I take another look at it tomorrow?

server/controllers/area.controllers.js
```diff
@@ -62,6 +63,19 @@ const AreaController = {
       })
       .catch((error) => next(error));
   },
+
+  getAreaById(req, res, next) {
+    const { id } = req.params;
+    getAreaById(id)
```
andela-hpopoola, line 69: Can we get all the content properties associated with an area and maybe the maximumPrice, miminumPrice and averagePrice of all the properties. This will make the area information more productive for anyone consuming the endpoint.

andela-hpopoola (changes requested):
Thank you for making these updates. I left a few comments but the most important ones are:

- Add content properties to returned areas
- Ensure we add test assertions for all returned values

server/controllers/area.controllers.js
```diff
@@ -4,7 +4,7 @@ import {
   getAreas,
   updateArea,
   deleteArea,
-  getAreaById,
+  getAreaAndContentPropertiesByAreaId,
```
andela-hpopoola, line 7: Not sure why you changed this though, I feel `getAreaById` is a better name though since the focus is in getting on the details pertaining to the area 🤷‍♂️

test/controllers/area.controller.test.js
```diff
@@ -679,4 +679,83 @@ describe('Area Controller', () => {
       });
     });
   });
+
+  describe('Get area by id', () => {
```
andela-hpopoola, line 683: Can you add an extra test for an area without content properties, please 🙏

andela-hpopoola (changes requested):
Thank you for the hard work here. This looks great, and your implementation looks brilliant. Waiting to you to resolve the merge conflict and fix the minor comments I added so that I can approve this. Nice work and Thank you once again.

server/controllers/area.controllers.js
tobieeniafe: `getAreaById` is an existing function that uses the `findById` method

andela-hpopoola: Makes perfect sense to me. Let's leave it like that. Thank you for the explanation.

andela-hpopoola (approved):
This looks good to me. Thank you for making the changes. Nice work. You can merge it asap.

PR #82, opened 2020-11-23
#175845002 make routes for searching content properties public

tobieeniafe:
#### What does this PR do?
Make routes for searching content properties public

Make routes for searching content properties public

`GET api/v1/area/states`
`GET api/v1/area/:state`
`GET /api/v1/content-property/area/:id`
`GET /api/v1/content-property?areaId={areaId}& houseType={houseType}`

#175845002

### DIFF
--- a/test/controllers/area.controller.test.js
+++ b/test/controllers/area.controller.test.js
@@ -162,44 +162,14 @@ describe('Area Controller', () => {
       await Area.insertMany(areasInOyoState);
     });
 
-    context('when a valid token is used', () => {
-      [...new Array(2)].map((_, index) =>
-        it('returns array of two states', (done) => {
-          request()
-            .get('/api/v1/area/states')
-            .set('authorization', [editorToken, adminToken][index])
-            .end((err, res) => {
-              expect(res).to.have.status(200);
-              expect(res.body.success).to.be.eql(true);
-              expect(res.body.states.length).to.be.eql(2);
-              done();
-            });
-        }),
-      );
-    });
-
-    context('when user token is is used', () => {
-      it('returns forbidden', (done) => {
-        request()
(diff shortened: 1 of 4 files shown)


test/controllers/area.controller.test.js
```diff
@@ -162,44 +162,14 @@ describe('Area Controller', () => {
       await Area.insertMany(areasInOyoState);
     });
 
-    context('when a valid token is used', () => {
```
andela-hpopoola, line 165: Can you re-add this test, please? It ensures that logged-in users (that is the editors and admin) too can access the endpoints with no issues.

test/controllers/area.controller.test.js
```diff
@@ -227,44 +196,14 @@ describe('Area Controller', () => {
       await Area.insertMany(areas);
     });
 
-    context('when a valid token is used', () => {
```
andela-hpopoola, line 230: Readd this too, please?

test/controllers/area.controller.test.js
```diff
@@ -227,44 +196,14 @@ describe('Area Controller', () => {
...
-              expect(res).to.have.status(200);
-              expect(res.body.success).to.be.eql(true);
-              expect(res.body.areas.length).to.be.eql(5);
-              done();
-            });
-        }),
-      );
-    });
-
-    context('when user token is is used', () => {
```
andela-hpopoola, line 246: You can remove this, we have no business with the user token accessing the endpoints even though it should return valid values (we simply don't care if it doesn't work for a signed-in user)

andela-hpopoola (approved):
Thank you for working on this. Just a few comments on reading the tests for admin and editor token (just to be sure that they return the right value even if these guys access it with their token).

PR #83, opened 2020-11-25
#175868402 #175868469 make area search by name & fix case sensitivity

tobieeniafe:
#### What does this PR do?
Fix content property search to search by area and not areaId
Allowed search parameters to be case insensitive

Fix content property search to search by area and not areaId
Allowed search parameters to be case insensitive

`GET /api/v1/content-property/search?area={area}&type={type}&state={state}`

#175868402 #175868469

### DIFF
--- a/server/services/contentProperty.service.js
+++ b/server/services/contentProperty.service.js
@@ -97,11 +97,7 @@ export const getPropertiesByParameters = async ({ area, state, houseType }) => {
   const prices = await ContentProperty.aggregate([
     {
       $match: {
-        $and: [
-          { houseType: houseType || /.*/ },
-          // { houseType: houseType ? new RegExp(['^', houseType, '$'], 'i') : /.*/ },
-          { areaId: ObjectId(areaId) || /.*/ },
-        ],
+        $and: [{ houseType: houseType || /.*/ }, { areaId: ObjectId(areaId) || /.*/ }],
       },
     },
     {
--- a/server/services/contentProperty.service.js
+++ b/server/services/contentProperty.service.js
@@ -3,7 +3,7 @@ import ContentProperty from '../models/contentProperty.model';
 import { ErrorHandler } from '../helpers/errorHandler';
 import httpStatus from '../helpers/httpStatus';
 // eslint-disable-next-line import/no-cycle
-import { getAreaById } from './area.service';
+import { getAreaById, getAreaByAreaAndStateName } from './area.service';
 import { generatePagination, generateFacetData } from '../helpers/pagination';
 
 const { ObjectId } = mongoose.Types.ObjectId;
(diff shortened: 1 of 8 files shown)


andela-hpopoola (changes requested):
This looks good. Thank you for working on this. Can you add more tests to ensure that this work for multiple cases? Let's do this for all the case-insensitivity endpoints that we have (area, state, content property, and search).

You can create an array of cases and loop through them for the tests. Something like this (just an idea, you need to flesh it out yourself).

```js
// 1. define the areaName
const areaName = 'Ikoyi'
const areaNameInDifferentCases ['Ikoyi', 'IKOYI', 'IKOyi', 'ikoyi'];

...

// 2. Loop through a single test with the different cases

areaNameInDifferentCases.map((area) => {

// 3. normal assertions here
...
request()
 .get(`/api/v1/content-property/search?area=${area.area}&state=${area}`)
...
// 4. test cases here
})



```

server/services/contentProperty.service.js
```diff
@@ -79,19 +79,29 @@ export const getHouseTypesByAreaId = async (areaId) => {
...
 
+  const areaId = areaExists[0]._id;
+
   const prices = await ContentProperty.aggregate([
     {
       $match: {
-        $and: [{ houseType: houseType || /.*/ }, { areaId: ObjectId(areaId) || /.*/ }],
+        $and: [
+          { houseType: houseType || /.*/ },
+          // { houseType: houseType ? new RegExp(['^', houseType, '$'], 'i') : /.*/ },
```
andela-hpopoola, line 102: Can we remove this please 🙏

test/controllers/contentProperty.controller.test.js
```diff
@@ -506,10 +508,33 @@ describe('Content Property Controller', () => {
       });
     });
 
-    context('when the areaId parameter is given only', () => {
+    context('when areas are sent in different cases', () => {
+      const areaNameInDifferentCases = ['Ikoyi', 'IKOYI', 'IKOyi', 'ikoyi'];
+      [...new Array(4)].map((_, index) =>
```
andela-hpopoola, line 513: This is more preferable:

```suggestion
     areaNameInDifferentCases.map((_, index) =>
```

andela-hpopoola (approved):
Feel free to merge this. I didn't properly test it on my local system so you should not be surprised if I find more bugs in this later. I have adequate faith that it should work as expected else we can roll it back or fix the bug later. Thank you for your hard work.

test/controllers/contentProperty.controller.test.js
```diff
@@ -506,10 +508,33 @@ describe('Content Property Controller', () => {
...
+        it('returns array of house types', (done) => {
+          request()
+            .get(
+              `/api/v1/content-property/search?area=${areaNameInDifferentCases[index]}&state=${area.state}&houseType=${houseType}`,
```
andela-hpopoola, line 517: If you fix the one above, you should be able to use areaName here
```suggestion
              `/api/v1/content-property/search?area=${areaName}&state=${area.state}&houseType=${houseType}`,
```

PR #84, opened 2020-11-29
#175868569 prevent return of areas without content properties

tobieeniafe:
#### What does this PR do?
Prevent return of areas without content properties

Prevent return of areas without content properties

`GET api/v1/area/state/{state}`

#175868569

### DIFF
--- a/server/controllers/area.controllers.js
+++ b/server/controllers/area.controllers.js
@@ -1,14 +1,13 @@
 import {
   addArea,
   getStates,
-  getAreas,
+  getAreasByState,
   updateArea,
   deleteArea,
   getAreaAndContentPropertiesByAreaId,
   getAllAreas,
 } from '../services/area.service';
 import httpStatus from '../helpers/httpStatus';
-import { decodeToken } from '../services/user.service';
 
 const AreaController = {
   add(req, res, next) {
@@ -32,10 +31,8 @@ const AreaController = {
 
   getAreas(req, res, next) {
     const { state } = req.params;
-    const adminOrEditorId = req.headers.authorization
-      ? decodeToken(req.headers.authorization).id
-      : null;
-    getAreas({ state, adminOrEditorId })
+    const tokenIsPresent = !!req.headers.authorization;
+    getAreasByState({ state, tokenIsPresent })
       .then((areas) => {
         res.status(httpStatus.OK).json({ success: true, areas: areas.areas });
       })
--- a/server/controllers/area.controllers.js
+++ b/server/controllers/area.controllers.js
@@ -8,6 +8,7 @@ import {
   getAllAreas,
 } from '../services/area.service';
 import httpStatus from '../helpers/httpStatus';
+import { decodeToken } from '../services/user.service';
 
 const AreaController = {
   add(req, res, next) {
@@ -31,7 +32,10 @@ const AreaController = {
 
   getAreas(req, res, next) {
     const { state } = req.params;
-    getAreas(state)
+    const adminOrEditorId = req.headers.authorization
+      ? decodeToken(req.headers.authorization).id
+      : null;
+    getAreas({ state, adminOrEditorId })
       .then((areas) => {
         res.status(httpStatus.OK).json({ success: true, areas: areas.areas });
       })
--- a/server/services/area.service.js
+++ b/server/services/area.service.js
@@ -4,8 +4,6 @@ import { ErrorHandler } from '../helpers/errorHandler';
 import httpStatus from '../helpers/httpStatus';
 // eslint-disable-next-line import/no-cycle
 import { getContentPropertyByAreaId } from './contentProperty.service';
-import { getUserById } from './user.service';
-import { USER_ROLE } from '../helpers/constants';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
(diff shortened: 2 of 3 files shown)


andela-hpopoola:
Thank you for working on this. I think you got the story wrong (even though I explained it when I shared my screen and when I presented on Friday). The purpose is not to return properties with areas but to ensure that for the frontend we return areas that have content properties only. Your implementation is a feature, not a bug fix (sometimes you might need to ask when a story is a bug and the bug it is fixing). 

Let me explain better the bug better:

1. Let's assume we have 2 areas in Lagos in the database - Lekki and Agege
2. Lekki has 2 content properties but Agege has no assigned content property. Here is the illustration:
```
Lekki has 2 Bungalow (2 Bedrooms)
Agege has 0 -- No content property
```

3. In the search form here -- http://appstaging.ballers.ng/, there are 2 scenarios for a visitor to move from `State -> Area -> HouseType`:
   a. The valid selection 
```
The visitor selects 1. Lagos --> 2. Lekki --> 3. Bungalow (2 Bedrooms).
```
From above the visitor will have no issue navigating the application to the search result page.

   b. The invalid selection 
```
The visitor selects 1. Lagos --> 2. Agege --> ????
```
From above, the visitor will be stuck with Agege since Agege has no content property. We could use a message that Agege has no Content Property, but that is a very bad User Experience (keep in mind that the app loads the content property for each area selected and the experience might not be too friendly for users with a slower internet connection).

For, maximum User Experience, it is better to hide the areas with no content property for this form and show all the areas on the administrator page. That way, the visitor doesn't even know that we have `Agege` in the database since they cannot get any useful information from it.

I hope you understand now.

andela-hpopoola:
Feel free to close this PR and implement the bug fix in another PR since the code here is not related to the expected bug fix. Thank you.

andela-hpopoola (changes requested):
Sorry for the poor description!!! Please always reach out when you don't understand a story on Pivotal Tracker to prevent working on a different thing entirely.

server/services/area.service.js
```diff
@@ -37,6 +37,17 @@ export const getStates = async () => {
 
 export const getAreas = async (state) => {
   const areas = await Area.find({ state }).collation({ locale: 'en', strength: 2 });
+  await Promise.all(
+    areas.map(async (area) => {
+      const houseType = await getHouseTypesByAreaId(area._id);
```
andela-hpopoola, line 42: Unfortunately, this is an O(n^2) implementation. It is costly to be retrieving houseTypes individually for each area, so it is better to try another implementation.

server/services/area.service.js
```diff
@@ -37,6 +37,17 @@ export const getStates = async () => {
 
 export const getAreas = async (state) => {
   const areas = await Area.find({ state }).collation({ locale: 'en', strength: 2 });
+  await Promise.all(
```
andela-hpopoola, line 40: I think a better implementation here will be to aggregate the content properties with the area. 

In normal SQL, I would add a `where contentProperty is NOT null` to remove areas with no content property (currently not sure if this is possible in NoSQL). If this is not possible, then you can loop through the result to remove the ones with content property is NULL. That is much faster and better than calling the database for each area.

Also, note that this behaviour is only expected in the frontend and but not needed in admin side because if we are removing all the area with no content property, there will be no way for the admin to add a new content property (because when he adds an area, this endpoint will not load the area since the newly added area will have no content property).

To get beyond this, it is better to add a query parameter like `showAll`, if this is true, then it will not filter the result else it will filter the result. The default for this is for `showAll` to be true. Another way is to check if there is a token in the request, if there is a token, `showAll` should be true else it should be false. The ball is in your court.

test/controllers/area.controller.test.js
```diff
@@ -207,35 +207,43 @@ describe('Area Controller', () => {
...
-      await Area.insertMany(areas);
+      await addArea(epe);
+      await Area.insertMany(randomAreas);
+      await ContentProperty.insertMany(contentProperties);
     });
 
     context('when a valid token is used', () => {
       [...new Array(2)].map((_, index) =>
-        it('returns array of five areas', (done) => {
+        it('returns array of one area', (done) => {
```
andela-hpopoola, line 223: From the explanation above, we would want this to return five areas for an admin/editor. We can use the test for `no token` to test that the public endpoint returns just one area.

andela-hpopoola (changes requested):
Thank you for your hard work on this and for the fast turnaround in implementing the story the right way. A left a few comments on ensuring the service performs a dual role of return all areas for admin/editor and just the needed area for requests with no token.

server/services/area.service.js
tobieeniafe: any suggestions?

andela-hpopoola: I left a few suggestions in this comment  

> I think a better implementation here will be to aggregate the content properties with the area.

https://github.com/Highrachy/ballers-api/pull/84#discussion_r532401306

server/services/area.service.js
```diff
@@ -35,8 +37,38 @@ export const getStates = async () => {
   return { states };
 };
 
-export const getAreas = async (state) => {
+export const getAreas = async ({ state, adminOrEditorId }) => {
   const areas = await Area.find({ state }).collation({ locale: 'en', strength: 2 });
```
andela-hpopoola, line 41: I believe we can make use of the linkedProperties query below instead of getting the areas 2 times.

server/controllers/area.controllers.js
```diff
@@ -31,7 +32,10 @@ const AreaController = {
 
   getAreas(req, res, next) {
     const { state } = req.params;
-    getAreas(state)
+    const adminOrEditorId = req.headers.authorization
```
andela-hpopoola, line 35: I appreciate your thoroughness here, however, since the area information doesn't have any sensitive information, we can keep this simple to check if only the authorization token exist only:

```suggestion
    const tokenIsPresent = !!req.headers.authorization;
```

server/services/area.service.js
```diff
@@ -35,8 +37,38 @@ export const getStates = async () => {
...
+
+  const user = await getUserById(adminOrEditorId);
```
andela-hpopoola, line 43: This is no longer needed since we don't care about the validity of the token. That saves us an extra query request 😉

server/services/area.service.js
```diff
@@ -35,8 +37,38 @@ export const getStates = async () => {
   return { states };
 };
 
-export const getAreas = async (state) => {
+export const getAreas = async ({ state, adminOrEditorId }) => {
```
andela-hpopoola, line 40: I believe we can rewrite the whole of this method in a more effective way as done below; Comments are added to explain the thought process, please remove comments before pushing your changes.

```js
// Changed method name to getAreasByState to make it more specific. getAreas is too generic and doesn't explain the essence of this method
export const getAreasByState = async ({ state, tokenIsPresent }) => {
// Used just 1 query to avoid unnecessary requests to the database
  const linkedProperties = await Area.aggregate([
    { $match: { state } },
    {
      $lookup: {
        from: 'contentproperties',
        localField: '_id',
        foreignField: 'areaId',
        as: 'linkedProperties',
      },
    },
  ]).collation({ locale: 'en', strength: 2 });

// Used reduce so that I can format the returned array with the valud of `_id` and `state`.
  const areas = linkedProperties.reduce((acc, area) => {
   // if the tokenIsPresent, it will return everything thing. That is why `OR` is used
  // if token is not present, then it will return only the areas with linked properties greater than 0
    if (tokenIsPresent || area.linkedProperties.length > 0) {
      acc.push({ _id: area._id, area: area.area });
    }
    return acc;
  }, []);

  return { areas };
};

```

andela-hpopoola: I tested this and saw that you might need to updated `area.controller.test.js:256` to aggregate to make the other tests pass:

```js
    context('when getStateAndArea service fails', () => {
      it('returns the error', (done) => {
        sinon.stub(Area, 'aggregate').throws(new Error('Type Error'));
        request()
          .get(`/api/v1/area/state/${state}`)
          .end((err, res) => {
            expect(res).to.have.status(500);
            done();
            Area.aggregate.restore();
          });
      });
    });
  });
```

andela-hpopoola (changes requested):
Thank you for the updates. I left a few code suggestions to prevent us from spending too much time on this. I believe this should be ready for merging once the suggestions have been implemented.

andela-hpopoola (approved):
Thank you for the great work on this. Feel free to merge 💪 .

PR #85, opened 2020-12-02
#175928810 load vendor info for verification

tobieeniafe:
#### What does this PR do?
Load vendor info for verification

Load vendor info for verification

`GET /api/v1/user/vendor/all`

#175928810

### DIFF
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -23,7 +23,7 @@ export const defaultPaginationResult = {
   totalPage: 2,
 };
 
-export const itReturnsTheRightPaginationValue = (token, endpoint) => {
+export const itReturnsTheRightPaginationValue = (endpoint, token) => {
   context('when no pagination parameter is passed', () => {
     it('returns the default values', (done) => {
       request()
@@ -90,12 +90,12 @@ export const itReturnsTheRightPaginationValue = (token, endpoint) => {
   });
 };
 
-export const itReturnsForbiddenForInvalidToken = (invalidToken, endpoint) => {
+export const itReturnsForbiddenForInvalidToken = (endpoint, token) => {
   context('with a invalid access token', () => {
     it('returns forbidden', (done) => {
       request()
         .get(endpoint)
-        .set('authorization', invalidToken)
+        .set('authorization', token)
         .end((err, res) => {
           expect(res).to.have.status(403);
           expect(res.body.success).to.be.eql(false);
@@ -121,7 +121,12 @@ export const itReturnsForbiddenForNoToken = (endpoint) => {
   });
 };
 
-export const itReturnsAnErrorWhenServiceFails = (token, endpoint, model, modelMethod) => {
+export const itReturnsAnErrorWhenServiceFails = (
+  endpoint,
+  token,
+  model,
+  modelMethod = 'aggregate',
+) => {
   context('when service fails', () => {
     it('returns the error', (done) => {
       sinon.stub(model, modelMethod).throws(new Error('Type Error'));
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -1,4 +1,4 @@
-import { expect } from './config';
+import { expect, request, sinon } from './config';
 
 export const expectsPaginationToReturnTheRightValues = (
   res,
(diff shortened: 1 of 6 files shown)


andela-hpopoola (changes requested):
Thank you for working on this. This looks good. Can you try updating the pagination tests as functions as stated in the story description on Pivotal Tracker?

> ...
Also, can you try to customize the pagination tests as a function so that we don't waste time repeating the same tests over and over again?
...

tobieeniafe:
> Thank you for working on this. This looks good. Can you try updating the pagination tests as functions as stated in the story description on Pivotal Tracker?
> 
> > ...
> > Also, can you try to customize the pagination tests as a function so that we don't waste time repeating the same tests over and over again?
> > ...

I was able to do this, but I kept running into an `undefined` error on the `adminToken` & `userToken`

test/helpers.js
```diff
@@ -22,3 +22,111 @@ export const defaultPaginationResult = {
   total: 18,
   totalPage: 2,
 };
+
+export const paginationTest = ({ validToken, invalidToken, endpoint, model, modelMethod }) => {
```
andela-hpopoola, line 26:
```suggestion
export const itReturnsTheRightPaginationValue = ({ validToken, invalidToken, endpoint, model, modelMethod }) => {
```

test/helpers.js
```diff
@@ -22,3 +22,111 @@ export const defaultPaginationResult = {
...
+  context('when no parameters are passed', () => {
```
andela-hpopoola, line 27:
```suggestion
  context('when no pagination parameter is passed', () => {
```

test/helpers.js
```diff
@@ -22,3 +22,111 @@ export const defaultPaginationResult = {
...
+            limit: 4,
+            result: 4,
+            totalPage: 5,
+          });
+          done();
+        });
+    });
+  });
+
+  context('with a invalid access token', () => {
```
andela-hpopoola, line 92: Let's write another function for an invalid access token and not tie them to pagination tests. The reason is that we might want to paginate for public endpoints that would be valid when no token is added. Something like `itReturnsForBiddenForInvalidToken`

Also, a separate function for invalid tokens would ensure that we can replace similar tests in the other test files.

test/helpers.js
```diff
@@ -22,3 +22,111 @@ export const defaultPaginationResult = {
...
+        .end((err, res) => {
+          expect(res).to.have.status(403);
+          expect(res.body.success).to.be.eql(false);
+          expect(res.body.message).to.be.eql('You are not permitted to perform this action');
+          done();
+        });
+    });
+  });
+
+  context('without token', () => {
```
andela-hpopoola, line 106: Same comment as above on creating another function for this too. The new function name can be something like `itReturnsAnErrorWhenNoTokenIsPresent`.

test/helpers.js
```diff
@@ -22,3 +22,111 @@ export const defaultPaginationResult = {
...
+        .end((err, res) => {
+          expect(res).to.have.status(403);
+          expect(res.body.success).to.be.eql(false);
+          expect(res.body.message).to.be.eql('Token needed to access resources');
+          done();
+        });
+    });
+  });
+
+  context('when service fails', () => {
```
andela-hpopoola, line 119: Same comment as above on creating another function for this too. The new function name can be something like `itReturnsAnErrorWhenServiceFails`.

test/helpers.js
andela-hpopoola, line 26: Separating the pagination tests into independent tests should reduce the parameters used to 3:

```suggestion
export const itReturnsTheRightPaginationValue= ({ endpoint, model, token }) => {
```

test/helpers.js
```diff
@@ -22,3 +22,111 @@ export const defaultPaginationResult = {
...
+    it('returns the error', (done) => {
+      sinon.stub(model, `${modelMethod}`).throws(new Error('Type Error'));
```
andela-hpopoola, line 121: Not sure why you added this as a string, I believe we can remove the string interpolation:

```suggestion
      sinon.stub(model, modelMethod).throws(new Error('Type Error'));
```

andela-hpopoola (changes requested):
Thank you for the great work. Great thinking and implementation. Once, we are through with this, it would reduce the number of tests we write, the size of our tests file and improve the time wasted in writing recurring tests.

I left a few comments of separating some of the tests so that we will be able to reuse them for similar stuff. This is to save time later as the same function can be reused for other tests in the codebase.

test/helpers.js
```diff
@@ -23,12 +23,12 @@ export const defaultPaginationResult = {
   totalPage: 2,
 };
 
-export const paginationTest = ({ validToken, invalidToken, endpoint, model, modelMethod }) => {
-  context('when no parameters are passed', () => {
+export const itReturnsTheRightPaginationValue = (token, endpoint) => {
```
andela-hpopoola, line 26: Can we ensure that the `endpoint` comes first before token since endpoint is needed for all of the functions you have created. From Javascript documentation, that is how similar functions are treated to make it easier to remember the parameter order.

```js
itReturnsTheRightPaginationValue(endpoint, token)
itReturnsForbiddenForInvalidToken(endpoint, token)
itReturnsForbiddenForNoToken(endpoint)
itReturnsAnErrorWhenServiceFails(endpoint, token, model, modelMethod)
```

test/helpers.js
```diff
@@ -115,13 +119,15 @@ export const paginationTest = ({ validToken, invalidToken, endpoint, model, mode
         });
     });
   });
+};
 
+export const itReturnsAnErrorWhenServiceFails = (token, endpoint, model, modelMethod) => {
```
andela-hpopoola, line 124: We can set a default here for modelMethod since it is setup for 'aggregate' to ensure that we can omit the last parameter when we use this function for aggregate.

```suggestion
export const itReturnsAnErrorWhenServiceFails = (token, endpoint, model, modelMethod = "aggregate")
```

test/helpers.js
```diff
@@ -88,7 +88,9 @@ export const paginationTest = ({ validToken, invalidToken, endpoint, model, mode
         });
     });
   });
+};
 
+export const itReturnsForbiddenForInvalidToken = (invalidToken, endpoint) => {
```
andela-hpopoola, line 93: Let's change the `invalidToken` here to `token` for consistency's sake. The function name already specifies that the token is invalid (and also the test will fail if you use a valid token)

andela-hpopoola (approved):
Thank you for the updates on this. I left a few more comments on ensuring that the functions are consistent to make it easier to use for a new developer. Feel free to merge after making the changes.

PR #86, opened 2020-12-03
#175634775 allow admin to approve vendor information

tobieeniafe:
#### What does this PR do?
Allow admin to approve vendor information

Allow admin to approve vendor information

`PUT /api/v1/user/vendor/verify`
`PUT /api/v1/user/vendor/verify/comment`
`PUT /api/v1/user/vendor/verify/step`

#175634775

### DIFF
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -13,6 +13,9 @@ import {
   getAccountOverview,
   upgradeUserToEditor,
   downgradeEditorToUser,
+  verifyCompanyInfo,
+  verifyBankDetails,
+  verifyDirectorInfo,
 } from '../services/user.service';
 import { sendMail } from '../services/mailer.service';
 import EMAIL_CONTENT from '../../mailer';
@@ -187,6 +190,42 @@ const UserController = {
         res
           .status(httpStatus.OK)
           .json({ success: true, message: 'Content Editor is now a User', user });
+      })
+      .catch((error) => next(error));
+  },
+
+  verifyCompanyInfo(req, res, next) {
+    const { vendorId } = req.locals;
+    const adminId = req.user._id;
+    verifyCompanyInfo({ vendorId, adminId })
+      .then((vendor) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Company information verified', vendor });
+      })
+      .catch((error) => next(error));
+  },
+
+  verifyBankDetails(req, res, next) {
+    const { vendorId } = req.locals;
+    const adminId = req.user._id;
+    verifyBankDetails({ vendorId, adminId })
+      .then((vendor) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Bank information verified', vendor });
+      })
+      .catch((error) => next(error));
+  },
+
+  verifyDirectorInfo(req, res, next) {
+    const { vendorId } = req.locals;
+    const adminId = req.user._id;
+    verifyDirectorInfo({ vendorId, adminId })
+      .then((vendor) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Director information verified', vendor });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -13,9 +13,7 @@ import {
   getAccountOverview,
   upgradeUserToEditor,
   downgradeEditorToUser,
-  verifyCompanyInfo,
-  verifyBankDetails,
-  verifyDirectorInfo,
+  verifyVendor,
 } from '../services/user.service';
 import { sendMail } from '../services/mailer.service';
 import EMAIL_CONTENT from '../../mailer';
@@ -194,38 +192,14 @@ const UserController = {
       .catch((error) => next(error));
   },
 
-  verifyCompanyInfo(req, res, next) {
-    const { vendorId } = req.locals;
+  verifyVendor(req, res, next) {
+    const vendorInfo = req.locals;
     const adminId = req.user._id;
-    verifyCompanyInfo({ vendorId, adminId })
+    verifyVendor({ ...vendorInfo, adminId })
       .then((vendor) => {
         res
           .status(httpStatus.OK)
           .json({ success: true, message: 'Company information verified', vendor });
-      })
-      .catch((error) => next(error));
-  },
-
-  verifyBankDetails(req, res, next) {
-    const { vendorId } = req.locals;
-    const adminId = req.user._id;
-    verifyBankDetails({ vendorId, adminId })
-      .then((vendor) => {
-        res
-          .status(httpStatus.OK)
-          .json({ success: true, message: 'Bank information verified', vendor });
-      })
-      .catch((error) => next(error));
-  },
-
-  verifyDirectorInfo(req, res, next) {
-    const { vendorId } = req.locals;
-    const adminId = req.user._id;
-    verifyDirectorInfo({ vendorId, adminId })
-      .then((vendor) => {
-        res
-          .status(httpStatus.OK)
-          .json({ success: true, message: 'Director information verified', vendor });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -215,7 +215,7 @@ const UserController = {
       .then((vendor) => {
         res
           .status(httpStatus.OK)
-          .json({ success: true, message: 'Company information verified', vendor });
+          .json({ success: true, message: 'Vendor information verified', vendor });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -13,8 +13,10 @@ import {
   getAccountOverview,
   upgradeUserToEditor,
   downgradeEditorToUser,
-  verifyVendor,
+  verifyVendorStep,
   getAllVendors,
+  addCommentToVerificationStep,
+  verifyVendor,
 } from '../services/user.service';
 import { sendMail } from '../services/mailer.service';
 import EMAIL_CONTENT from '../../mailer';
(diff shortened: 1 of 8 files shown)


server/controllers/user.controllers.js
```diff
@@ -190,6 +193,42 @@ const UserController = {
       })
       .catch((error) => next(error));
   },
+
+  verifyCompanyInfo(req, res, next) {
```
andela-hpopoola, line 197: Instead of duplicating the same service 3 times, we can actually use just 1 service and update the needed field (companyInfo, bankDetails, directorInfo) as necessary.

Here is my thought process:
1. Create 1 endpoint that accepts any given step as a parameter
2. Create an object that contains all the given steps.  This is used to know if the given parameter in step 1 is valid.
3. For valid step names, verify the information as necessary (by substituting the step name when needed).

server/helpers/constants.js
```diff
@@ -31,3 +31,9 @@ export const CONCERN_STATUS = {
   RESOLVED: 'Resolved',
   PENDING: 'Pending',
 };
+
+export const VENDOR_VERIFICATION_INFO_STATUS = {
```
andela-hpopoola, line 35:
```suggestion
export const VENDOR_INFO_STATUS = {
```

server/helpers/constants.js
```diff
@@ -31,3 +31,9 @@ export const CONCERN_STATUS = {
...
+  PENDING: 'Pending',
+  VERIFIED: 'Verified',
+  NOT_VERIFIED: 'Not Verified',
```
andela-hpopoola, line 38:
```suggestion
  IN_REVIEW: 'In Review', // the vendor is waiting for the admin
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments but I am particularly concerned about removing code duplication and abstracting the step endpoints to accommodate new steps without creating a new endpoint.

From the story on Pivotal tracker, we would also need to add another endpoint to add comments for each of the steps. 

Also, we would need one last endpoint to update the global `verified` object for the vendor. That is the `verified` object that we would use to know if a vendor is verified or not.

server/services/user.service.js
```diff
@@ -371,12 +371,12 @@ export const getAllVendors = async (page = 1, limit = 10) => {
   return { pagination, result };
 };
 
-export const verifyVendor = async ({ vendorId, adminId, step }) => {
+export const verifyVendorStep = async ({ vendorId, adminId, step }) => {
   const validSteps = ['companyInfo', 'bankDetails', 'directorInfo'];
```
andela-hpopoola, line 375: Lets define this in constant.js file so that we can have one source of truth for it when we want to update.

```js
const VENDOR_STEPS = ['companyInfo', 'bankDetails', 'directorInfo']
```

server/services/user.service.js
```diff
@@ -404,3 +404,65 @@ export const verifyVendor = async ({ vendorId, adminId, step }) => {
     throw new ErrorHandler(httpStatus.BAD_REQUEST, `Error verifying ${step}`, error);
   }
 };
+
+export const addCommentToVerificationStep = async ({ vendorId, adminId, step, comment }) => {
+  const validSteps = ['companyInfo', 'bankDetails', 'directorInfo'];
```
andela-hpopoola, line 409: This should be removed. We need one source of truth for this 🙏

server/services/user.service.js
```diff
@@ -404,3 +404,65 @@ export const verifyVendor = async ({ vendorId, adminId, step }) => {
...
+  const stepIsValid = validSteps.includes(step);
+
+  if (!stepIsValid) {
+    throw new ErrorHandler(httpStatus.PRECONDITION_FAILED, 'User not found');
```
andela-hpopoola, line 413:
```js
    throw new ErrorHandler(httpStatus.PRECONDITION_FAILED, 'Invalid step');
```

server/controllers/user.controllers.js
```diff
@@ -204,6 +207,38 @@ const UserController = {
...
   },
+
+  verifyVendorStep(req, res, next) {
+    const verificationInfo = req.locals;
+    const adminId = req.user._id;
+    verifyVendorStep({ ...verificationInfo, adminId })
+      .then((vendor) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Company information verified', vendor });
```
andela-hpopoola, line 218: We are not sure if the vendor is  a company or an individual
```suggestion
          .json({ success: true, message: 'Vendor information verified', vendor });
```

server/models/user.model.js
```diff
@@ -208,6 +208,74 @@ const UserSchema = new mongoose.Schema(
...
+                type: String,
+              },
+              addedBy: {
+                type: ObjectId,
+              },
+            },
+          ],
+          status: {
+            type: String,
+            default: VENDOR_INFO_STATUS.IN_REVIEW,
```
andela-hpopoola, line 225: default should be PENDING 🙏 
```suggestion
            default: VENDOR_INFO_STATUS.PENDING,
```

Here is the breakdown:

```
PENDING: 'Pending',   -> We are expecting information from the user
IN_REVIEW: 'In Review', -> The user has submitted info, we are currently reviewing it. If a comment is added, that this should go back to PENDING
VERIFIED: 'Verified', -> We have approved the Step.
```

server/models/user.model.js
```diff
@@ -208,6 +208,74 @@ const UserSchema = new mongoose.Schema(
...
```
andela-hpopoola, line 247: Same comment as above

server/models/user.model.js
```diff
@@ -208,6 +208,74 @@ const UserSchema = new mongoose.Schema(
...
+          },
+          verifiedBy: {
+            type: ObjectId,
+          },
+          verifiedOn: {
+            type: Date,
+          },
+        },
+        bankDetails: {
```
andela-hpopoola, line 234: Instead of duplicating the steps schema for each step, lets create the schema in a variable and assign them to the steps.

Something like:

```js
const vendorInfoSchema = {
...
}

...

verification: {
        companyInfo: vendorInfoSchema,
        bankDetails: vendorInfoSchema,
        directorInfo: vendorInfoSchema
}
...
```

This way, we would get rid of code duplication and save lots of time if we need to update the schema.

andela-hpopoola (changes requested):
Thank you for the awesome work here. This looks better and closer to completion. I left a few comments on code duplication. Once you add the tests, this should be almost ready to merge.

PS: I noticed that we should be able to merge the comments and step verification step services together, however, I decided to leave it as you have implemented it so that I don't over complicate it for you and to also save time.

server/services/user.service.js
```diff
@@ -369,3 +370,99 @@ export const getAllVendors = async (page = 1, limit = 10) => {
...
+
+  if (!vendor) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'User not found');
+  }
+
+  if (vendor && vendor.role !== USER_ROLE.VENDOR) {
+    throw new ErrorHandler(httpStatus.PRECONDITION_FAILED, 'User is not a vendor');
+  }
+
+  const commentInfo = {
```
andela-hpopoola, line 426: Adding a comment should set the status of the step to PENDING. As explained above, `PENDING` indicates that we are expecting information from the user.

server/routes/user.routes.js
```diff
@@ -483,4 +486,115 @@ router.put(
  */
 router.get('/vendor/all', authenticate, isAdmin, UserController.getAllVendors);
 
+/**
+ * @swagger
+ * /user/vendor/verify:
+ *   put:
+ *     tags:
+ *       - User
+ *     description: Allows an admin to verify a vendor's
```
andela-hpopoola, line 495:
```suggestion
 *     description: Allows an admin to verify a vendor
```

test/helpers.js
```diff
@@ -24,7 +24,7 @@ export const defaultPaginationResult = {
   totalPage: 2,
 };
 
-export const itReturnsTheRightPaginationValue = (endpoint, user) => {
+export const itReturnsTheRightPaginationValue = (endpoint, method, user, data = {}) => {
```
andela-hpopoola, line 27: Nice work here. 

One of the best practices for parameters is to make them an object parameter when they are 3. Let's make the parameters an object to make it easier to remember and be consistent. I don't mind leaving it like this but I am afraid that it will not be consistent with `itReturnsAnErrorForInvalidToken` which now has 6 parameters. We might even update it later to accommodate more edge cases. 

To fix this and future problems, let's make this and other functions an object parameter. That way this function is easier to maintain and update before the place we have used them become too much to update.

```suggestion
export const itReturnsTheRightPaginationValue = ({endpoint, method, user, data}) => {
```

test/helpers.js
```diff
@@ -138,9 +144,11 @@ export const itReturnsForbiddenForNoToken = (endpoint) => {
 
 export const itReturnsAnErrorWhenServiceFails = (
```
andela-hpopoola, line 145: This has 6 parameters. It will be difficult to maintain and remember all as normal parameters. Let's make this and others object parameters as discussed above.

server/services/user.service.js
```diff
@@ -369,3 +370,100 @@ export const getAllVendors = async (page = 1, limit = 10) => {
...
+  if (!vendor) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'User not found');
+  }
+
+  if (vendor && vendor.role !== USER_ROLE.VENDOR) {
+    throw new ErrorHandler(httpStatus.PRECONDITION_FAILED, 'User is not a vendor');
+  }
+
+  try {
+    return User.findByIdAndUpdate(
```
andela-hpopoola, line 455: One more thing here, let's ensure that the other steps are verified but we term the vendor as verified. 

For example, currently, we have 3 steps` 'companyInfo', 'bankDetails', 'directorInfo'`, the ideal thing to do before marking a vendor as verified is to check if the 3 Steps above is marked as verified. 

My implementation would be to run a loop through the vendor step and check all the values of the `VENDOR_STEPS` array has `verified: true` in the vendor object gotten from line 444. I would personally want to use the `every` prototype for this but feel free to implement this as you deem fit.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

andela-hpopoola (changes requested):
Thank you for the great work here. You did an awesome job with the tests and show that you are already growing as a developer tremendously. Nice work.

I left a few comments to ensure that we get this perfectly right. Thank you and Nice work once again.

andela-hpopoola (approved):
This looks good to me. Feel free to merge when you have resolved the comments I left. Thank you for your hard work.

PR #87, opened 2020-12-06
#175634772 vendor update information for verification

tobieeniafe:
#### What does this PR do?
Allows a  vendor to update their information for verification

Allows a  vendor to update their information for verification

#175634772

### DIFF
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -14,6 +14,7 @@ import {
   upgradeUserToEditor,
   downgradeEditorToUser,
   getAllVendors,
+  updateVendor,
 } from '../services/user.service';
 import { sendMail } from '../services/mailer.service';
 import EMAIL_CONTENT from '../../mailer';
@@ -201,6 +202,18 @@ const UserController = {
           pagination,
           result,
         });
+      })
+      .catch((error) => next(error));
+  },
+
+  updateVendor(req, res, next) {
+    const updatedVendor = req.locals;
+    const vendorId = req.user._id;
+    updateVendor({ updatedVendor, vendorId })
+      .then((vendor) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Vendor information updated', vendor });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -245,10 +245,10 @@ const UserController = {
     const updatedVendor = req.locals;
     const vendorId = req.user._id;
     updateVendor({ updatedVendor, vendorId })
-      .then((vendor) => {
+      .then((user) => {
         res
           .status(httpStatus.OK)
-          .json({ success: true, message: 'Vendor information updated', vendor });
+          .json({ success: true, message: 'Vendor information updated', user });
       })
       .catch((error) => next(error));
   },
--- a/server/schemas/user.schema.js
+++ b/server/schemas/user.schema.js
@@ -11,14 +11,26 @@ const phone = Joi.string().allow(null, '').optional().default('');
 const requiredString = (label) => Joi.string().label(label).required();
 const optionalString = (label) => Joi.string().label(label).optional();
 
-const vendor = Joi.object().keys({
-  companyName: optionalString('Company Name'),
-});
+const directors = [
+  {
+    name: optionalString('Name'),
+    isSignatory: Joi.boolean().label('Signatory').optional(),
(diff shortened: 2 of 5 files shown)


server/schemas/user.schema.js
```diff
@@ -11,14 +11,26 @@ const phone = Joi.string().allow(null, '').optional().default('');
 const requiredString = (label) => Joi.string().label(label).required();
 const optionalString = (label) => Joi.string().label(label).optional();
 
-const vendor = Joi.object().keys({
-  companyName: optionalString('Company Name'),
-});
+const directors = [
+  {
+    name: optionalString('Name'),
+    isSignatory: Joi.boolean().label('Signatory').optional(),
```
andela-hpopoola, line 17: This can be abstracted into a function like the optionalString function above. 

```js
const optionalBoolean = (label) => Joi.boolean().label(label).optional();
```

We might need to abstract these functions into a `schema-helper.js` file to ensure that we can easily reuse them later.

andela-hpopoola (changes requested):
Nice implementation. I think we need the test for this so that we can merge asap.

server/schemas/user.schema.js
```diff
@@ -12,21 +12,21 @@ const requiredString = (label) => Joi.string().label(label).required();
 const optionalString = (label) => Joi.string().label(label).optional();
 const optionalBoolean = (label) => Joi.boolean().label(label).optional();
 
-const directors = [
-  {
+const directors = Joi.array().items(
```
andela-hpopoola, line 15: Nice one. Seeing the changes made me realize that the change was necessary. It looks like the test forced you to catch this 😛

test/controllers/user.controller.test.js
```diff
@@ -1666,6 +1666,112 @@ describe('User Controller', () => {
         });
       });
     });
+
+    describe('Add comment to vendor info', () => {
```
andela-hpopoola, line 1670:
```suggestion
    describe('Update vendor info', () => {
```

server/controllers/user.controllers.js
```diff
@@ -239,6 +240,18 @@ const UserController = {
...
   },
+
+  updateVendor(req, res, next) {
+    const updatedVendor = req.locals;
+    const vendorId = req.user._id;
+    updateVendor({ updatedVendor, vendorId })
+      .then((vendor) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Vendor information updated', vendor });
```
andela-hpopoola, line 251: Let's return this as a user so that we can have something like `user.vendor` and not `vendor.vendor` to get vendor specific information. 
```suggestion
          .json({ success: true, message: 'Vendor information updated', user });
```

It will also make the tests more readable. Having 2 vendors can be confusing:
```js
expect(res.body.vendor.vendor.companyAddress).to.be.eql(data.companyAddress);
```

Please note that some tests might fail when you update this so ensure you update the tests too.

test/controllers/user.controller.test.js
```diff
@@ -1666,6 +1666,112 @@ describe('User Controller', () => {
...
+              expect(res.body.vendor.vendor.companyName).to.be.eql(vendorUser.vendor.companyName);
+              expect(res.body.vendor.vendor.accountNumber).to.be.eql(data.accountNumber);
+              expect(res.body.vendor.vendor.companyAddress).to.be.eql(data.companyAddress);
+              expect(res.body.vendor.vendor.socialMedia.length).to.be.eql(data.socialMedia.length);
+              expect(res.body.vendor.vendor.directors.length).to.be.eql(data.directors.length);
+              done();
+            });
+        });
+      });
+
```
andela-hpopoola, line 1736: Can we add another test to be sure that we can update each of the fields one by one. Here is the implementation I came up with:

```js
      context('updating a single field', () => {
        Object.keys(data).map((field) => {
          it('returns updated vendor', (done) => {
            request()
              [method](endpoint)
              .set('authorization', vendorToken)
              .send({ [field]: data[field] })
              .end((err, res) => {
                expect(res.body.message).to.be.eql('Vendor information updated');
                expect(res.body.vendor._id).to.be.eql(vendorId.toString());
                expect(res.body.vendor.vendor).to.have.property(field);
                done();
              });
          });
        });
      });
```

Note: I used `.to.have.property` because you have already tested for the right values in the test above 😉

andela-hpopoola (approved):
Thank you for working on this. I am approving this because I have resolved the most important test. Feel free to merge when you fixed everything. Thank you and Nice work.

PR #88, opened 2020-12-08
#176029266 update visitation for vendor

tobieeniafe:
#### What does this PR do?
Update visitation schedule for vendor

Update visitation schedule for vendor

`POST /api/v1/visitation/schedule`

#176029266

### DIFF
--- a/server/controllers/visitation.controllers.js
+++ b/server/controllers/visitation.controllers.js
@@ -3,25 +3,23 @@ import EMAIL_CONTENT from '../../mailer';
 import { sendMail } from '../services/mailer.service';
 import httpStatus from '../helpers/httpStatus';
 
-const BALLERS_EMAIL = process.env.BALLERS_EMAIL || 'dev@highrachy.com';
-
 const VisitationController = {
   book(req, res, next) {
     const booking = req.locals;
     const { user } = req;
     scheduleVisitation({ ...booking, userId: user._id })
-      .then((schedule) => {
+      .then(({ newSchedule, vendorEmail }) => {
         const contentBottom = `
-          <strong> Name: </strong>: ${schedule.visitorName}<br /> 
-          <strong> Phone: </strong> ${schedule.visitorPhone}<br />
+          <strong> Name: </strong>: ${newSchedule.visitorName}<br /> 
+          <strong> Phone: </strong> ${newSchedule.visitorPhone}<br />
           <strong> Email: </strong> ${
-            schedule.visitorEmail ? schedule.visitorEmail : 'No email provided'
+            newSchedule.visitorEmail ? newSchedule.visitorEmail : 'No email provided'
           } <br />
         `;
-        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: BALLERS_EMAIL }, { contentBottom });
+        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: vendorEmail }, { contentBottom });
         res
           .status(httpStatus.CREATED)
-          .json({ success: true, message: 'Visit scheduled successfully', schedule });
+          .json({ success: true, message: 'Visit scheduled successfully', schedule: newSchedule });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/visitation.controllers.js
+++ b/server/controllers/visitation.controllers.js
@@ -25,16 +25,11 @@ const VisitationController = {
   },
 
   getAll(req, res, next) {
+    const { page, limit } = req.query;
     const { user } = req;
-    getAllVisitations(user)
-      .then((schedules) => {
-        if (schedules.length > 0) {
-          res.status(httpStatus.OK).json({ success: true, schedules });
-        } else {
-          res
-            .status(httpStatus.NOT_FOUND)
-            .json({ success: false, message: 'No schedules available', schedules });
-        }
+    getAllVisitations(user, page, limit)
+      .then(({ result, pagination }) => {
+        res.status(httpStatus.OK).json({ success: true, pagination, result });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/visitation.controllers.js
+++ b/server/controllers/visitation.controllers.js
@@ -10,7 +10,7 @@ const VisitationController = {
     scheduleVisitation({ ...booking, userId: user._id })
       .then(({ schedule, vendor }) => {
         const contentBottom = `
-          <strong> Name: </strong>: ${schedule.visitorName}<br /> 
+          <strong> Name: </strong>: ${schedule.visitorName}<br />
           <strong> Phone: </strong> ${schedule.visitorPhone}<br />
           <strong> Email: </strong> ${
             schedule.visitorEmail ? schedule.visitorEmail : 'No email provided'
--- a/server/controllers/visitation.controllers.js
+++ b/server/controllers/visitation.controllers.js
@@ -8,7 +8,7 @@ const VisitationController = {
     const booking = req.locals;
     const { user } = req;
     scheduleVisitation({ ...booking, userId: user._id })
-      .then(({ schedule, vendorEmail }) => {
+      .then(({ schedule, vendor }) => {
         const contentBottom = `
           <strong> Name: </strong>: ${schedule.visitorName}<br /> 
           <strong> Phone: </strong> ${schedule.visitorPhone}<br />
@@ -16,7 +16,7 @@ const VisitationController = {
             schedule.visitorEmail ? schedule.visitorEmail : 'No email provided'
           } <br />
         `;
-        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: vendorEmail }, { contentBottom });
+        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, vendor, { contentBottom });
         res
           .status(httpStatus.CREATED)
           .json({ success: true, message: 'Visit scheduled successfully', schedule });
@@ -25,15 +25,15 @@ const VisitationController = {
   },
 
   getAll(req, res, next) {
-    const userId = req.user._id;
-    getAllVisitations(userId)
+    const { user } = req;
+    getAllVisitations(user)
       .then((schedules) => {
         if (schedules.length > 0) {
           res.status(httpStatus.OK).json({ success: true, schedules });
         } else {
           res
             .status(httpStatus.NOT_FOUND)
-            .json({ success: false, message: 'No schedules available' });
+            .json({ success: false, message: 'No schedules available', schedules });
         }
       })
       .catch((error) => next(error));
--- a/server/controllers/visitation.controllers.js
+++ b/server/controllers/visitation.controllers.js
@@ -8,25 +8,33 @@ const VisitationController = {
     const booking = req.locals;
     const { user } = req;
     scheduleVisitation({ ...booking, userId: user._id })
-      .then(({ newSchedule, vendorEmail }) => {
+      .then(({ schedule, vendorEmail }) => {
         const contentBottom = `
-          <strong> Name: </strong>: ${newSchedule.visitorName}<br /> 
-          <strong> Phone: </strong> ${newSchedule.visitorPhone}<br />
+          <strong> Name: </strong>: ${schedule.visitorName}<br /> 
+          <strong> Phone: </strong> ${schedule.visitorPhone}<br />
           <strong> Email: </strong> ${
-            newSchedule.visitorEmail ? newSchedule.visitorEmail : 'No email provided'
+            schedule.visitorEmail ? schedule.visitorEmail : 'No email provided'
           } <br />
         `;
         sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: vendorEmail }, { contentBottom });
         res
           .status(httpStatus.CREATED)
-          .json({ success: true, message: 'Visit scheduled successfully', schedule: newSchedule });
+          .json({ success: true, message: 'Visit scheduled successfully', schedule });
       })
       .catch((error) => next(error));
   },
+
   getAll(req, res, next) {
-    getAllVisitations()
+    const userId = req.user._id;
+    getAllVisitations(userId)
       .then((schedules) => {
-        res.status(httpStatus.OK).json({ success: true, schedules });
+        if (schedules.length > 0) {
+          res.status(httpStatus.OK).json({ success: true, schedules });
+        } else {
+          res
+            .status(httpStatus.NOT_FOUND)
+            .json({ success: false, message: 'No schedules available' });
+        }
       })
       .catch((error) => next(error));
   },
--- a/server/services/visitation.service.js
+++ b/server/services/visitation.service.js
@@ -20,7 +20,7 @@ export const scheduleVisitation = async (schedule) => {
 
     try {
       const newSchedule = await new Visitation({ ...schedule, vendorId: property.addedBy }).save();
-      return { schedule: newSchedule, vendorEmail: vendor.email };
+      return { schedule: newSchedule, vendor };
     } catch (error) {
       throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error scheduling visit', error);
     }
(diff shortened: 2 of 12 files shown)


server/services/visitation.service.js
```diff
@@ -2,15 +2,21 @@ import Visitation from '../models/visitation.model';
...
+
   if (property) {
+    const vendor = await getUserById(property.addedBy).catch((error) => {
+      throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+    });
+
     try {
       const newSchedule = await new Visitation(schedule).save();
-      return newSchedule;
+      return { newSchedule, vendorEmail: vendor.email };
```
andela-hpopoola, line 19: Nice changes here. Even though it doesn't really matter here because the number of changes you made is not too much, it might be better to leave this as schedule to avoid making those changes.

```suggestion
      return { schedule: newSchedule, vendorEmail: vendor.email };
```

server/controllers/visitation.controllers.js
```diff
@@ -3,25 +3,23 @@ import EMAIL_CONTENT from '../../mailer';
...
-          <strong> Phone: </strong> ${schedule.visitorPhone}<br />
+          <strong> Name: </strong>: ${newSchedule.visitorName}<br /> 
+          <strong> Phone: </strong> ${newSchedule.visitorPhone}<br />
           <strong> Email: </strong> ${
-            schedule.visitorEmail ? schedule.visitorEmail : 'No email provided'
+            newSchedule.visitorEmail ? newSchedule.visitorEmail : 'No email provided'
           } <br />
         `;
-        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: BALLERS_EMAIL }, { contentBottom });
+        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: vendorEmail }, { contentBottom });
```
andela-hpopoola, line 19: We need to update the test in the controller file to be sure from `admin` user to a vendor. 

https://github.com/Highrachy/ballers-api/blob/b5cc8e0b234de242ff5f90eaa0467758c9d84fec/test/controllers/visitation.controller.test.js#L32

Also for each of our mail tests, we want to ensure that we send the right mail and to the right email. So it is important to add a test like this line to the test too so that we can test the content being sent and the email it is sending it to.

https://github.com/Highrachy/ballers-api/blob/b5cc8e0b234de242ff5f90eaa0467758c9d84fec/test/controllers/user.controller.test.js#L69

andela-hpopoola (changes requested):
Thank you for working on this. This looks good but I am most concerned about updating the `visitation.controller.test.js` to ensure we test for vendor and also ensure we test the mail we are sending out. Thank you.

test/controllers/visitation.controller.test.js
```diff
@@ -269,50 +293,97 @@ describe('Visitation Controller', () => {
...
+          request()
+            .get('/api/v1/visitation/all')
+            .set('authorization', vendorToken)
+            .end((err, res) => {
+              expect(res).to.have.status(200);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body).to.have.property('schedules');
+              expect(res.body.schedules.length).to.be.eql(1);
+              expect(testVendorPropertyId.equals(res.body.schedules[0].propertyId)).to.be.eql(true);
+              expect(userId.equals(res.body.schedules[0].userId)).to.be.eql(true);
```
andela-hpopoola, line 374: To make this test more logical, this should test that the `vendorId` is equal to this VendorId Testing that it is true will not catch the bug if the test is returning the wrong Vendor property.

server/controllers/visitation.controllers.js
```diff
@@ -3,32 +3,38 @@ import EMAIL_CONTENT from '../../mailer';
...
+      .then(({ schedule, vendorEmail }) => {
         const contentBottom = `
           <strong> Name: </strong>: ${schedule.visitorName}<br /> 
           <strong> Phone: </strong> ${schedule.visitorPhone}<br />
           <strong> Email: </strong> ${
             schedule.visitorEmail ? schedule.visitorEmail : 'No email provided'
           } <br />
         `;
-        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: BALLERS_EMAIL }, { contentBottom });
+        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, { email: vendorEmail }, { contentBottom });
```
andela-hpopoola, line 19: Nice thinking here. However, we want the mail to also retrieve the company name, so we might need to add the vendor company name here so that it can work as expected. 

https://github.com/Highrachy/ballers-api/blob/76129cd26df03c27bc2518302c3965ecd7de6cbb/server/services/mailer.service.js#L64

I didn't test it though, you can start by passing the whole vendor Object where vendor stands for all the user information.

```suggestion
        sendMail(EMAIL_CONTENT.SCHEDULE_VISIT, vendor, { contentBottom });
```

server/controllers/visitation.controllers.js
```diff
@@ -3,32 +3,38 @@ import EMAIL_CONTENT from '../../mailer';
...
+    const userId = req.user._id;
+    getAllVisitations(userId)
       .then((schedules) => {
-        res.status(httpStatus.OK).json({ success: true, schedules });
+        if (schedules.length > 0) {
+          res.status(httpStatus.OK).json({ success: true, schedules });
+        } else {
+          res
+            .status(httpStatus.NOT_FOUND)
+            .json({ success: false, message: 'No schedules available' });
```
andela-hpopoola, line 36: Even though the schedules array is empty, we need to add it here so that it can be consistent with line 32 and so that we can also test that it is 0.

```suggestion
            .json({ success: false, message: 'No schedules available', schedules });
```

server/services/visitation.service.js
```diff
@@ -19,14 +29,38 @@ export const scheduleVisitation = async (schedule) => {
...
+export const getAllVisitations = async (userId) => {
+  let schedules;
+
+  const user = await getUserById(userId).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (user.role === USER_ROLE.VENDOR) {
+    schedules = await Visitation.aggregate([
+      { $match: { vendorId: ObjectId(userId) } },
```
andela-hpopoola, line 41: Since this is the only difference between the 2 queries, we can use 1 query instead of splitting them off, and add the $match if it is a vendor that is making the request.

server/controllers/visitation.controllers.js
```diff
@@ -3,32 +3,38 @@ import EMAIL_CONTENT from '../../mailer';
...
         res
           .status(httpStatus.CREATED)
           .json({ success: true, message: 'Visit scheduled successfully', schedule });
       })
       .catch((error) => next(error));
   },
+
   getAll(req, res, next) {
-    getAllVisitations()
+    const userId = req.user._id;
```
andela-hpopoola, line 28: We can send the whole user object instead of getting the `_id` only here.
```suggestion
    const { user } = req;
```

server/services/visitation.service.js
```diff
@@ -19,14 +29,38 @@ export const scheduleVisitation = async (schedule) => {
...
-    {
-      $lookup: {
-        from: 'properties',
-        localField: 'propertyId',
-        foreignField: '_id',
-        as: 'propertyInfo',
+export const getAllVisitations = async (userId) => {
+  let schedules;
+
+  const user = await getUserById(userId).catch((error) => {
```
andela-hpopoola, line 35: I dont think this is necessary, I think we should already have the user object defined in the `req` variable in the controller, we can pass that here instead of making an additional request.

test/controllers/visitation.controller.test.js
```diff
@@ -19,31 +19,48 @@ const sandbox = sinon.createSandbox();
...
+  _id: demoPropertyId,
+  addedBy: vendorId,
+  updatedBy: vendorId,
+});
 
 describe('Visitation Controller', () => {
   beforeEach(async () => {
     userToken = await addUser(user);
     adminToken = await addUser(admin);
+    vendorToken = await addUser(vendor);
```
andela-hpopoola, line 46: Can we add another vendor here again and test that each vendor is getting their own content. This is necessary because we want to ensure that each of the vendors get their own information only. Using one vendor for the test might not catch this, so it is better to add 2 vendors and test that the information returned is for the right vendor.

Feel free to use a loop or a function to make the test easier.

test/controllers/visitation.controller.test.js
```diff
@@ -53,9 +70,16 @@ describe('Visitation Controller', () => {
             expect(res.body.success).to.be.eql(true);
             expect(res.body.message).to.be.eql('Visit scheduled successfully');
             expect(res.body).to.have.property('schedule');
-            expect(propId.equals(res.body.schedule.propertyId)).to.be.eql(true);
+            expect(demoPropertyId.equals(res.body.schedule.propertyId)).to.be.eql(true);
```
andela-hpopoola, line 73: Can we update this to a more readable format, 🙏 

```suggestion
            expect(demoPropertyId.toString()).to.be.eql(res.body.schedule.propertyId.toString());
```

test/controllers/visitation.controller.test.js
```diff
@@ -53,9 +70,16 @@ describe('Visitation Controller', () => {
...
             expect(sendMailStub.callCount).to.eq(1);
-            expect(sendMailStub).to.have.be.calledWith(EMAIL_CONTENT.SCHEDULE_VISIT);
+            expect(sendMailStub).to.have.be.calledWith();
```
andela-hpopoola, line 75: Not sure why this line passed but this is not necessary.

test/controllers/visitation.controller.test.js
```diff
@@ -269,50 +293,97 @@ describe('Visitation Controller', () => {
...
+        await scheduleVisitation(bookingForAdmin);
+        await scheduleVisitation(bookingForVendor);
       });
 
-      context('with a valid token & id', () => {
-        beforeEach(async () => {
-          await scheduleVisitation(booking);
-        });
-        it('returns successful payload', (done) => {
+      context('when an admin token is used', () => {
```
andela-hpopoola, line 333: Lets add 3 scheduled visitations, 
1. for admin
2. for vendor1
3. for vendor2

Lets write tests to ensure that 3 is returned for admin, vendor1 and vendor2 returns just 1 scheduled visits and the id matches their property.

test/services/visitation.service.test.js
```diff
@@ -121,19 +129,32 @@ describe('Visitation Service', () => {
...
 
     beforeEach(async () => {
+      await addUser(user);
+      await addUser(vendor);
       await Property.create(property);
       await scheduleVisitation(validBooking);
       await scheduleVisitation(validBooking);
     });
 
     context('when schedule added is valid', () => {
```
andela-hpopoola, line 155: We can use a loop here to make the tests available for an admin and a vendor. This is optional though (to save time) as we already have tests for this in the controller.

andela-hpopoola (changes requested):
Thank you for making the changes. We are almost there, I left a couple of comments to ensure we get the implementation right. I am sure that once we get this right, other endpoints will be easier to implement for you. Thank you and Nice work.

test/controllers/visitation.controller.test.js
tobieeniafe: you suggested this format a while back tho

server/services/visitation.service.js
tobieeniafe: couldn't find a solution for this. suggestions?

server/services/visitation.service.js
```diff
@@ -19,14 +29,34 @@ export const scheduleVisitation = async (schedule) => {
...
 
-export const getAllVisitations = async () =>
-  Visitation.aggregate([
-    {
-      $lookup: {
-        from: 'properties',
-        localField: 'propertyId',
-        foreignField: '_id',
-        as: 'propertyInfo',
+export const getAllVisitations = async (user) => {
```
andela-hpopoola, line 32: I thought I left a comment on ensuring that we use just one options for the vendor and admin. This should work:

```js
export const getAllVisitations = async (user) => {
  const scheduleOptions = [
    {
      $lookup: {
        from: 'properties',
        localField: 'propertyId',
        foreignField: '_id',
        as: 'propertyInfo',
      },
    },
  ];

  if (user.role === USER_ROLE.VENDOR) {
    scheduleOptions.push({ $match: { vendorId: ObjectId(user._id) } });
  }

  const schedules = await Visitation.aggregate(scheduleOptions);

  return schedules;
};
```

test/controllers/visitation.controller.test.js
```diff
@@ -269,50 +289,134 @@ describe('Visitation Controller', () => {
...
+    const bookingForVendor = VisitationFactory.build({ propertyId: testVendorPropertyId, userId });
+    const bookingForVendor2 = VisitationFactory.build({ propertyId: vendor2PropertyId, userId });
 
     context('when no schedule exists', () => {
       it('returns not found', (done) => {
         request()
           .get('/api/v1/visitation/all')
           .set('authorization', adminToken)
           .end((err, res) => {
-            expect(res).to.have.status(200);
```
andela-hpopoola, line 280: Can we revert the tests here, please? I prefer we use 404 when an individual item is not found not when no visitation present in the database. If no visitation is not in the database, we can assume that the request is successful and doesn't require an error message. However, if we are trying to get a single visitation and it is not found, we should use 404 for that.

This is the pattern we adopted for this project. Changing it now would force us to change for other endpoints which will be time-consuming. Consistency matters a lot when developing APIs, so let's try to ensure we are consistent for all implementation. 

Lets revert this, please.

test/controllers/visitation.controller.test.js
```diff
@@ -19,31 +19,51 @@ const sandbox = sinon.createSandbox();
 
 let userToken;
 let adminToken;
+let vendorToken;
+let vendor2Token;
 
-const user = UserFactory.build({ role: USER_ROLE.USER, activated: true });
+const userId = mongoose.Types.ObjectId();
+const user = UserFactory.build({ _id: userId, role: USER_ROLE.USER, activated: true });
```
andela-hpopoola, line 26: Instead of defining the userId and user factory generation separately, you can define the userId within the user factory by using the option we defined in the factory:

https://github.com/Highrachy/ballers-api/blob/a502bdb3b690501378058ff9f9e49e0dcc6c9446/test/factories/user.factory.js#L7

Here is an example of where it is used
https://github.com/Highrachy/ballers-api/blob/a502bdb3b690501378058ff9f9e49e0dcc6c9446/test/controllers/property.controller.test.js#L1717

In places you need the Id, you could as well get them by using `vendor._id` or `user._id` depending on the name of the factory.

test/controllers/visitation.controller.test.js
```diff
@@ -269,50 +289,134 @@ describe('Visitation Controller', () => {
   });
 
   describe('Get all visitations', () => {
-    const booking = VisitationFactory.build({ propertyId: propId, userId: adminId });
+    const testAdminPropertyId = mongoose.Types.ObjectId();
+    const testAdminProperty = PropertyFactory.build({
```
andela-hpopoola, line 293: Can we remove the testAdminProperty? This is to prevent the test from failing in the future. When using the application, the admin should be unable to add properties but should only be able to view all the visitations.

test/controllers/visitation.controller.test.js
```diff
@@ -269,50 +289,134 @@ describe('Visitation Controller', () => {
...
+              );
+              expect(res.body.schedules[0].propertyInfo[0].description).to.be.eql(
+                testVendorProperty.description,
+              );
+              done();
+            });
+        });
+      });
+
+      context('when an vendor token is used', () => {
```
andela-hpopoola, line 400:
```suggestion
      context('when another vendor token is used', () => {
```

andela-hpopoola (changes requested):
Thank you for your hard work on this. I left more comments to ensure we get this totally right.

Also, we need to add pagination to the `all` endpoint. Anytime we are working with any `all` endpoint, it makes sense to add pagination to it.



Can we substitute the test on line 439 
```
      context('when admin token is not available', () => {
```
for the `itReturnsAnErrorForInvalidToken` helper test and ensure we test for an admin and a single vendor.

Also, Line 456 and Line 470 can be substituted too with the right helper test.

tobieeniafe:
@andela-hpopoola is there anything left to complete on this PR? so we can merge and proceed

andela-hpopoola (approved):
Apologies, I have forgotten that I need to approve this. I have fixed the test typo so feel free to merge anytime you want.

PR #89, opened 2020-12-09
#176029232 update offer letter for vendor

tobieeniafe:
#### What does this PR do?
Update offer letter endpoint to support vendor

Update offer letter endpoint to support vendor

`GET /api/v1/offer/all`

#176029232


andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to ensure we get this right. Also, we need to ensure that property addition is done by vendors only (admins should be unable to add properties). Also, offer letter generation should be a vendor only thing. 

The admin should only be able to view. This is to resolve conflict, for investigation (in case of dispute) and to know that people are using this features right.

test/services/property.service.test.js
```diff
@@ -111,26 +123,25 @@ describe('Property Service', () => {
...
   describe('#deleteProperty', () => {
-    const _id = mongoose.Types.ObjectId();
-    before(async () => {
-      await Property.create(PropertyFactory.build({ _id, addedBy: _id, updatedBy: _id }));
+    beforeEach(async () => {
+      await addProperty(property);
     });
 
     context('when property is deleted', () => {
       it('returns a valid updated user', async () => {
```
andela-hpopoola, line 131: Deleted property returns a valid updated user 🤣 🤣 🤣

andela-hpopoola (commented):
Well done and thank you for working on this. Reviewing the code only is giving me enough headache, I am not sure how you were able to wrap your head around it. 

I have reviewed the essential code, will need time to totally review tests and try the updated endpoints on my system. Thank you for your hardwork.

andela-hpopoola (changes requested):
Thank you for working on this. I have been able to review some of the tests. Truthfully, I couldn't review all (the comments will be too much for you to implement), so I added only the comments we can use to improve the offer letter flow for the vendor. We would revisit other controllers when we add the pagination service to them.

When you add more experience, you will learn how to break a PR into multiple PRs to make the code organisation better and reviews easier. This is overly complicated, and I can only imagine the pain you went through to make these changes. 

Thank you for your hard work.

server/controllers/user.controllers.js
```diff
@@ -110,7 +110,7 @@ const UserController = {
   assignProperty(req, res, next) {
     const toBeAssigned = req.locals;
     const { user } = req;
-    assignPropertyToUser({ ...toBeAssigned, assignedBy: user._id })
+    assignPropertyToUser({ ...toBeAssigned, assignedBy: user._id, vendor: user })
```
andela-hpopoola, line 113: I am curious why we need to add the vendor object here.

test/controllers/enquiry.controller.test.js
```diff
@@ -8,18 +8,24 @@ import PropertyFactory from '../factories/property.factory';
...
+const adminUser = UserFactory.build({ role: USER_ROLE.ADMIN, activated: true });
+const vendorUser = UserFactory.build(
+  { role: USER_ROLE.VENDOR, activated: true },
+  { generateId: true },
+);
+const regularUser = UserFactory.build({ _id, role: USER_ROLE.USER, activated: true });
 
 describe('Enquiry Controller', () => {
   beforeEach(async () => {
     adminToken = await addUser(adminUser);
```
andela-hpopoola, line 27: You might want to test for the vendorToken everywhere the adminToken was previously used. Maybe in a different PR.

andela-hpopoola (changes requested):
Thank you for working on this. It doesn't seem like we are making so much progress with this approach:
 - Some of the tests are poorly written (instances where we changed vendorToken to adminToken, we also need to add tests to ensure that the adminToken can no longer access the enpoints)
 - Reviewing this PR is very painful and lots of unexplained code. You have not resolved over 50% of the comments I left 4 days ago.
 - Most of the tests are incomplete and we would need to rewrite most of them again in a later PR. Most of them are just patches that would need to be resolved in a later PR. I don't think it is safe to merge them without resolving them.
 - Lastly, the progress made on this makes it seem like you haven't been working even though you have been doing more works than the previous PRs.

test/controllers/offer.controller.test.js
```diff
@@ -1444,299 +837,628 @@ describe('Offer Controller', () => {
...
-      offerId,
-      question: 'Are all rooms ensuite',
-    };
 
-    describe('when offers exist in db', () => {
-      beforeEach(async () => {
-        await addEnquiry(enquiry);
-        await createOffer(offer);
-      });
+    describe('Get all offers of a user', () => {
```
andela-hpopoola, line 841: We need to ensure that a vendor can get all offers of a user too. I believe this functionality is even more important for a vendor than an admin.

test/controllers/offer.controller.test.js
```diff
@@ -1444,299 +837,628 @@ describe('Offer Controller', () => {
...
+          });
+        });
+      });
+    });
+
+    describe('Get all active offers', () => {
+      const endpoint = '/api/v1/offer/active';
+      const method = 'get';
+
+      const enquiry1 = EnquiryFactory.build(
```
andela-hpopoola, line 946: A loop or buildList is more preferred here than incremental variables. Even if you need to create the variables here in a pattern, you can define the pattern in an array and loop through it. 

You might need to add the expires value in a different array assign them while looping.

test/controllers/offer.controller.test.js
```diff
@@ -1444,299 +837,628 @@ describe('Offer Controller', () => {
...
-                expect(res.body.error).to.be.eql('"Offer Id" is not allowed to be empty');
-                expect(sendMailStub.callCount).to.eq(0);
+                expect(res).to.have.status(200);
+                expect(res.body.success).to.be.eql(true);
+                expect(res.body).to.have.property('offer');
+                expect(res.body.offer.concern.length).to.be.eql(1);
+                expect(res.body.offer.concern[0].status).to.be.eql('Pending');
+                expect(res.body.offer.concern[0].question).to.be.eql(concern.question);
+                expect(sendMailStub.callCount).to.eq(1);
+                expect(sendMailStub).to.have.be.calledWith(EMAIL_CONTENT.RAISE_CONCERN);
```
andela-hpopoola, line 1175: For subsequent PRs, lets ensure that we test for the email and name when testing for  `sendMailStub`

test/controllers/offer.controller.test.js
```diff
@@ -1444,299 +837,628 @@ describe('Offer Controller', () => {
...
+                });
+            });
+          });
+
+          context('when concern id is empty', () => {
+            it('returns an error', (done) => {
+              request()
+                .put('/api/v1/offer/resolve-concern')
+                .set('authorization', vendorToken)
+                .send({ concernId: '', offerId: offer._id, question: 'Are all rooms ensuite' })
```
andela-hpopoola, line 1349: With a few tweaking, we can use the `itReturnsErrorForEmptyFields` here too.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to make the existing code better. The number of code changes are too much so it is getting even more difficult to review for me. 

Once you fix this PR, we might have to merge it in good faith - May God help us - so that we can proceed with perfecting other endpoints.

test/controllers/offer.controller.test.js
```diff
@@ -1416,24 +1316,52 @@ describe('Offer Controller', () => {
...
+      10,
+      { addedBy: vendorUser._id, updatedBy: vendorUser._id },
+      { generateId: true },
+    );
+
+    const user2Properties = PropertyFactory.buildList(
+      8,
+      { addedBy: vendorUser._id, updatedBy: vendorUser._id },
+      { generateId: true },
+    );
```
andela-hpopoola, line 1338: You should be able to aggregate userProperties and user2Properties into an array instead of doing it separately. This will ensure you use one variable for userEnquiries and userOffers.

const allProperties = [...userProperties, ...user2Properties];

test/controllers/offer.controller.test.js
```diff
@@ -1442,24 +1370,345 @@ describe('Offer Controller', () => {
...
+      ),
+    );
+
+    beforeEach(async () => {
+      userToken = await addUser(regularUser);
+      vendorToken = await addUser(vendorUser);
+      user2Token = await addUser(user2);
+    });
+
+    context('when no offer is found', () => {
```
andela-hpopoola, line 1392: Not sure why we are not using the pagination test function for this.

test/controllers/offer.controller.test.js
```diff
@@ -1416,24 +1316,52 @@ describe('Offer Controller', () => {
...
+    const user2 = UserFactory.build(
+      { role: USER_ROLE.USER, activated: true },
+      { generateId: true },
+    );
 
-    const demoVendorUser = UserFactory.build({ role: USER_ROLE.VENDOR, activated: true });
+    const editorUser = UserFactory.build({ role: USER_ROLE.EDITOR, activated: true });
+
+    const userProperties = PropertyFactory.buildList(
+      10,
```
andela-hpopoola, line 1329: Let's make this 12 or any other number apart from 10 so that we can easily differentiate the returned number from the default limit of 10.

test/controllers/offer.controller.test.js
```diff
@@ -1416,24 +1316,52 @@ describe('Offer Controller', () => {
...
+      { addedBy: vendorUser._id, updatedBy: vendorUser._id },
+      { generateId: true },
+    );
+
+    const user2Properties = PropertyFactory.buildList(
+      8,
+      { addedBy: vendorUser._id, updatedBy: vendorUser._id },
```
andela-hpopoola, line 1336: Let's make this another vendor's id here so that we would be able to verify that the vendor tests here return for the right vendor.

test/controllers/offer.controller.test.js
```diff
@@ -1442,24 +1370,345 @@ describe('Offer Controller', () => {
...
+      userToken = await addUser(regularUser);
+      vendorToken = await addUser(vendorUser);
+      user2Token = await addUser(user2);
+    });
+
+    context('when no offer is found', () => {
+      it('returns empty array of offers', (done) => {
+        request()
+          [method](endpoint)
+          .set('authorization', vendorToken)
```
andela-hpopoola, line 1396: For getAllOffers, the endpoint is catering for 3 roles -- Admin, Vendor and User. For the tests, we need to ensure that we cater for the three roles. That is, We must write the same tests for the three roles to be sure that this is working as expected.

test/controllers/offer.controller.test.js
```diff
@@ -1442,24 +1370,345 @@ describe('Offer Controller', () => {
...
+    );
+
+    beforeEach(async () => {
+      userToken = await addUser(regularUser);
+      vendorToken = await addUser(vendorUser);
+      user2Token = await addUser(user2);
+    });
+
+    context('when no offer is found', () => {
+      it('returns empty array of offers', (done) => {
```
andela-hpopoola, line 1393: We need similar tests for admin and user to be sure that they also return an empty array of offers.

test/controllers/offer.controller.test.js
```diff
@@ -1442,24 +1370,345 @@ describe('Offer Controller', () => {
...
 
       itReturnsForbiddenForNoToken({ endpoint, method });
 
-      itReturnsAnErrorWhenServiceFails({
-        endpoint,
-        method,
-        user: demoVendorUser,
-        model: Offer,
-        modelMethod: 'aggregate',
+      context('when both page and limit parameters are set', () => {
```
andela-hpopoola, line 1444: I think we should be fine if we test the returned total for the vendors (as you have done for the users below) since we have tested the page and limit parameters in the adminTest above.

test/controllers/offer.controller.test.js
```diff
@@ -1442,24 +1370,345 @@ describe('Offer Controller', () => {
...
+              expect(res.body.pagination.limit).to.be.eql(10);
+              expect(res.body.pagination.total).to.be.eql(18);
+              expect(res.body.pagination.offset).to.be.eql(0);
+              expect(res.body.result.length).to.be.eql(10);
+              done();
+            });
+        });
+      });
+
+      itReturnsTheRightPaginationValue({ endpoint, method, user: adminUser });
```
andela-hpopoola, line 1438: Lets put this in a context block so that the purpose of the test easily be understood in the future.

test/controllers/offer.controller.test.js
```diff
@@ -1442,24 +1370,345 @@ describe('Offer Controller', () => {
...
+    );
+
+    beforeEach(async () => {
+      adminToken = await addUser(adminUser);
+      userToken = await addUser(regularUser);
+      vendorToken = await addUser(vendorUser);
+      await addUser(user2);
+    });
+
+    context('when no offer is found', () => {
```
andela-hpopoola, line 1608: I think we have a test method for this.

test/controllers/offer.controller.test.js
```diff
@@ -1442,24 +1370,345 @@ describe('Offer Controller', () => {
...
+  describe('Get all offers of a user', () => {
+    const method = 'get';
+    const endpoint = `/api/v1/offer/user/${regularUser._id}`;
+
+    const user2 = UserFactory.build(
+      { role: USER_ROLE.USER, activated: true },
+      { generateId: true },
+    );
+
+    const userProperties = PropertyFactory.buildList(
```
andela-hpopoola, line 1544: I think it would have been much better to add 18 properties for this user and a random number of properties for user2. This will enable you to use the pagination test functions without rewritting the whole tests.

test/controllers/offer.controller.test.js
```diff
@@ -1442,24 +1370,345 @@ describe('Offer Controller', () => {
...
+      { addedBy: vendorUser._id, updatedBy: vendorUser._id },
+      { generateId: true },
+    );
+
+    const user2Properties = PropertyFactory.buildList(
+      8,
+      { addedBy: vendorUser._id, updatedBy: vendorUser._id },
+      { generateId: true },
+    );
+    const userEnquiries = userProperties.map((_, index) =>
```
andela-hpopoola, line 1555: Also, we can combine user1 and user2 properties to ensure we don't separate enquiries and offers.

test/factories/offer.factory.js
```diff
@@ -10,7 +10,7 @@ export default new Factory()
   .attr('handOverDate', '2090-09-08')
   .attr('deliveryState', 'new')
   .attr('totalAmountPayable', 50000000)
-  .sequence('allocationInPercentage', (i) => i)
+  .attr('allocationInPercentage', 40)
```
andela-hpopoola, line 13: Not sure why you updated this -- Can we revert this as the sequence, please. If you need an allocationInPercentage of 40, you can always pass it into the factory.

test/services/offer.service.test.js
```diff
@@ -191,102 +191,105 @@ describe('Offer Service', () => {
...
+        expect(vendorOffers.pagination.currentPage).to.be.eql(1);
+        expect(vendorOffers.pagination.total).to.be.eql(10);
+        expect(vendorOffers.result.length).to.be.eql(10);
+        expect(vendorOffers.result[1].userId).to.be.eql(user._id);
       });
     });
     context('when new offer is added', () => {
       beforeEach(async () => {
-        await createOffer(offer3);
+        await Offer.insertMany(user2Offers);
```
andela-hpopoola, line 278: I think the implementation of your test here is different from the description. The test is currently testing that adding a new offer increases the number of offers for the same user. I think it would be better to implement that first, before adding more tests to confirm if another user's offer letter affects an existing user offer letter.

Also, can we try to stick as close as possible to the existing service test as the pagination tests has been handled by the controller test. The current test you wrote is a replica of the controller test; A more productive one would be writting the test like this:

- Lets add just one offer for user, and test that user, vendor and admin sees just 1 offer
- Add another offer for the same user, and test that user, vendor and admin sees 2 offers
- Add another offer for another user and test that user sees 1, vendor and admin sees 2
- Add another offer for the same user but another vendor and test that user sees 2, vendor sees 1 and admin sees 2
- Add another offer for another user and another vendor and test that user and vendor sees 1 but admin sees 2.

This will enable us to catch more bugs if you can work on this.

andela-hpopoola (changes requested):

Most of the comments are on the pagination tests. We are spending too much time on this, please try to fix everything asap to enable us to move to other tasks.

It would have been preferable if you have removed the pagination implementation as discussed. Also, you failed to communicate that you are keeping it. Let's ensure we try to communicate more next time 🙏🏽

test/factories/offer.factory.js
andela-hpopoola:
```suggestion
  .sequence('allocationInPercentage', (i) => i <= 100 ? i : 99)
```

test/controllers/offer.controller.test.js
```diff
@@ -1444,297 +841,823 @@ describe('Offer Controller', () => {
...
+
+    beforeEach(async () => {
+      adminToken = await addUser(adminUser);
+      userToken = await addUser(regularUser);
+      vendorToken = await addUser(vendorUser);
+      user2Token = await addUser(user2);
+      vendor2Token = await addUser(vendor2);
+    });
+
+    itReturnsEmptyValuesWhenNoItemExistInDatabase({
```
andela-hpopoola, line 1403: It is advisable to always group related tests to make reading it easier. However, the three tests can be aggregated together as:

```js
    context('when no offers exists in db', () => {
      [adminUser, regularUser, vendorUser].map((user) =>
        itReturnsEmptyValuesWhenNoItemExistInDatabase({
          endpoint,
          method,
          user,
          useExistingUser: true,
        }),
      );
    });
```

test/controllers/offer.controller.test.js
```diff
@@ -1444,297 +841,823 @@ describe('Offer Controller', () => {
...
+      endpoint,
+      method,
+      user: vendorUser,
+      useExistingUser: true,
+    });
+
+    describe('when offers exist in db', () => {
```
andela-hpopoola, line 1424: Lets change this to context since you started the sentence with when

```suggestion
    context('when offers exist in db', () => {
```

test/controllers/offer.controller.test.js
```diff
@@ -1444,297 +841,823 @@ describe('Offer Controller', () => {
...
+              expect(res.body.result.length).to.be.eql(8);
               done();
             });
         });
       });
 
-      context('when getActiveOffers service fails', () => {
-        it('returns the error', (done) => {
-          sinon.stub(Offer, 'findOneAndUpdate').throws(new Error('Type Error'));
+      context('when request is sent by user 2', () => {
```
andela-hpopoola, line 1478:
```suggestion
      context('when request is sent by user2', () => {
```

test/controllers/offer.controller.test.js
```diff
@@ -1444,297 +841,823 @@ describe('Offer Controller', () => {
...
+              expect(res.body.result.length).to.be.eql(10);
               done();
             });
         });
       });
 
-      context('without token', () => {
-        it('returns error', (done) => {
+      context('when request is sent by vendor2 token', () => {
+        it('returns 8 offers', (done) => {
```
andela-hpopoola, line 1461:
```suggestion
        it('returns user2 offers', (done) => {
```

test/controllers/offer.controller.test.js
```diff
@@ -1444,297 +841,823 @@ describe('Offer Controller', () => {
...
             .end((err, res) => {
-              expect(res).to.have.status(403);
-              expect(res.body.success).to.be.eql(false);
-              expect(res.body.message).to.be.eql('Token needed to access resources');
-              expect(sendMailStub.callCount).to.eq(0);
+              expect(res).to.have.status(200);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body.pagination.currentPage).to.be.eql(1);
+              expect(res.body.pagination.limit).to.be.eql(10);
+              expect(res.body.pagination.total).to.be.eql(8);
```
andela-hpopoola, line 1470: Lets try to use the length of the offers for most of the tests instead of writing the number. Lets fix this for related tests too.
```suggestion
              expect(res.body.pagination.total).to.be.eql(user2Offers.length);
```

test/controllers/offer.controller.test.js
andela-hpopoola: Also, it is preferable not to write the number it would return in the description part of the describe or context block. It causes more confusion if we later update the value of the test. This is a description so the test will not fail but it can make debugging very difficult.

andela-hpopoola (approved):
Thank you for your hard work. I left a few comments but I believe this is good to go after fixing the comments. Feel free to merge after fixing. 

May God help us as we merge the PR.

PR #93, opened 2021-01-04
#176337684 add route to get single user

tobieeniafe:
#### What does this PR do?
Add route to get single user

Add route to get single user

`GET /api/v1/user/:id`

#176337684

### DIFF
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -18,7 +18,7 @@ import {
   updateVendor,
   addCommentToVerificationStep,
   verifyVendor,
-  getUserById,
+  getOneUser,
 } from '../services/user.service';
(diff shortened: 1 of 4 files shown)


andela-hpopoola (approved):
Thank you for working on this. I am approving this since the comments are nothing major. 

You should also try to always review your code on GitHub when you push before requesting an external review to ensure that you catch over 95% of typos. 

I believe this endpoint would later get more complex as we would want to load all the user-related stuff via this endpoint. For example, when retrieving a user role, the endpoint should load all the properties a user has been assigned to, his scheduled visitation e.tc. When retrieving information for a vendor, the endpoint should load all their added properties, payments made, the payment received e.t.c. Don't fret about this yet, we will overcome that blocker when we get there 💪🏽 .

PR #94, opened 2021-01-05
#176337881 edit & remove directors or signatories

tobieeniafe:
#### What does this PR do?
Add routes to edit & remove directors or signatories

Add routes to edit & remove directors or signatories

`PUT /api/v1/user/vendor/director/add`
`DELETE /api/v1/user/vendor/director/remove/:id`

#176337881

### DIFF
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -269,8 +269,8 @@ const UserController = {
     const directorId = req.params.id;
     const vendor = req.user;
     removeDirector({ directorId, vendor })
-      .then(() => {
-        res.status(httpStatus.OK).json({ success: true, message: 'Director removed' });
+      .then((user) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Director removed', user });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -258,12 +258,14 @@ const UserController = {
 
   editDirector(req, res, next) {
     const directorInfo = req.locals;
-    const userId = req.user._id;
-    editDirector({ directorInfo, userId })
-      .then((user) => {
-        res
-          .status(httpStatus.OK)
-          .json({ success: true, message: 'Director information was successfully updated', user });
+    const { user } = req;
+    editDirector({ directorInfo, user })
+      .then((updatedUser) => {
+        res.status(httpStatus.OK).json({
+          success: true,
+          message: 'Director information was successfully updated',
+          user: updatedUser,
+        });
       })
       .catch((error) => next(error));
   },
@@ -272,8 +274,10 @@ const UserController = {
     const directorId = req.params.id;
     const { user } = req;
     removeDirector({ directorId, user })
-      .then(() => {
-        res.status(httpStatus.OK).json({ success: true, message: 'Director removed' });
+      .then((updatedUser) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Director removed', user: updatedUser });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -18,7 +18,7 @@ import {
   updateVendor,
   addCommentToVerificationStep,
   verifyVendor,
-  addDirector,
+  editDirector,
   removeDirector,
   getOneUser,
 } from '../services/user.service';
@@ -256,22 +256,24 @@ const UserController = {
       .catch((error) => next(error));
   },
 
-  addDirector(req, res, next) {
+  editDirector(req, res, next) {
     const directorInfo = req.locals;
     const userId = req.user._id;
-    addDirector({ directorInfo, userId })
+    editDirector({ directorInfo, userId })
       .then((user) => {
-        res.status(httpStatus.OK).json({ success: true, message: 'Director added', user });
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Director information was successfully updated', user });
       })
       .catch((error) => next(error));
   },
 
   removeDirector(req, res, next) {
     const directorId = req.params.id;
-    const vendor = req.user;
-    removeDirector({ directorId, vendor })
-      .then((user) => {
-        res.status(httpStatus.OK).json({ success: true, message: 'Director removed', user });
+    const { user } = req;
+    removeDirector({ directorId, user })
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Director removed' });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -18,6 +18,8 @@ import {
   updateVendor,
   addCommentToVerificationStep,
   verifyVendor,
+  addDirector,
+  removeDirector,
 } from '../services/user.service';
 import { sendMail } from '../services/mailer.service';
 import EMAIL_CONTENT from '../../mailer';
@@ -249,6 +251,26 @@ const UserController = {
         res
           .status(httpStatus.OK)
           .json({ success: true, message: 'Vendor information updated', user });
+      })
+      .catch((error) => next(error));
+  },
+
+  addDirector(req, res, next) {
+    const directorInfo = req.locals;
+    const userId = req.user._id;
+    addDirector({ directorInfo, userId })
+      .then((user) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Director added', user });
+      })
+      .catch((error) => next(error));
+  },
+
+  removeDirector(req, res, next) {
+    const directorId = req.params.id;
+    const vendor = req.user;
+    removeDirector({ directorId, vendor })
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Director removed' });
       })
       .catch((error) => next(error));
   },
--- a/server/schemas/user.schema.js
+++ b/server/schemas/user.schema.js
@@ -145,5 +145,5 @@ export const updateVendorSchema = Joi.object({
 
 export const updateDirectorSchema = Joi.object({
   ...directorSchema,
-  directorId: Joi.objectId().label('Director id').required(),
+  _id: Joi.objectId().label('Director id').required(),
 });
--- a/server/schemas/user.schema.js
+++ b/server/schemas/user.schema.js
@@ -11,16 +11,15 @@ const phone = Joi.string().allow(null, '').optional().default('');
 const requiredString = (label) => Joi.string().label(label).required();
 const optionalString = (label) => Joi.string().label(label).optional();
 const optionalBoolean = (label) => Joi.boolean().label(label).optional();
-const requiredBoolean = (label) => Joi.boolean().label(label).required();
 
-const directors = Joi.array().items(
-  Joi.object().keys({
-    name: optionalString('Name'),
-    isSignatory: optionalBoolean('Signatory'),
-    signature: optionalString('Signature'),
-    phone: optionalString('Phone'),
-  }),
-);
+const directorSchema = {
+  name: optionalString('Name'),
+  phone: optionalString('Phone'),
+  isSignatory: optionalBoolean('Signatory'),
+  signature: optionalString('Signature'),
+};
+
+const directors = Joi.array().items(Joi.object().keys(directorSchema));
 
 const socialMedia = Joi.array().items(
   Joi.object().keys({
@@ -144,9 +143,7 @@ export const updateVendorSchema = Joi.object({
   vendor,
 });
 
-export const addDirectorSchema = Joi.object({
-  name: requiredString('Name'),
-  isSignatory: requiredBoolean('Signatory'),
-  signature: optionalString('Signature'),
-  phone: requiredString('Phone'),
+export const updateDirectorSchema = Joi.object({
+  ...directorSchema,
+  directorId: Joi.objectId().label('Director id').required(),
 });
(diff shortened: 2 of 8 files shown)


server/services/user.service.js
```diff
@@ -525,3 +525,44 @@ export const updateVendor = async ({ updatedVendor, vendorId }) => {
...
+    );
+  } catch (error) {
+    throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error adding director', error);
+  }
+};
+
+export const removeDirector = async ({ directorId, vendor }) => {
+  let { directors } = vendor.vendor;
+
+  directors = directors.filter((director) => director._id.toString() !== directorId.toString());
```
andela-hpopoola, line 548: I believe the two lines can be merged together into 1
```suggestion
  const signatories = directors.filter((director) => director._id.toString() !== directorId.toString() && director.isSignatory);
```

You can then proceed to check if there are signatories for verified vendors

```js
if (signatories.length === 0 && user.vendor.verified) {
...
```

server/services/user.service.js
```diff
@@ -525,3 +525,44 @@ export const updateVendor = async ({ updatedVendor, vendorId }) => {
...
+
+  const signatoryExists = directors.some((director) => director.isSignatory === true);
```
andela-hpopoola, line 550: This is no longer necessary based on the suggestion above

server/services/user.service.js
```diff
@@ -525,3 +525,44 @@ export const updateVendor = async ({ updatedVendor, vendorId }) => {
...
+
+  if (!signatoryExists) {
+    throw new ErrorHandler(
+      httpStatus.PRECONDITION_FAILED,
+      'Last account signatory cannot be deleted',
```
andela-hpopoola, line 555:
```suggestion
      'Signatory cannot be deleted. Verified vendors must have at least one signatory.',
```

server/controllers/user.controllers.js
```diff
@@ -252,6 +254,26 @@ const UserController = {
...
+    const userId = req.user._id;
+    addDirector({ directorInfo, userId })
+      .then((user) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Director added', user });
+      })
+      .catch((error) => next(error));
+  },
+
+  removeDirector(req, res, next) {
+    const directorId = req.params.id;
```
andela-hpopoola, line 269: `directorId` is currently missing in the director schema. Ensure you add this to make this to work as expected.

test/controllers/user.controller.test.js
```diff
@@ -1857,6 +1857,201 @@ describe('User Controller', () => {
...
+              },
+              {
+                _id: signatoryId,
+                name: 'Samuel',
+                isSignatory: true,
+                signature: 'signature.png',
+                phone: '08012345678',
+              },
+              {
+                name: 'John',
```
andela-hpopoola, line 1886: Not sure why you omitted an `_id` for this. You can automatically generate an id for this with
 
```js
_id: mongoose.Types.ObjectId()
```

test/controllers/user.controller.test.js
```diff
@@ -1857,6 +1857,201 @@ describe('User Controller', () => {
...
+                isSignatory: false,
+                phone: '08012345678',
+              },
+            ],
+          },
+        },
+        { generateId: true },
+      );
+      const invalidUser = UserFactory.build({ role: USER_ROLE.ADMIN }, { generateId: true });
```
andela-hpopoola, line 1895: It is preferable to use a USER role for an invalid user here. An Admin role for be used for subsequent tests

```suggestion
        const invalidUser = UserFactory.build({ role: USER_ROLE.USER }, { generateId: true });
```

test/controllers/user.controller.test.js
```diff
@@ -1857,6 +1857,201 @@ describe('User Controller', () => {
...
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body.message).to.be.eql('Director removed');
+              expect(res.body.user._id).to.be.eql(vendorUser._id.toString());
+              expect(res.body.user.vendor.directors.length).to.be.eql(2);
+              done();
+            });
+        });
+      });
+
+      context('when id passed is last account signatory', () => {
```
andela-hpopoola, line 1919: Lets ensure that we test this for verified and unverified users.

test/controllers/user.controller.test.js
```diff
@@ -1857,6 +1857,201 @@ describe('User Controller', () => {
...
+              expect(res).to.have.status(400);
+              expect(res.body.success).to.be.eql(false);
+              done();
+              User.findByIdAndUpdate.restore();
+            });
+        });
+      });
+    });
+
+    describe('Add director or signatories', () => {
```
andela-hpopoola, line 1967: Can we ensure that this test for updates not addition.

server/controllers/user.controllers.js
```diff
@@ -252,6 +254,26 @@ const UserController = {
       })
       .catch((error) => next(error));
   },
+
+  addDirector(req, res, next) {
+    const directorInfo = req.locals;
+    const userId = req.user._id;
+    addDirector({ directorInfo, userId })
```
andela-hpopoola, line 261: The director id is needed here to be sure of the director the user wish to update.

andela-hpopoola (changes requested):
Thank you for working on this. I think the most important part is that you need to add an id to the director's schema to ensure that we are updating/deleting the right directors.

server/controllers/user.controllers.js
```diff
@@ -254,6 +256,26 @@ const UserController = {
...
+      .then((user) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Director added', user });
+      })
+      .catch((error) => next(error));
+  },
+
+  removeDirector(req, res, next) {
+    const directorId = req.params.id;
+    const vendor = req.user;
+    removeDirector({ directorId, vendor })
```
andela-hpopoola, line 272: Might be better to use `user` here to prevent using something like vendor.vendor.

```js
const { user } = req;

removeDirector({ directorId, vendor })
```

server/services/user.service.js
andela-hpopoola: Feel free to ignore the above comment since you are replacing the whole director array (though I prefer using mongodb to remove it). However, you might want to make a few updates to prevent mutating the director array and making the code to make it more understandable:

```js
const { directors } = vendor.vendor;

  const updatedDirectors = directors.filter((director) => director._id.toString() !== directorId.toString());
...
```

andela-hpopoola (commented):
Just added a few more comments to clear doubts on some of my previous comments.

server/services/user.service.js
tobieeniafe: care to clarify on the ```js if (signatories.length === 0 && user.vendor.verified) ``` part and why adding `&& director.isSignatory`?

andela-hpopoola: Explained the reason here and told you to feel free to ignore it here
https://github.com/Highrachy/ballers-api/pull/94/files#r552943592

server/controllers/user.controllers.js
```diff
@@ -256,22 +256,24 @@ const UserController = {
...
   removeDirector(req, res, next) {
     const directorId = req.params.id;
-    const vendor = req.user;
-    removeDirector({ directorId, vendor })
-      .then((user) => {
-        res.status(httpStatus.OK).json({ success: true, message: 'Director removed', user });
+    const { user } = req;
+    removeDirector({ directorId, user })
+      .then(() => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Director removed' });
```
andela-hpopoola, line 276: Can you please return the updated user to make the returned object similar to edit directors. 

```suggestion
        res.status(httpStatus.OK).json({ success: true, message: 'Director removed', user: updatedUser });
```

server/schemas/user.schema.js
```diff
@@ -144,9 +143,7 @@ export const updateVendorSchema = Joi.object({
...
 });
 
-export const addDirectorSchema = Joi.object({
-  name: requiredString('Name'),
-  isSignatory: requiredBoolean('Signatory'),
-  signature: optionalString('Signature'),
-  phone: requiredString('Phone'),
+export const updateDirectorSchema = Joi.object({
+  ...directorSchema,
+  directorId: Joi.objectId().label('Director id').required(),
```
andela-hpopoola, line 148: I think it is preferable to use the exact name here for the id to make it easier to send from any part.
```suggestion
  _id: Joi.objectId().label('Director id').required(),
```

server/services/user.service.js
```diff
@@ -541,39 +541,55 @@ export const getOneUser = async (userId) =>
     },
   ]);
 
-export const addDirector = async ({ directorInfo, userId }) => {
+export const editDirector = async ({ directorInfo, userId }) => {
   const user = await getUserById(userId).catch((error) => {
```
andela-hpopoola, line 545: I believe we can get the user from the controller so this line is not necessary.

server/controllers/user.controllers.js
```diff
@@ -256,22 +256,24 @@ const UserController = {
       .catch((error) => next(error));
   },
 
-  addDirector(req, res, next) {
+  editDirector(req, res, next) {
     const directorInfo = req.locals;
     const userId = req.user._id;
-    addDirector({ directorInfo, userId })
+    editDirector({ directorInfo, userId })
```
andela-hpopoola, line 262: I believe it is preferable to send the user object into the edit Director to prevent retrieving the user details again. You might need to update other user variable to `updatedUser`

```suggestion
    const { user } = req;
    editDirector({ directorInfo, user })
```

server/services/user.service.js
```diff
@@ -541,39 +541,55 @@ export const getOneUser = async (userId) =>
...
     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
   });
 
+  const vendorIsForDirector = (director) =>
```
andela-hpopoola, line 549: The variable name is so confusing. Can we merge the function with the validVendor some statement below to make it more understandable

server/services/user.service.js
```diff
@@ -541,39 +541,55 @@ export const getOneUser = async (userId) =>
...
+    director._id.toString() === directorInfo.directorId.toString();
+
+  const validVendor = user.vendor.directors.some(vendorIsForDirector);
```
andela-hpopoola, line 552: `validVendor` is a confusing variable name since we are checking if the director is valid. You can choose a better name or use this:

```suggestion
  const directorIsValid = user.vendor.directors.some(vendorIsForDirector);
```

test/controllers/user.controller.test.js
```diff
@@ -1909,8 +1911,6 @@ describe('User Controller', () => {
               expect(res).to.have.status(200);
               expect(res.body.success).to.be.eql(true);
               expect(res.body.message).to.be.eql('Director removed');
```
andela-hpopoola, line 1913: We need to test that the returned user doesn't have the deleted director to be 100% sure the director has been removed.

test/controllers/user.controller.test.js
```diff
@@ -1868,6 +1868,7 @@ describe('User Controller', () => {
           activated: true,
           vendor: {
             companyName: 'Highrachy Investment Limited',
+            verified: true,
```
andela-hpopoola, line 1871: We need to add test to be sure that the last director can be deleted when verified is false.

test/controllers/user.controller.test.js
```diff
@@ -1974,6 +1977,7 @@ describe('User Controller', () => {
             companyName: 'Highrachy Investment Limited',
             directors: [
```
andela-hpopoola, line 1978: I believe it is preferable to use at least 3 director information to be sure that the edit director is editing the right director. As long as you are editing the first the first one, I believe the test should still be valid.

test/controllers/user.controller.test.js
```diff
@@ -2007,9 +2012,12 @@ describe('User Controller', () => {
             .end((err, res) => {
               expect(res).to.have.status(200);
               expect(res.body.success).to.be.eql(true);
-              expect(res.body.message).to.be.eql('Director added');
+              expect(res.body.message).to.be.eql('Director information was successfully updated');
               expect(res.body.user._id).to.be.eql(vendorUser._id.toString());
-              expect(res.body.user.vendor.directors.length).to.be.eql(2);
+              expect(res.body.user.vendor.directors[0]._id).to.be.eql(signatoryId.toString());
```
andela-hpopoola, line 2017: We need to test that the length of the returned director is the same (to be sure that the code didn't add it as a new director). 

Also, we need to test that the other 2 directors don't have the updated information after testing that the right director information is updated.

test/controllers/user.controller.test.js
```diff
@@ -1862,6 +1858,209 @@ describe('User Controller', () => {
...
+            });
+        });
+      });
+    });
+
+    describe('Edit director or signatories', () => {
+      let vendorToken;
+      const signatoryId = mongoose.Types.ObjectId();
+      const vendorUser = UserFactory.build(
+        {
```
andela-hpopoola, line 1973: This is the second time of using this in this PR. Might make sense to have a vendor factory to simplify vendor creation in the future.  It is not compulsory that factories should only match the models we have 🤷🏽‍♂️ .

andela-hpopoola (changes requested):
Thank you for working on this. I left a few more comments to make the code quality better.

test/controllers/user.controller.test.js
```diff
@@ -1862,6 +1858,252 @@ describe('User Controller', () => {
...
+              expect(res.body.user.vendor.directors[1]).to.eql({
+                ...vendorUser.vendor.directors[1],
+                _id: vendorUser.vendor.directors[1]._id.toString(),
+              });
+              expect(res.body.user.vendor.directors[2]).to.eql({
+                ...vendorUser.vendor.directors[2],
+                _id: vendorUser.vendor.directors[2]._id.toString(),
+              });
+              expect(res.body.user.vendor.directors[0]._id).to.be.eql(signatoryId.toString());
+              expect(res.body.user.vendor.directors[0].name).to.be.eql(data.name);
```
andela-hpopoola, line 2061: The 4 assertions for the directors[0] can be combined into a single test assertion as done for directors[1] and directors[2]

andela-hpopoola (approved):
Thank you for working on this. I have approved this PR, so you should be able to merge once you have fixed the pending comments. Thank you.

PR #96, opened 2021-01-08
#176409047 update middleware to support verified & unverified vendors

tobieeniafe:
#### What does this PR do?
Update middleware to support verified & unverified vendors

Update middleware to support verified & unverified vendors

#176409047

### DIFF
--- a/server/helpers/middleware.js
+++ b/server/helpers/middleware.js
@@ -102,7 +102,11 @@ export const isEditor = async (req, res, next) => {
 
 export const isVendorOrAdmin = async (req, res, next) => {
   const { user } = req;
-  if (user && (user.role === USER_ROLE.VENDOR || user.role === USER_ROLE.ADMIN)) {
+  if (
+    user &&
+    ((user.role === USER_ROLE.VENDOR && user.vendor.verified === true) ||
+      (user && user.role === USER_ROLE.ADMIN))
+  ) {
     next();
   } else {
     return res.status(httpStatus.FORBIDDEN).json({
--- a/server/helpers/middleware.js
+++ b/server/helpers/middleware.js
@@ -63,6 +63,19 @@ export const isAdmin = async (req, res, next) => {
 
 export const isVendor = async (req, res, next) => {
   const { user } = req;
+  if (user && user.role === USER_ROLE.VENDOR && user.vendor.verified === true) {
+    next();
+  } else {
+    return res.status(httpStatus.FORBIDDEN).json({
+      success: false,
+      message: 'You are not permitted to perform this action',
+    });
+  }
+  return null;
+};
+
+export const isUnverifiedVendor = async (req, res, next) => {
+  const { user } = req;
   if (user && user.role === USER_ROLE.VENDOR) {
     next();
   } else {
--- a/server/helpers/middleware.js
+++ b/server/helpers/middleware.js
@@ -48,9 +48,18 @@ export const authenticate = async (req, res, next) => {
   return null;
 };
 
+const isVerifiedVendor = (user) =>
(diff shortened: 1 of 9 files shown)


andela-hpopoola (changes requested):
Thank you for working on this. This looks good to go as you have made the implementation exactly as we discussed 💯 .

However, we need to add tests to confirm that unverified Vendors cannot do what a verified vendor can do. Yes, I know they should not be able to do that now but we need to add the test to always confirm this when we update the codebase.

The trick is to look at the places where isVendor or IsVendorOrAdmin or IsVendorOrAdminOrUser is used and add at least one test to confirm that an unverified vendor cannot access this endpoint.

Even though this might look unnecessary since nothing is breaking, however, this will increase our confidence that we have handled all edge cases when a vendor has been suspended or converted to unverified. 

Here is a sample test you can use to fast track the implementation:

https://github.com/Highrachy/ballers-api/blob/b998b5adc965cdfe433a05c88a478a9224ab978b/test/controllers/property.controller.test.js#L540

Thank you.

server/helpers/middleware.js
```diff
@@ -89,7 +102,11 @@ export const isEditor = async (req, res, next) => {
 
 export const isVendorOrAdmin = async (req, res, next) => {
   const { user } = req;
-  if (user && (user.role === USER_ROLE.VENDOR || user.role === USER_ROLE.ADMIN)) {
+  if (
```
andela-hpopoola, line 105: Lets create a function for verified vendors to make it easier to manage:

```js
const isVerifiedVendor = (user) =>
  user && user.role === USER_ROLE.VENDOR && user.vendor.verified === true;

const isValidAdmin = (user) =>  user && user.role === USER_ROLE.ADMIN;
```

This should replace every code where the same sections are used. This will make the code even more readable as this will become: 

```js
  if (isVerifiedVendor(user) || isValidAdmin(user)) {
```

server/helpers/middleware.js
```diff
@@ -89,7 +102,11 @@ export const isEditor = async (req, res, next) => {
 
 export const isVendorOrAdmin = async (req, res, next) => {
```
andela-hpopoola, line 103: Nice work here. Let's make the same changes for `isAdminOrUserOrVendor` too.

test/helpers.js
```diff
@@ -320,3 +320,40 @@ export const itReturnsErrorForEmptyFields = ({
...
+  useExistingUser = false,
+}) => {
+  let token;
+
+  beforeEach(async () => {
+    if (useExistingUser) {
+      const loggedInUser = await loginUser(user);
+      token = loggedInUser.token;
+    } else {
+      token = await addUser({ ...user, vendor: { verified: false } });
```
andela-hpopoola, line 338: Nice work here. However, it is prefereble to update only the verified key and not the whole vendor object.

```suggestion
      token = await addUser({ ...user, vendor: { ...user.vendor, verified: false } });
```

test/helpers.js
```diff
@@ -320,3 +320,40 @@ export const itReturnsErrorForEmptyFields = ({
...
+    }
+  });
+
+  context('when vendor is unverified', () => {
+    beforeEach(async () => {
+      await User.findByIdAndUpdate(user._id, { 'vendor.verified': false });
```
andela-hpopoola, line 344: This is only necessary when you are using existing users. This can be updated to 


```js
    beforeEach(async () => {
      if (useExistingUser) {
        await User.findByIdAndUpdate(user._id, { 'vendor.verified': false });
      }
    });
```

andela-hpopoola (approved):
Nice work. I left a few more comments. Kindly fix them before merging 🙏🏽 .

PR #97, opened 2021-01-10
#176408984 make enquiries support vendor

tobieeniafe:
#### What does this PR do?
Make enquiries support vendor

Make enquiries support vendor

#176408984

### DIFF
--- a/server/models/enquiry.model.js
+++ b/server/models/enquiry.model.js
@@ -86,6 +86,10 @@ const EnquirySchema = new mongoose.Schema(
       type: ObjectId,
       required: true,
     },
+    vendorId: {
+      type: ObjectId,
+      required: true,
+    },
     firstName: {
       type: String,
       required: true,
--- a/server/services/enquiry.service.js
+++ b/server/services/enquiry.service.js
@@ -4,6 +4,8 @@ import { ErrorHandler } from '../helpers/errorHandler';
 import httpStatus from '../helpers/httpStatus';
 // eslint-disable-next-line import/no-cycle
 import { getPropertyById } from './property.service';
+import { USER_ROLE } from '../helpers/constants';
+import { generatePagination, generateFacetData, getPaginationTotal } from '../helpers/pagination';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
@@ -27,22 +29,26 @@ export const addEnquiry = async (enquiry) => {
     );
   }
   try {
-    const newEnquiry = await new Enquiry(enquiry).save();
+    const newEnquiry = await new Enquiry({ ...enquiry, vendorId: property.addedBy }).save();
     return newEnquiry;
   } catch (error) {
     throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error adding enquiry', error);
   }
 };
 
-export const approveEnquiry = async (approvedEnquiry) => {
-  const enquiry = await getEnquiryById(approvedEnquiry.enquiryId).catch((error) => {
+export const approveEnquiry = async ({ enquiryId, vendor }) => {
+  const enquiry = await getEnquiryById(enquiryId).catch((error) => {
     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
   });
 
   try {
+    if (vendor.role === USER_ROLE.VENDOR && vendor._id.toString() !== enquiry.vendorId.toString()) {
+      throw new ErrorHandler(httpStatus.FORBIDDEN, 'You are not permitted to perform this action');
+    }
+
     return Enquiry.findOneAndUpdate(
       { _id: enquiry.id },
-      { $set: { approved: true, approvedBy: approvedEnquiry.adminId, approvalDate: Date.now() } },
+      { $set: { approved: true, approvedBy: vendor._id, approvalDate: Date.now() } },
       { new: true },
     );
   } catch (error) {
@@ -50,8 +56,8 @@ export const approveEnquiry = async (approvedEnquiry) => {
   }
 };
 
-export const getAllEnquiries = async () =>
-  Enquiry.aggregate([
+export const getAllEnquiries = async (user, page = 1, limit = 10) => {
+  const enquiryOptions = [
     {
       $lookup: {
         from: 'properties',
@@ -61,25 +67,34 @@ export const getAllEnquiries = async () =>
       },
     },
     {
-      $lookup: {
-        from: 'users',
-        localField: 'approvedBy',
-        foreignField: '_id',
-        as: 'approvedBy',
+      $facet: {
+        metadata: [{ $count: 'total' }, { $addFields: { page, limit } }],
+        data: generateFacetData(page, limit),
       },
     },
     {
       $project: {
         'propertyInfo.assignedTo': 0,
-        'approvedBy.assignedProperties': 0,
-        'approvedBy.password': 0,
-        'approvedBy.referralCode': 0,
       },
     },
-  ]);
+  ];
+  if (user.role === USER_ROLE.VENDOR) {
+    enquiryOptions.unshift({ $match: { vendorId: ObjectId(user._id) } });
+  }
+  if (user.role === USER_ROLE.USER) {
+    enquiryOptions.unshift({ $match: { userId: ObjectId(user._id) } });
+  }
 
-export const getEnquiry = async (enquiryId) =>
-  Enquiry.aggregate([
+  const enquiries = await Enquiry.aggregate(enquiryOptions);
+
+  const total = getPaginationTotal(enquiries);
+  const pagination = generatePagination(page, limit, total);
+  const result = enquiries[0].data;
+  return { pagination, result };
+};
+
+export const getEnquiry = async ({ enquiryId, user }) => {
+  const enquiryOptions = [
     { $match: { _id: ObjectId(enquiryId) } },
     {
       $lookup: {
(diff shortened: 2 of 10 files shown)


test/controllers/enquiry.controller.test.js
```diff
@@ -521,7 +531,7 @@ describe('Enquiry Controller', () => {
       it('returns a approved enquiry', (done) => {
```
andela-hpopoola, line 531:
```suggestion
      it('returns an approved enquiry', (done) => {
```

test/controllers/enquiry.controller.test.js
```diff
@@ -659,104 +669,138 @@ describe('Enquiry Controller', () => {
...
+          useExistingUser: true,
+        }),
+      );
     });
 
     describe('when enquiries exist in db', () => {
       beforeEach(async () => {
+        await Property.insertMany(dummyProperties);
+        await Enquiry.insertMany(userEnquiries);
         await addEnquiry(enquiry);
```
andela-hpopoola, line 742: Not sure why you left this. I believe you can remove that line and use the custom pagination test 🤷🏽‍♂️

test/controllers/enquiry.controller.test.js
```diff
@@ -659,104 +669,138 @@ describe('Enquiry Controller', () => {
...
+      context('when request is sent by admin token', () => {
+        it('returns all enquiries', (done) => {
           request()
-            .get('/api/v1/enquiry/all')
+            [method](endpoint)
             .set('authorization', adminToken)
             .end((err, res) => {
               expect(res).to.have.status(200);
               expect(res.body.success).to.be.eql(true);
-              expect(res.body).to.have.property('enquiries');
```
andela-hpopoola, line 699: Can you reintroduce this tests back as the new test doesn't replace it. In testing an endpoint with pagination, lets try to keep the following in mind. In this case, we have admin, user and vendor. 

Based on the information on the comment section, here is the breakdown
1. Test when nothing is in the database for the three  (already done on line 727)
2. Test that the pagination returns the right value for the three 
   - It is preferable to use the custom pagination test as it contains multiple test
   - A recurring mistake I notice is that you usually try to force the test in a single test assertion (i.e a describe/context block), it is not compulsory to do this. You might split the test into different blocks to make it more readable and better organized.
   - user test for this seems missing
3. Test that the right value is returned. 
    - The pagination test only deals with the numbers returned, not the validity of the data returned, so the pagination test cannot replace this
    - You will need to reintroduce this test you removed.
4. Add more tests based on the scenarios you are testing. Some of this can be incorporated in the pagination test setup:
   - You added some here like testing that vendor2 cannot access vendor1 enquiries
   - You can add more like ensuring that a user can only access his enquiry
   - Admin should be able to view all enquiries (This can be incorporated in the pagination test)

server/services/enquiry.service.js
```diff
@@ -89,20 +104,21 @@ export const getEnquiry = async (enquiryId) =>
...
         'propertyInfo.assignedTo': 0,
-        'approvedBy.assignedProperties': 0,
-        'approvedBy.password': 0,
-        'approvedBy.referralCode': 0,
       },
     },
-  ]);
+  ];
+
+  if (user.role === USER_ROLE.VENDOR) {
```
andela-hpopoola, line 114: Tests should be written to validate that this is works as expected.

server/services/enquiry.service.js
```diff
@@ -89,20 +104,21 @@ export const getEnquiry = async (enquiryId) =>
...
+    enquiryOptions.unshift({ $match: { vendorId: ObjectId(user._id) } });
+  }
+  if (user.role === USER_ROLE.USER) {
```
andela-hpopoola, line 117: Tests should be written to validate that this works too.

server/models/enquiry.model.js
```diff
@@ -86,6 +86,10 @@ const EnquirySchema = new mongoose.Schema(
       type: ObjectId,
       required: true,
     },
+    vendorId: {
```
andela-hpopoola, line 89: I believe this should be added to `test/factories/enquiry.factory.js` too since it is required to make it easier to generate test objects for enquiries.

server/services/enquiry.service.js
```diff
@@ -27,31 +29,35 @@ export const addEnquiry = async (enquiry) => {
...
 
-export const approveEnquiry = async (approvedEnquiry) => {
-  const enquiry = await getEnquiryById(approvedEnquiry.enquiryId).catch((error) => {
+export const approveEnquiry = async ({ enquiryId, vendor }) => {
+  const enquiry = await getEnquiryById(enquiryId).catch((error) => {
     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
   });
 
   try {
+    if (vendor.role === USER_ROLE.VENDOR && vendor._id.toString() !== enquiry.vendorId.toString()) {
```
andela-hpopoola, line 45: We might need to add a test to ensure that this works as expected too.

andela-hpopoola (changes requested):
Awesome. You did a great job here.

Thank you for working on this. The implementation makes sense but I am not satisfied with the tests as some of them are not comprehensive enough.

Can you try to always follow this implementation when writing tests for pagination: Please read this thoroughly to ensure that you understand them and always use them as reference for pagination tests.

1. Test when nothing is in the database (when database table contains no data)
2. Test that the pagination returns the right value (use pagination test 99% of the time, the remaining 1% is for edge case)
3. Test that the right values is returned.  The pagination test only deals with the numbers returned, not the validity of the data returned, so the pagination test cannot replace this

4. Add more customized tests based on the scenarios you are testing. These are tests directly related to the code. Examples are 
    - Testing for Valid user
    - Ensuring that vendor2 doesn't have access to vendor 1 information
    - Ensuring that user2 gets his own message
    - Testing for all if statements used in code with the else part
    - E.t.c.

test/controllers/enquiry.controller.test.js
tobieeniafe: this was added to be sure that the user and admin do not return the same data. the user token uses the pagination tests

server/models/enquiry.model.js
tobieeniafe: this is added automatically in service. gotten from id of the owner of the property

test/controllers/enquiry.controller.test.js
```diff
@@ -510,6 +510,17 @@ describe('Enquiry Controller', () => {
   });
 
   describe('Approve Enquiry', () => {
+    let vendor2Token;
+    const vendor2 = UserFactory.build(
```
andela-hpopoola, line 514: Lets make this `vendorUser2` to make it consistent with `vendorUser`.

test/controllers/enquiry.controller.test.js
```diff
@@ -606,22 +656,132 @@ describe('Enquiry Controller', () => {
...
+              expect(res.body.message).to.be.eql('Enquiry not found');
+              done();
+            });
+        });
+      });
+    });
+
+    context('with a token with valid access', () => {
+      // eslint-disable-next-line array-callback-return
+      [demoUser, vendor2].map((user) => {
```
andela-hpopoola, line 719: Instead of `demoUser`, lets use something like `regularUser2` to make it easier to relate it.  It will even make more sense to add if you can add `1` to the the first one too.

That way, `regularUser1` makes an enquiry on `vendorUser1` property called `enquiry1` and `regularUser2` makes an enquiry on `vendorUser2` called `enquiry2`. This will make it easier to understand the user, vendor and enquiry you are referring to without stress.

Reading through the code was kinda of difficult to relate with because of the naming convention. I had to revisit the variable definition to understand what they are doing.

test/controllers/enquiry.controller.test.js
```diff
@@ -606,22 +656,132 @@ describe('Enquiry Controller', () => {
...
+              expect(res.body).to.have.property('enquiry');
+              expect(res.body.enquiry._id).to.be.eql(vendor2Enquiry._id.toString());
+              done();
+            });
+        });
+      });
+    });
+
+    context('with a token with invalid access', () => {
+      // eslint-disable-next-line array-callback-return
```
andela-hpopoola, line 742: Lets remove this and add `return null` at the end of the test block.

test/controllers/enquiry.controller.test.js
```diff
@@ -606,22 +656,132 @@ describe('Enquiry Controller', () => {
...
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body).to.have.property('enquiry');
+              expect(res.body.enquiry._id).to.be.eql(vendor2Enquiry._id.toString());
+              done();
+            });
+        });
+      });
+    });
+
+    context('with a token with invalid access', () => {
```
andela-hpopoola, line 741: The token actually has a valid access but they are not authorized to view another user's enquiry.

```suggestion
    context('when viewing an unauthorized enquiry', () => {
```

test/controllers/enquiry.controller.test.js
```diff
@@ -606,22 +656,132 @@ describe('Enquiry Controller', () => {
...
+              expect(res).to.have.status(404);
+              expect(res.body.success).to.be.eql(false);
+              expect(res.body.message).to.be.eql('Enquiry not found');
+              done();
+            });
+        });
+      });
+    });
+
+    context('with a token with valid access', () => {
```
andela-hpopoola, line 764:
```suggestion
	    context('when viewing an authorized enquiry', () => {
```

test/controllers/enquiry.controller.test.js
```diff
@@ -657,7 +817,7 @@ describe('Enquiry Controller', () => {
       it('returns the error', (done) => {
         sinon.stub(Enquiry, 'aggregate').throws(new Error('Type Error'));
         request()
-          .get(`/api/v1/enquiry/${enquiry._id}`)
+          .get(`/api/v1/enquiry/${vendorEnquiry._id}`)
```
andela-hpopoola, line 820: Can we revert the variable back to `enquiry`. `vendorEnquiry` sometimes give the enquiry a different meaning as the enquiry is made by the user. Lets use `enquiry` to be more neutral.

andela-hpopoola (changes requested):
Thank you for working on this and implementing the requested changes. Thank you for the response to the comments too, they are really on point 👍🏽 .

I left a few comments on proper naming and removing the eslint comments. I should have approved this but would love you to fix the failing test assertion before approving. Thank you and nice work.

andela-hpopoola (approved):
Thank you for making the changes. This looks good to me.

PR #99, opened 2021-01-13
#176454677 add in review for vendor update

tobieeniafe:
#### What does this PR do?
Add in review for vendor update

Add in review for vendor update

#176454677

### DIFF
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -479,9 +479,7 @@ export const verifyVendor = async ({ vendorId, adminId }) => {
   }
 };
 
-export const updateVendor = async ({ updatedVendor, vendorId }) => {
-  const user = await getUserById(vendorId);
-
+const getStepsReadyForReview = (updatedVendor, user) => {
   const stepToReview = {
     verification: {},
   };
@@ -531,6 +529,14 @@ export const updateVendor = async ({ updatedVendor, vendorId }) => {
       status: VENDOR_INFO_STATUS.IN_REVIEW,
     };
   }
+
+  return stepToReview;
+};
+
+export const updateVendor = async ({ updatedVendor, vendorId }) => {
+  const user = await getUserById(vendorId);
+
+  const stepToReview = getStepsReadyForReview(updatedVendor, user);
 
   if (
     updatedVendor.vendor &&
@@ -634,7 +640,9 @@ export const removeDirector = async ({ directorId, user }) => {
       {
         $set: {
           'vendor.directors': directors,
-          'vendor.verification.directorInfo.status': VENDOR_INFO_STATUS.IN_REVIEW,
+          'vendor.verification.directorInfo.status': user.vendor.verified
+            ? user.vendor.verification.directorInfo.status
+            : VENDOR_INFO_STATUS.IN_REVIEW,
         },
       },
       { new: true, fields: '-password' },
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -480,7 +480,7 @@ export const verifyVendor = async ({ vendorId, adminId }) => {
 };
 
 export const updateVendor = async ({ updatedVendor, vendorId }) => {
-  // const steps = [];
+  const user = await getUserById(vendorId);
 
   const stepToReview = {
     verification: {},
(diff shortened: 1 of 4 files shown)


server/services/user.service.js
```diff
@@ -480,25 +480,68 @@ export const verifyVendor = async ({ vendorId, adminId }) => {
 };
 
 export const updateVendor = async ({ updatedVendor, vendorId }) => {
-  const user = await getUserById(vendorId);
-  if (!user) {
-    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Vendor not found');
-  }
+  // const steps = [];
+
+  const stepToReview = {
```
andela-hpopoola, line 485: This method is too long, so you might want to create a custom function for step review to make it more readable.

```suggestion
  const stepToReview = getStepsReadyForReview(updatedVendor);
```
All of the logic to get the steps should be defined in the function.

```js
const getStepsReadyForReview = (vendor) => {
...
}

```

server/services/user.service.js
```diff
@@ -480,25 +480,68 @@ export const verifyVendor = async ({ vendorId, adminId }) => {
...
+    stepToReview.verification.documentUpload = {
+      status: VENDOR_INFO_STATUS.IN_REVIEW,
+    };
   }
+
+  if (
+    updatedVendor.phone ||
+    updatedVendor.phone2 ||
+    updatedVendor.address ||
+    (updatedVendor.vendor && updatedVendor.vendor.companyName)
```
andela-hpopoola, line 521: You will also need to check for the following fields too:

* companyLogo
* entity
* redanNumber
* socialMedia
* website

Yes, I know that some of them are not used in the current frontend, however, we don't want to wait till they are used before we do the right thing.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments, you are currently on the right track. You will need to remove the commented out code for the previous implementation and write tests to ensure that is ready to be merged. Nice work 👍🏽

server/services/user.service.js
andela-hpopoola: Can you please add the step to review code in a new function as suggested above to make the code easier to read?

server/services/user.service.js
```diff
@@ -585,7 +631,12 @@ export const removeDirector = async ({ directorId, user }) => {
   try {
     return User.findByIdAndUpdate(
       user._id,
-      { $set: { 'vendor.directors': directors } },
+      {
+        $set: {
+          'vendor.directors': directors,
+          'vendor.verification.directorInfo.status': VENDOR_INFO_STATUS.IN_REVIEW,
```
andela-hpopoola, line 637: Can we ensure that this only takes place when an admin has not been verified? If an admin has been verified, there is no need to set this to `IN REVIEW` when updating/deleting directors.

andela-hpopoola: Also, can we set the status of unverified vendor to `IN REVIEW` when a director is deleted as discussed this morning?

test/controllers/user.controller.test.js
```diff
@@ -1796,6 +1792,12 @@ describe('User Controller', () => {
               expect(res.body.user.vendor.taxCertificate).to.be.eql(data.vendor.taxCertificate);
               expect(res.body.user.vendor.socialMedia.length).to.be.eql(2);
               expect(res.body.user.vendor.directors.length).to.be.eql(2);
+              expect(res.body.user.vendor.verification.companyInfo.status).to.be.eql('In Review');
```
andela-hpopoola, line 1795: We will need a different test to verify this. You might also need to add `phone` to the data object (on line 1730) so that we can test that it is also updated.

https://github.com/Highrachy/ballers-api/blob/66fd1c6072febf7127efcf3638cdcd4945da1ef8/test/controllers/user.controller.test.js#L1730

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on ensuring that we cover all edge cases.

test/controllers/user.controller.test.js
```diff
@@ -1917,6 +1918,25 @@ describe('User Controller', () => {
...
+            });
+        });
+      });
+
+      context('when a valid token is used & vendor is not verified', () => {
+        beforeEach(async () => {
+          await User.findByIdAndUpdate(vendorUser._id, { 'vendor.verified': false });
+        });
+
+        it('returns deletes director', (done) => {
```
andela-hpopoola, line 1932:
```suggestion
	        it('sets director verification to `In Review`', (done) => {
```

test/controllers/user.controller.test.js
```diff
@@ -1917,6 +1918,25 @@ describe('User Controller', () => {
               expect(res.body.user.vendor.directors.length).to.be.eql(2);
               expect(res.body.user.vendor.directors[0]._id).to.not.eql(nonSignatoryId);
               expect(res.body.user.vendor.directors[1]._id).to.not.eql(nonSignatoryId);
+              expect(res.body.user.vendor.verification.directorInfo.status).to.be.eql('Pending');
```
andela-hpopoola, line 1921: Can we set the default of this to `verified` so that this returns `Verified`. Having it as `Pending` feels weird for a verified vendor (not a major blocker, it is just to ensure that the future reading of the test is easily understood).

andela-hpopoola (approved):
Thank you for working on this. I added 2 non-blocking comments, so feel free to merge when you have resolved them. Nice work.

PR #102, opened 2021-01-18
#176533740 make vendor unverified when info changes

tobieeniafe:
#### What does this PR do?
Change vendor to unverified when info changes

Change vendor to unverified when info changes

`PUT /api/v1/user/vendor/director`

#176533740

### DIFF
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -552,7 +552,7 @@ export const updateVendor = async ({ updatedVendor, user }) => {
     Array.prototype.push.apply(updatedVendor.vendor.socialMedia, user.vendor.socialMedia);
   }
 
-  const vendor = { ...user.vendor, ...updatedVendor.vendor };
+  const vendor = { ...user.vendor, ...updatedVendor.vendor, verified: false };
 
   vendor.verification = { ...vendor.verification, ...stepToReview.verification };
 
@@ -599,7 +599,10 @@ export const editDirector = async ({ directorInfo, user }) => {
 
   try {
     await User.findByIdAndUpdate(user._id, {
-      $set: { 'vendor.verification.directorInfo.status': VENDOR_INFO_STATUS.IN_REVIEW },
+      $set: {
+        'vendor.verification.directorInfo.status': VENDOR_INFO_STATUS.IN_REVIEW,
+        'vendor.verified': false,
+      },
     });
     return User.findOneAndUpdate(
       { 'vendor.directors._id': directorInfo._id },
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -244,12 +244,70 @@ export const resetPasswordViaToken = async (password, token) => {
   }
 };
 
-export const updateUser = async (updatedUser) => {
+export const updateUser = async ({ updateInfo, user }) => {
+  let vendor;
+  const updatedUser = updateInfo;
+
+  if (user.role === USER_ROLE.VENDOR) {
+    const stepToReview = {
+      verification: {},
+    };
+
+    delete updatedUser.phone;
+    delete updatedUser.phone2;
+    delete updatedUser.address;
+
+    if (
+      updatedUser.vendor &&
+      updatedUser.vendor.directors &&
+      updatedUser.vendor.directors.length > 0
+    ) {
+      stepToReview.verification.directorInfo = {
+        ...user.vendor.verification.directorInfo,
+        status: VENDOR_INFO_STATUS.IN_REVIEW,
+      };
+    }
+
+    if (
+      updatedUser.vendor &&
+      ((updatedUser.vendor.socialMedia && updatedUser.vendor.socialMedia.length > 0) ||
+        updatedUser.vendor.website)
+    ) {
+      stepToReview.verification.companyInfo = {
+        ...user.vendor.verification.companyInfo,
+        status: VENDOR_INFO_STATUS.IN_REVIEW,
+      };
+    }
+
+    if (
+      updatedUser.vendor &&
+      updatedUser.vendor.directors &&
+      updatedUser.vendor.directors.length > 0
+    ) {
+      Array.prototype.push.apply(updatedUser.vendor.directors, user.vendor.directors);
+    }
+
+    if (
+      updatedUser.vendor &&
+      updatedUser.vendor.socialMedia &&
+      updatedUser.vendor.socialMedia.length > 0
+    ) {
+      Array.prototype.push.apply(updatedUser.vendor.socialMedia, user.vendor.socialMedia);
+    }
+
+    vendor = { ...user.vendor, ...updatedUser.vendor };
+
+    vendor.verification = { ...vendor.verification, ...stepToReview.verification };
+  }
+
   try {
-    return User.findOneAndUpdate({ _id: updatedUser.id }, updatedUser, {
-      new: true,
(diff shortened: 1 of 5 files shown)


server/services/user.service.js
```diff
@@ -599,7 +599,10 @@ export const editDirector = async ({ directorInfo, user }) => {
 
   try {
     await User.findByIdAndUpdate(user._id, {
-      $set: { 'vendor.verification.directorInfo.status': VENDOR_INFO_STATUS.IN_REVIEW },
+      $set: {
+        'vendor.verification.directorInfo.status': VENDOR_INFO_STATUS.IN_REVIEW,
+        'vendor.verified': false,
```
andela-hpopoola, line 604: I don't think we need to update this to false as it would prevent vendors from adding new directors and signatories.

andela-hpopoola (changes requested):

For this to work effectively, as we would want the vendor to also update some of his information without putting him back to an `Unverified` status. Remember that `Unverified` will make the user unable to do basic stuff.

I think we need to update the `/user/update/` endpoint to support the updates for the following vendor information:
- adding of directors
- website
- adding of social media (updating and deleting should be done via a different endpoint)

We can do the social media story in a different PR when we need it though.

Thank you.

server/services/user.service.js
tobieeniafe: I believe editing a director info should make the vendor `unverified`. A case scenario where there a verified director's info is edited into fake information is very possible and would be missed. Which circumvents the whole process of director verification

tobieeniafe:
> Thank you for working on this.
> 
> For this to work effectively, as we would want the vendor to also update some of his information without putting him back to an `Unverified` status. Remember that `Unverified` will make the user unable to do basic stuff.
> 
> I think we need to update the `/user/update/` endpoint to support the updates for the following vendor information:
> 
> * adding of directors
> * website
> * adding of social media (updating and deleting should be done via a different endpoint)
> 
> We can do the social media story in a different PR when we need it though.
> 
> Thank you.

A vendor would be able update `phone` and `address` from `/user/update` without becoming unverified tho

server/services/user.service.js
```diff
@@ -244,12 +244,70 @@ export const resetPasswordViaToken = async (password, token) => {
...
+export const updateUser = async ({ updateInfo, user }) => {
+  let vendor;
+  const updatedUser = updateInfo;
+
+  if (user.role === USER_ROLE.VENDOR) {
+    const stepToReview = {
+      verification: {},
+    };
+
+    delete updatedUser.phone;
```
andela-hpopoola, line 256: Not sure why you are deleting this and others below.

server/services/user.service.js
```diff
@@ -244,12 +244,70 @@ export const resetPasswordViaToken = async (password, token) => {
   }
 };
 
-export const updateUser = async (updatedUser) => {
+export const updateUser = async ({ updateInfo, user }) => {
+  let vendor;
+  const updatedUser = updateInfo;
+
+  if (user.role === USER_ROLE.VENDOR) {
+    const stepToReview = {
```
andela-hpopoola, line 252: The  purpose of adding the non essential items here is to allow vendors update their information without triggering a review from the admin. Not sure why you reintroduced it here again?

server/services/user.service.js
```diff
@@ -244,12 +244,70 @@ export const resetPasswordViaToken = async (password, token) => {
...
+      updatedUser.vendor &&
+      updatedUser.vendor.socialMedia &&
+      updatedUser.vendor.socialMedia.length > 0
+    ) {
+      Array.prototype.push.apply(updatedUser.vendor.socialMedia, user.vendor.socialMedia);
+    }
+
+    vendor = { ...user.vendor, ...updatedUser.vendor };
+
+    vendor.verification = { ...vendor.verification, ...stepToReview.verification };
```
andela-hpopoola, line 300: I believe this service shouldn't touch the verification step.

server/services/user.service.js
```diff
@@ -484,16 +542,16 @@ const getStepsReadyForReview = (updatedVendor, user) => {
...
-    updatedVendor.vendor &&
-    updatedVendor.vendor.directors &&
-    updatedVendor.vendor.directors.length > 0
-  ) {
-    stepToReview.verification.directorInfo = {
-      ...user.vendor.verification.directorInfo,
-      status: VENDOR_INFO_STATUS.IN_REVIEW,
-    };
-  }
+  // if (
```
andela-hpopoola, line 545: Not sure why you are removing this here. This is necessary during the initial verification step.

server/services/user.service.js
```diff
@@ -520,9 +578,9 @@ const getStepsReadyForReview = (updatedVendor, user) => {
       (updatedVendor.vendor.companyName ||
         updatedVendor.vendor.companyLogo ||
         updatedVendor.vendor.entity ||
-        updatedVendor.vendor.redanNumber ||
-        (updatedVendor.vendor.socialMedia && updatedVendor.vendor.socialMedia.length > 0) ||
-        updatedVendor.vendor.website))
+        updatedVendor.vendor.redanNumber)) // ||
+    // (updatedVendor.vendor.socialMedia && updatedVendor.vendor.socialMedia.length > 0) ||
```
andela-hpopoola, line 582: Social media and website are necessary too during the initial verification.

server/services/user.service.js
```diff
@@ -536,23 +594,23 @@ const getStepsReadyForReview = (updatedVendor, user) => {
...
   const stepToReview = getStepsReadyForReview(updatedVendor, user);
 
-  if (
-    updatedVendor.vendor &&
-    updatedVendor.vendor.directors &&
-    updatedVendor.vendor.directors.length > 0
-  ) {
-    Array.prototype.push.apply(updatedVendor.vendor.directors, user.vendor.directors);
-  }
+  // if (
```
andela-hpopoola, line 597: Can we return this, please?

server/services/user.service.js
andela-hpopoola: Thank you for the observation. After much thought, I have the following comments:

1. Directors information are needed before verification to ensure the vendor is legit
2. Directors are no longer needed after verification. However, Signatories will still be used for offer letters.
3. Adding and Updating a director information into fake information doesn't add value to a scammer as the director is not used anywhere else. It might affect the signatory because that would mean that a fake signature will be on the offer letter but I don't think it is a major concern.
4. My major concern is that making the vendor `unverified` after they have been verified just because the director's information (like phone number) was updated is a major deal breaker because this would prevent the vendor from doing every major thing like confirming payments, allowing people to view their properties, updating their properties e.t.c
5. Even if we want to make them unverified, it should be only when a signatory is added not because a director is added.

andela-hpopoola (changes requested):
Thank you for the updates. I believe this shouldn't affect the previous service for updating vendor information. I left a few comments on this, though. Thank you.

server/services/user.service.js
```diff
@@ -244,12 +244,70 @@ export const resetPasswordViaToken = async (password, token) => {
   }
 };
 
-export const updateUser = async (updatedUser) => {
+export const updateUser = async ({ updateInfo, user }) => {
+  let vendor;
+  const updatedUser = updateInfo;
```
andela-hpopoola, line 249: We could make use of updatedUser from the parameter 🤷🏽‍♂️

PR #103, opened 2021-01-19
#176533740 allow sensitive data update unverify user

tobieeniafe:
#### What does this PR do?
Change vendor to unverified when sensitive information is updated.

Change vendor to unverified when sensitive information is updated.

#176533740

### DIFF
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -533,7 +533,7 @@ const getStepsReadyForReview = (updatedVendor, user) => {
   return stepToReview;
 };
 
-const containsSensitiveInfo = (updatedVendor) => {
+const containsSensitiveInfo = (user) => {
   const sensitive = [
     'bankInfo',
     'companyLogo',
@@ -544,9 +544,9 @@ const containsSensitiveInfo = (updatedVendor) => {
     'taxCertificate',
   ];
 
-  const isSensitive = sensitive.some((key) => Object.keys(updatedVendor).includes(key));
+  const updatedVendorInfo = Object.keys({ ...user, ...user.vendor });
 
-  return !isSensitive;
+  return sensitive.some((info) => updatedVendorInfo.includes(info));
 };
 
 export const updateVendor = async ({ updatedVendor, user }) => {
@@ -571,9 +571,7 @@ export const updateVendor = async ({ updatedVendor, user }) => {
   const vendor = {
     ...user.vendor,
     ...updatedVendor.vendor,
-    verified: updatedVendor.vendor
-      ? containsSensitiveInfo(updatedVendor.vendor)
-      : user.vendor.verified,
+    verified: containsSensitiveInfo(updatedVendor) ? false : user.vendor.verified,
   };
 
   vendor.verification = { ...vendor.verification, ...stepToReview.verification };
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -533,6 +533,22 @@ const getStepsReadyForReview = (updatedVendor, user) => {
   return stepToReview;
 };
 
+const containsSensitiveInfo = (updatedVendor) => {
+  const sensitive = [
+    'bankInfo',
+    'companyLogo',
+    'companyName',
+    'entity',
+    'identification',
+    'redanNumber',
+    'taxCertificate',
+  ];
+
+  const isSensitive = sensitive.some((key) => Object.keys(updatedVendor).includes(key));
+
+  return !isSensitive;
+};
+
 export const updateVendor = async ({ updatedVendor, user }) => {
   const stepToReview = getStepsReadyForReview(updatedVendor, user);
 
@@ -552,7 +568,13 @@ export const updateVendor = async ({ updatedVendor, user }) => {
     Array.prototype.push.apply(updatedVendor.vendor.socialMedia, user.vendor.socialMedia);
   }
 
-  const vendor = { ...user.vendor, ...updatedVendor.vendor };
+  const vendor = {
+    ...user.vendor,
+    ...updatedVendor.vendor,
+    verified: updatedVendor.vendor
+      ? containsSensitiveInfo(updatedVendor.vendor)
+      : user.vendor.verified,
+  };
 
   vendor.verification = { ...vendor.verification, ...stepToReview.verification };
 
--- a/test/controllers/user.controller.test.js
+++ b/test/controllers/user.controller.test.js
@@ -1704,35 +1704,37 @@ describe('User Controller', () => {
 
     describe('Update vendor info', () => {
       let vendorToken;
-      const vendorId = mongoose.Types.ObjectId();
-      const vendorUser = UserFactory.build({
-        _id: vendorId,
-        role: USER_ROLE.VENDOR,
-        activated: true,
-        phone: '08012345678',
-        phone2: '09012345678',
-        address: AddressFactory.build(),
-        vendor: {
-          companyName: 'Highrachy Investment Limited',
-          directors: [
-            {
-              name: 'Jane Doe',
-              isSignatory: false,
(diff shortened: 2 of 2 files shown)


server/services/user.service.js
```diff
@@ -533,6 +533,22 @@ const getStepsReadyForReview = (updatedVendor, user) => {
   return stepToReview;
 };
 
+const containsSensitiveInfo = (updatedVendor) => {
```
andela-hpopoola, line 536: I believe the function can be greatly improved as the current one caters for vendor information only. While it satisfies the current implementation, it doesn't cater for non vendor attributes like phone, address incase we want to make them sensitive info. 

Here is the improved version (comments added to explain thought process, should be removed in code):

```js

const containsSensitiveInfo = (user) => {
// user attributes are mapped below, so we can add phone, phone2, address e.t.c to the array below
  const sensitiveInfo = [
    'bankInfo',
    'companyLogo',
    'companyName',
    'entity',
    'identification',
    'redanNumber',
    'taxCertificate',
  ];
  
 // here, I am flattening the keys in user and user.vendor to cater for user attributes
  const updatedVendorInfo = Object.keys({ ...user, ...user.vendor });

// there is no need assigning this to a variable, we can return the value asap
  return sensitiveInfo.some((info) => updatedVendorInfo.includes(info));
};
```

server/services/user.service.js
```diff
@@ -552,7 +568,13 @@ export const updateVendor = async ({ updatedVendor, user }) => {
     Array.prototype.push.apply(updatedVendor.vendor.socialMedia, user.vendor.socialMedia);
   }
 
-  const vendor = { ...user.vendor, ...updatedVendor.vendor };
+  const vendor = {
+    ...user.vendor,
+    ...updatedVendor.vendor,
+    verified: updatedVendor.vendor
```
andela-hpopoola, line 574: Also here, you didn't that non-vendor attributes can be sensitive too 😏 

```suggestion
    verified: containsSensitiveInfo(updatedVendor) ? false : user.vendor.verified,
```

test/controllers/user.controller.test.js
```diff
@@ -1815,6 +1816,55 @@ describe('User Controller', () => {
         });
       });
 
+      context('when non sensitive data is sent', () => {
```
andela-hpopoola, line 1819: To properly test for sensitive data, we would need to add 4 tests assertions

1. When a vendor has been verified
   - updating sensitive information
   - updating non-sensitive information

2. When a vendor has not been verified
   - updating sensitive information
   - updating non-sensitive information

You might want to consider my offer of creating a custom vendor factory to simplify vendor creation 🤷🏽‍♂️

andela-hpopoola (changes requested):
Thank you for working on this. The current implementation looks great and much better than the previous one. Thank you for making time spent on brainstorming this worthwhile.

Your implementation looks great, however, you didn't cover sensitive information from a non vendor information view (like phone number). This is a major deal-breaker.

For your tests. when you add new functions or implementation, always ensure you add new   tests to ensure that they work as expected instead of updating the old ones. Edge cases should be considered and tests should be written for them too. It is not certain when this is necessary, so you will have to use your discretion to determine if a new test is needed or not.

test/controllers/user.controller.test.js
```diff
@@ -1815,6 +1843,98 @@ describe('User Controller', () => {
...
+              });
+              expect(res.body.user.vendor.companyName).to.be.eql(vendorUser.vendor.companyName);
+              expect(res.body.user.vendor.companyLogo).to.be.eql(data.vendor.companyLogo);
+              expect(res.body.user.vendor.bankInfo).to.be.eql(data.vendor.bankInfo);
+              expect(res.body.user.vendor.entity).to.be.eql(data.vendor.entity);
+              expect(res.body.user.vendor.taxCertificate).to.be.eql(data.vendor.taxCertificate);
+              expect(res.body.user.vendor.socialMedia.length).to.be.eql(2);
+              expect(res.body.user.vendor.directors.length).to.be.eql(2);
+              expect(res.body.user.vendor.verification.companyInfo.status).to.be.eql('In Review');
+              expect(res.body.user.vendor.verification.bankDetails.status).to.be.eql('In Review');
```
andela-hpopoola, line 1900: We need to add this assertion to the other 3 tests to be sure that they work as expected.

test/controllers/user.controller.test.js
```diff
@@ -1815,6 +1843,98 @@ describe('User Controller', () => {
...
+        beforeEach(async () => {
+          await User.findByIdAndUpdate(vendorUser._id, { 'vendor.verified': false });
+        });
+        it('returns updated vendor', (done) => {
+          request()
+            [method](endpoint)
+            .set('authorization', vendorToken)
+            .send(data)
+            .end((err, res) => {
+              expect(res).to.have.status(200);
```
andela-hpopoola, line 1881: Since we are expecting the same output from the 4 verification tests (Only res.body.user.vendor.verified is changing) you could write a custom function to cater for the validations instead of repeating them.

Improving the comment on grouping similar tests, here is a skeletal breakdown of my workflow (please note that it is a skeletal view only just to explain how it should look like)

```js
describe('Vendor verification', () => {
 ....
  const itReturnsUpdatedVendor = (res, data) => {
              expect(res.body.message).to.be.eql('Vendor information updated');
              expect(res.body.user._id).to.be.eql(vendorUser._id.toString());
              expect(res.body.user.phone).to.be.eql(data.phone);
              expect(res.body.user.phone2).to.be.eql(data.phone2);
              expect(res.body.user.vendor.verified).to.be.eql(false);
              expect(res.body.user.address).to.be.eql({
                ...vendorUser.address,
                ...data.address,
              });
              expect(res.body.user.vendor.companyName).to.be.eql(vendorUser.vendor.companyName);
              expect(res.body.user.vendor.companyLogo).to.be.eql(data.vendor.companyLogo);
              expect(res.body.user.vendor.bankInfo).to.be.eql(data.vendor.bankInfo);
              expect(res.body.user.vendor.entity).to.be.eql(data.vendor.entity);
              expect(res.body.user.vendor.taxCertificate).to.be.eql(data.vendor.taxCertificate);
              expect(res.body.user.vendor.socialMedia.length).to.be.eql(2);
              expect(res.body.user.vendor.directors.length).to.be.eql(2);
              expect(res.body.user.vendor.verification.companyInfo.status).to.be.eql('In Review');
              expect(res.body.user.vendor.verification.bankDetails.status).to.be.eql('In Review');
             ....
  }
   context('when vendor is verified', () => {
       context('when sensitive data is sent', () => {
        ....
          .end((err, res) => {
           itReturnsUpdatedVendor(res, data);
           // I didn't add the below assertion to the function to make it easier for a newbie to understand the purpose of the test without visiting the function
           expect(res.body.user.vendor.verified).to.be.eql(false);
         }
       }
       context('when non sensitive data is sent', () => {....
          .end((err, res) => {
           itReturnsUpdatedVendor(res, data);
           expect(res.body.user.vendor.verified).to.be.eql(true);
         }
       }
   })

   context('when vendor is unverified', () => {
       context('when sensitive data is sent', () => {
....
          .end((err, res) => {
           itReturnsUpdatedVendor(res, data);
           expect(res.body.user.vendor.verified).to.be.eql(false);
         }
       }
       context('when non sensitive data is sent', () => {
....
          .end((err, res) => {
           itReturnsUpdatedVendor(res, data);
           expect(res.body.user.vendor.verified).to.be.eql(false);
         }
       }
  })
});
```

andela-hpopoola (changes requested):
Thank you for the updates. I left more comments on ensuring that the tests are properly grouped and also created a function for recurring tests.

andela-hpopoola (approved):
Thank you for working on this. Looks good to me.

PR #105, opened 2021-01-22
#176533718 update Joi schema validation

tobieeniafe:
#### What does this PR do?
Update Joi schema validation

Update Joi schema validation

#176533718


andela-hpopoola (changes requested):
Thank you for working on this. I believe we can do a better job with our schema definition. Here are major things we can do to improve:

1. Let's have a `helper.schema.js` file and define all of our schema definitions inside (like requiredString, optionalString, email, password, phone, address e.t.c. This can easily be imported into where they would be used.  The best part is that it provides a single source of truth for schema definition which would make it easier for us to reuse and update predefined schemas.

2. The definition of optionalString is confusing as we have 2 types of optional fields in the code
   a. **Schema Optional Fields** - These are optional fields in the db schema itself. They are fields that doesn't have `required: true`, example of this in the userSchema are phone2, profileImage e.t.c
   b. **Optional Update Fields** - These are fields that are required in the db schema but are made optional in Joi schema because we are updating. One of reasons why we have to do this is to ensure that we can update part of the database without supplying all the existing values. 
   
For this update, lets define a `notRequired` function that shows that an joi property is optional and accepts  empty string and null. This will be used for `Schema Optional Fields` defined in `a` above. 

We should update the optionalString function and ensure it doesn't accept empty or null field as it will be used for `Optional Update Fields`. Failing tests should be fixed to ensure we are making the right changes.

3. Schemas that are not reusable can be defined in the schema file they are used. For example, the preference joi schema is specific to the user schema, so there is no need to define it in helper.schema.js file, that can be done in user.schema.js.

I hope this clarifies everything on the schema update. Please let me know if you have any questions about this or anything else. Thank you.

server/schemas/user.schema.js
```diff
@@ -1,21 +1,27 @@
...
+  optionalPhoneNumber,
+  optionalString,
+  optionalBoolean,
+  optionalNumber,
+  optionalAddress,
+  notRequiredPhoneNumber,
+} from './helper.schema';
+
+const email = requiredEmail('Email Address');
 const password = Joi.string().label('Password').min(6).required().strict();
```
andela-hpopoola, line 17: Lets define this and confirmPassword in helper.schema with the current naming convention:

```js
requiredPassword
requiredConfirmPassword
```

server/schemas/user.schema.js
```diff
@@ -67,7 +73,7 @@ export const registerSchema = Joi.object({
   lastName: requiredString('Last Name'),
   vendor: vendorCompanyName,
   email,
-  phone,
+  phone: notRequiredPhoneNumber('Phone'),
```
andela-hpopoola, line 76: For registration, I believe the phone number is required 🤷🏽‍♂️

server/schemas/visitation.schema.js
```diff
@@ -1,20 +1,18 @@
...
-  visitDate: Joi.date()
-    .greater(getTodaysDateStandard())
-    .label('Visit Date')
-    .required()
-    .messages({
-      'date.greater': `"Visit Date" should a date later than ${getTodaysDateInWords()}`,
-    }),
+  propertyId: requiredObjectId('Property id'),
+  visitorName: requiredString('Name'),
+  visitorEmail: optionalEmail('Email address'),
```
andela-hpopoola, line 13: This should be `nonRequiredEmail`

andela-hpopoola (changes requested):
Thank you for working on this. Our Joi schema now looks beautiful 😺 

I left a couple of comments on ensuring that the validations are testing for the right thing. Please let me know if I made a wrong assumption anywhere. Thank you.

server/schemas/user.schema.js
tobieeniafe: I thought about this too. But following your advice, since `password` is only used in the `user.schema.js` file, there was no need to define it outside the file.

andela-hpopoola: Actually, the password would always be reused -- when we create an endpoint to 'change password', we would need to confirm that the new password is secured too 😏. However, if you think otherwise or you feel we should create a function at the time it is needed, then it is better to make it an inline definition. 

The bone of contention here is to ensure that the function definition follows the pattern you have created which makes a Joi function either be `requiredFunctionName`, `optionalFunctionName` or `nonRequiredFunctionName`. Following this pattern makes life for everyone including yourself in the future. 

I saw some other definitions that didn't follow this pattern though like `step`, `director`, I left them since I feel they would be hardly reused and can be modified later. `Password` and `confirmPassword` would sure be reused, that is why I am more concerned about them.

server/models/user.model.js
```diff
@@ -127,7 +127,10 @@ const UserSchema = new mongoose.Schema(
       required: true,
       select: false,
     },
-    phone: String,
+    phone: {
+      type: String,
+      required: true,
```
andela-hpopoola, line 132: Not sure if making the phone number is a good idea. We might need to discuss with others to make important changes like this. Please lets return it back to optional.

andela-hpopoola (approved):
Just one more comment before merging 🙏🏽 . Please let's return the phone number back to optional for the user schema. 

Feel free to merge when you have fixed this. Thank you and Nice work.

server/models/user.model.js
tobieeniafe: but you said [this](https://github.com/Highrachy/ballers-api/pull/105#discussion_r563030067) earlier. so should phone number be `required` or `notRequired`?

PR #106, opened 2021-01-24
#176533745 add routes to ban and unban user

tobieeniafe:
#### What does this PR do?
Add endpoints to enable an admin ban or unban a user

Add endpoints to enable an admin ban or unban a user

`PUT /api/v1/user/ban`
`PUT /api/v1/user/unban`

#176533745

### DIFF
--- a/server/helpers/middleware.js
+++ b/server/helpers/middleware.js
@@ -33,7 +33,8 @@ export const authenticate = async (req, res, next) => {
       if (user.banned.status) {
         return res.status(httpStatus.UNAUTHORIZED).json({
           success: false,
-          message: 'Account has been banned. Contact admin for more information',
+          message:
+            'Your account has been locked. Kindly contact Ballers Support for more information',
         });
       }
       req.user = user.toJSON();
--- a/server/helpers/middleware.js
+++ b/server/helpers/middleware.js
@@ -30,6 +30,12 @@ export const authenticate = async (req, res, next) => {
     const { id } = decodeToken(req.headers.authorization);
     const user = await getUserById(id);
     if (user) {
+      if (user.banned.status) {
+        return res.status(httpStatus.UNAUTHORIZED).json({
+          success: false,
+          message: 'Account has been banned. Contact admin for more information',
+        });
+      }
       req.user = user.toJSON();
       next();
     } else {
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -89,6 +89,21 @@ const UserSchema = new mongoose.Schema(
     },
     activationDate: Date,
     address: addressSchema,
+    banned: {
+      by: {
+        type: ObjectId,
+      },
+      date: {
+        type: Date,
+      },
+      status: {
+        type: Boolean,
+        default: false,
+      },
+      reason: {
+        type: [String],
+      },
+    },
     email: {
       type: String,
       unique: true,
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -90,19 +90,32 @@ const UserSchema = new mongoose.Schema(
     activationDate: Date,
     address: addressSchema,
     banned: {
-      by: {
-        type: ObjectId,
-      },
-      date: {
-        type: Date,
-      },
       status: {
         type: Boolean,
         default: false,
       },
-      reason: {
-        type: [String],
-      },
+      info: [
+        {
+          bannedBy: {
+            type: ObjectId,
+          },
+          bannedDate: {
+            type: Date,
+          },
+          bannedReason: {
+            type: String,
+          },
+          unBannedBy: {
+            type: ObjectId,
+          },
+          unBannedDate: {
+            type: Date,
+          },
+          unBannedReason: {
+            type: String,
+          },
+        },
+      ],
     },
     email: {
       type: String,
--- a/server/models/user.model.js
+++ b/server/models/user.model.js
@@ -94,7 +94,7 @@ const UserSchema = new mongoose.Schema(
         type: Boolean,
         default: false,
       },
-      info: [
+      case: [
         {
           bannedBy: {
             type: ObjectId,
--- a/server/schemas/user.schema.js
+++ b/server/schemas/user.schema.js
@@ -152,7 +152,13 @@ export const resolveCommentVendorSchema = Joi.object({
   commentId: requiredObjectId('Comment id'),
 });
 
-export const banOrUnbanUserSchema = Joi.object({
+export const banUserSchema = Joi.object({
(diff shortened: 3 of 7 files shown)


server/models/user.model.js
```diff
@@ -89,6 +89,21 @@ const UserSchema = new mongoose.Schema(
     },
     activationDate: Date,
     address: addressSchema,
+    banned: {
```
andela-hpopoola, line 92: Hmmm 🤔 

Your implementation here is slightly different from what we discussed. With your current implementation, we would only have comprehensive information on the last update and won't know when last the user was banned and who banned them (as you are saving only the comment in the array). 

Banning is a sensitive information so we would want to keep all this information as they would be used for audit / security logs incase there are discrepancies on the reason why the user is banned. Also, it makes it easier for multiple admin to keep track of the reasons a user is being banned.

To make this more productive, I believe this will be a better schema:

```js
banned: {
   status: {
      type: Boolean,
      default: false,
    },
    info: [{ 
     bannedBy: {
        type: ObjectId,
      },
     bannedDate: {
       type: Date
     },
     bannedReason: {
         type: String,
     },
      unBannedBy: {
        type: ObjectId,
      },
     unBannedDate: {
       type: Date
     },
     unBannedReason: {
         type: String,
     },
     }]
}

```

andela-hpopoola (changes requested):
Thank you for the quick draft on this. 

I didn't bother to review further as the schema needs to be slightly improved to ensure we have as much information as necessary when a user is banned.

server/schemas/user.schema.js
```diff
@@ -152,7 +152,13 @@ export const resolveCommentVendorSchema = Joi.object({
...
 });
 
-export const banOrUnbanUserSchema = Joi.object({
+export const banUserSchema = Joi.object({
+  userId: requiredObjectId('User id'),
+  reason: requiredString('Reason'),
+});
+
+export const unbanUserSchema = Joi.object({
+  caseId: requiredObjectId('Case id'),
```
andela-hpopoola, line 161: I love the use of `case` here. Can we change the info I suggested in the schema to `case` to make them uniform 🙏🏽

server/models/user.model.js
```diff
@@ -90,19 +90,32 @@ const UserSchema = new mongoose.Schema(
...
-        type: Date,
-      },
       status: {
         type: Boolean,
         default: false,
       },
-      reason: {
-        type: [String],
-      },
+      info: [
```
andela-hpopoola, line 97: As suggested in the last comment:

```suggestion
      case: [
```

server/services/user.service.js
```diff
@@ -756,20 +756,44 @@ export const banOrUnbanUser = async ({ adminId, userId, reason, status }) => {
...
+  const user = await getUserById(userId);
+
+  if (!user) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'User not found');
+  }
+
+  try {
+    await User.findByIdAndUpdate(userId, { $set: { 'banned.status': false } });
+
+    return User.findOneAndUpdate(
```
andela-hpopoola, line 785: Just an edge case, I feel we should add the comment first before unbanning the user since we are using the comments to track when the user is banned and by who so that if the comment addition fails for any reason (could be server error), the user is not unbanned and the admin can try again. 

This might not actually happen but increases our confidence that we catered for it.

andela-hpopoola (changes requested):
Thank you for making the changes. This looks much better now to me. I left a few comments, this should be good to go when you add tests. Nice Work and thank you for the awesome implementation.

server/helpers/middleware.js
```diff
@@ -30,6 +30,12 @@ export const authenticate = async (req, res, next) => {
     const { id } = decodeToken(req.headers.authorization);
     const user = await getUserById(id);
     if (user) {
+      if (user.banned.status) {
+        return res.status(httpStatus.UNAUTHORIZED).json({
+          success: false,
+          message: 'Account has been banned. Contact admin for more information',
```
andela-hpopoola, line 36:
```suggestion
          message: 'Your account has been locked. Kindly contact Ballers Support for more information',
```

test/controllers/user.controller.test.js
```diff
@@ -2847,6 +2847,178 @@ describe('User Controller', () => {
         });
       });
     });
+
+    describe('Ban User', () => {
+      const user = UserFactory.build(
+        {
+          role: USER_ROLE.USER,
+          activated: true,
+          banned: { status: false },
```
andela-hpopoola, line 2856: Since we have added banned to `authenticate` middleware, let's add this to the user.factory.js file so that it can be the default state.

test/controllers/user.controller.test.js
```diff
@@ -2847,6 +2847,178 @@ describe('User Controller', () => {
...
+        userId: user._id,
+        reason: 'Suspected fraud',
+      };
+
+      beforeEach(async () => {
+        await addUser(user);
+      });
+
+      context('when a valid token is used', () => {
+        it('returns banned user', (done) => {
```
andela-hpopoola, line 2874: Can we add another test that already contains a ban case to ensure that when a user is banned again, the new case is added to the database without overwriting the existing one?

test/controllers/user.controller.test.js
```diff
@@ -2847,6 +2847,178 @@ describe('User Controller', () => {
...
+          role: USER_ROLE.USER,
+          activated: true,
+          banned: {
+            status: true,
+            case: [
+              {
+                _id: mongoose.Types.ObjectId(),
+                bannedBy: adminUser._id,
+                bannedReason: 'Suspected fraud',
+                unBannedBy: '5fd3794445973e647c624d49',
```
andela-hpopoola, line 2942: In the real world scenario, I believe the `unBanned...` properties should be empty / not yet created when a user is banned. Let's replicate that for a more robust test by removing the unBanned...`  properties.

andela-hpopoola (changes requested):
Thank you for the updates. I left a few more comments on the tests to perfect this story. Thank you for the awesome job.

test/controllers/user.controller.test.js
tobieeniafe: banned `status` is `false` by default. addinng it to the user factory would flag a validation error with the joi schema

andela-hpopoola: Nice catch 🤦🏽 . I totally forgot about JOI. Thank you for the reminder.

andela-hpopoola (approved):
This looks good to me. Thank you for the hard work.

PR #110, opened 2021-01-26
#176630115 add what changed to vendor verification

tobieeniafe:
#### What does this PR do?
Add what changed to vendor verification

Add what changed to vendor verification to help admin identify updated fields

`PUT /api/v1/user/vendor/verify`

#176630115

### DIFF
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -551,12 +551,30 @@ const containsSensitiveInfo = (user) => {
   ];
 
   const updatedVendorInfo = Object.keys({ ...user, ...user.vendor });
-  const logs = sensitive.filter((element) => updatedVendorInfo.includes(element));
+  const updatedFields = sensitive.filter((element) => updatedVendorInfo.includes(element));
 
   return {
-    status: logs.length > 0,
-    logs,
+    status: updatedFields.length > 0,
+    updatedFields,
   };
+};
+
+export const generateLog = ({ updatedFields, updatedVendor, user }) => {
+  const oldValues = { ...user, ...user.vendor };
+  const newValues = { ...updatedVendor, ...updatedVendor.vendor };
+  const log = [];
+
+  updatedFields.forEach((field) => {
+    log.push({
+      [field]: {
+        old: JSON.stringify(oldValues[field]),
+        new: JSON.stringify(newValues[field]),
+      },
+    });
+  });
+  const updateDate = { updatedAt: getTodaysDateStandard() };
+
+  return Object.assign({}, updateDate, ...log);
 };
 
 export const updateVendor = async ({ updatedVendor, user }) => {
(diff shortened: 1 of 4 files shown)


test/controllers/user.controller.test.js
```diff
@@ -2164,6 +2166,16 @@ describe('User Controller', () => {
         },
       };
 
+      const updatedFields = [
```
andela-hpopoola, line 2169: We can generate the updated fields from sensitiveData by 
```js
      let updatedFields = Object.keys({ ...sensitiveData, ...sensitiveData.vendor });
      updatedFields = updatedFields.filter((key) => {
        return key !== 'vendor';
      });
```

Please note that the test will fail because you forgot to remove phone number from the list of sensitive data here

https://github.com/Highrachy/ballers-api/blob/1e38acfd7adcaa01e54ceea36af96d1e18a2aa12/test/controllers/user.controller.test.js#L2151

server/services/user.service.js
```diff
@@ -552,7 +553,10 @@ const containsSensitiveInfo = (user) => {
 
   const updatedVendorInfo = Object.keys({ ...user, ...user.vendor });
 
-  return sensitive.some((info) => updatedVendorInfo.includes(info));
+  return {
+    status: sensitive.some((info) => updatedVendorInfo.includes(info)),
```
andela-hpopoola, line 557: 🤔 I don't think we need to run 2 loops in this object. Some returns true if the array returns at least one true which satisfies the loop used in updatedFields below. We can optimize this by:

```js
 const updatedFields = sensitive.filter((element) => updatedVendorInfo.includes(element));

 return {
    status: updatedFields.length > 0,
    updatedFields,
  };
```

andela-hpopoola (changes requested):
Thank you for working on this. Your implementation looks good and you have done an awesome job. Kudos to your new-found zeal and commitment, it is really impressive.

However, after so much thoughts and deliberation, the implementation can still be improved for a better experience. Not knowing the value that changed and when it changed can be a major deal breaker to the admin.

We can improve this by doing the following:
1. Changing the updatedFields field in user schema to logs
2. Adding a updatedAt field to the schema
3. Not clearing the previous log when the admin has verified the vendor
4. Maybe getting user ip, city, and country to enable the admin to detect fraudulent activities.
5. Subsequent updates should be added as the first item to the log

Here is how the schema will look like afterward

```
  logs: [{
    phone: { old: 'xxxx', new: 'yyyyy' }
    bankInfo: { old: { accountName: ...}, new: { accountName: ...}}
    ...
    ip,
    city,
    country,
    updatedAt,
  }]
```

test/controllers/user.controller.test.js
tobieeniafe: According to what was discussed on monday, `phone` is a sensitive data. Its because the array isn't sorted. Adding `.sort()` fixed it

test/controllers/user.controller.test.js
```diff
@@ -2217,7 +2212,7 @@ describe('User Controller', () => {
         expect(res.body.user.vendor.verification.companyInfo.status).to.be.eql('In Review');
         expect(res.body.user.vendor.verification.documentUpload.status).to.be.eql('In Review');
         expect(res.body.user.vendor.verification.directorInfo.status).to.be.eql('Pending');
-        expect(res.body.user.vendor.updatedFields).to.be.eql(updatedFields);
+        // expect(res.body.user.vendor.logs).to.be.eql(logs);
```
andela-hpopoola, line 2215: It might be better to write an entirely separate test for logs.

andela-hpopoola (changes requested):
Thank you for working on this. Even though I know you have not completed this task, I feel you approached it in the wrong direction.

Here is how I would have approached mine:

1. Create a function to generate log
2. Use the updatedFields to generate the old and new value
3. Append this to the existing log.


Here is the pseudocode:

https://github.com/Highrachy/ballers-api/blob/99e9ec15de50901442a6d8e6b5e95adb244561af/server/services/user.service.js#L581

```js
// starting from line 581 above
  const { updatedFields, status } = containsSensitiveInfo(updatedVendor);

 // generate the log
  const log = generateLog({
      updatedFields, // updated fields returned from sensitive info
      updatedVendor, // the fields the user wants to update
      user, // the old values in the database
    });


  // the normal parole for the logs schema in the database
   const logs = [log, ...user.vendor.logs];
```

Here is the function to generate the log:

```js
const generateLog = ({ updatedFields, updatedVendor, user }) => {
  const oldValues = { ...user, ...user.vendor }; // spreading out so that it works with updatedFields
  const newValues = { ...updatedVendor, ...updatedVendor.vendor }; // same as above

  const log = []; // log container

  updatedFields.forEach((field) => {
    log.push({
      [field]: { // e.g bankInfo
        old: JSON.stringify(oldValues[field]), // e.g undefined -- this can contain a value like below too but it would be undefined the first time we are updating.
        new: JSON.stringify(newValues[field]), // e.g '{"accountNumber":"1234567890","accountName":"Highrachy Investment Limited","bankName":"ABC Bank"}'
      },
    });
  });

  return log;
};
```

I hope this helps.

andela-hpopoola (approved):
Nice work. This looks perfect to me. 

One more thing before you merge please, lets add another test for logs when the user has existing logs to confirm that the new log is the first in the logs and the old logs are not overwritten. To be more specific, the vendor should have existing log before the test, we will make changes in the test, and confirm that the new changes are saved in the log and the old one still exists too. 

Thank you and great work.

PR #112, opened 2021-01-28
#176630222 implement filter functionality for `GET users/all` endpoint

tobieeniafe:
#### What does this PR do?
Implement filter functionality for `GET users/all` endpoint

-Implement filter functionality for `GET users/all` endpoint.
-Deprecate `/api/v1/user/vendor/all` endpoint.

`GET /api/v1/user/all`

```
activated: BOOLEAN 
activationDate: DATE 
banned: BOOLEAN 
bankDetails: STRING 
certified: BOOLEAN 
certifiedOn: DATE
city: STRING 
companyInfo: STRING 
companyName: STRING 
country: STRING 
createdAt:DATE 
directorInfo:  STRING 
documentUpload: STRING 
email:STRING 
entity: STRING 
firstName:STRING 
lastName:STRING 
phone:STRING 
phone2:STRING 
referralCode:STRING 
redanNumber:  STRING 
role:INTEGER
state: STRING 
street1: STRING 
street2: STRING 
verified:  BOOLEAN 
verifiedOn: DATE 

```

#176630222


server/controllers/user.controllers.js
```diff
@@ -134,8 +133,8 @@ const UserController = {
   },
 
   getAllRegisteredUsers(req, res, next) {
```
andela-hpopoola, line 135: We might need to deprecate this service too and create a `getAllUsers` that works with filter and do all of the hardwork for services like this. 

In the new implementation, we can do the following to get this
- to get all vendors, we will use the `role=2` filter (we can also use role=vendor from the request and map to 2 in the services)
- to get all certified users, we will use the `registered=true` filter
- to get registered vendors, we will use the `role=2&registered=true` filter
- to get all certified vendors, we will use the `role=2&certified=true` filter e.t.c.

server/services/user.service.js
```diff
@@ -261,24 +261,34 @@ export const updateUser = async (updatedUser) => {
...
     { $project: { preferences: 0, password: 0, notifications: 0 } },
-  ]);
+  ];
+
+  if (role === USER_ROLE.VENDOR.toString()) {
+    userOptions.unshift({ $sort: { 'vendor.verified': 1, 'vendor.companyName': 1 } });
+  }
+
+  if (role === USER_ROLE.USER.toString()) {
+    userOptions.unshift({
```
andela-hpopoola, line 281: Not sure if this will still be needed if we are listing all users.

test/controllers/user.controller.test.js
```diff
@@ -3104,36 +3106,27 @@ describe('User Controller', () => {
...
-        user: adminUser,
-        model: User,
-        modelMethod: 'aggregate',
-      });
-      itReturnsNotFoundForInvalidToken({
-        endpoint,
-        method,
-        user: adminUser,
-        userId: adminUser._id,
+    context('when role for vendor is passed', () => {
```
andela-hpopoola, line 3110: Can we ensure we test for editors too and admin roles too.

andela-hpopoola (changes requested):
Thank you for working on this. Awesome job as usual, I believe this will be a good stepping stone for the filter implementation. 

I also believe that we might still deprecate the `getAllRegisteredUsers` service and use an encompassing solution to cater for similar endpoints.

server/controllers/user.controllers.js
```diff
@@ -132,9 +132,35 @@ const UserController = {
       .catch((error) => next(error));
   },
 
-  getAllRegisteredUsers(req, res, next) {
-    const { page, limit, role } = req.query;
-    getAllRegisteredUsers({ page, limit, role })
+  getAllUsers(req, res, next) {
+    const {
```
andela-hpopoola, line 136: I think it is better to leave the destructuring for the service so that we have just one point to update the query information instead of multiple places.

server/services/user.service.js
```diff
@@ -261,9 +261,38 @@ export const updateUser = async (updatedUser) => {
   }
 };
 
-export const getAllRegisteredUsers = async ({ page = 1, limit = 10, role }) => {
+export const getAllUsers = async ({
+  page = 1,
+  limit = 10,
+  role,
```
andela-hpopoola, line 267: You might just destructure page and limit and pass the rest to query to prevent multiple updates as the remaining values will be later grouped.

```js
export const getAllUsers = async ({ page = 1, limit = 10, ...query} => ....
```

server/services/user.service.js
```diff
@@ -261,9 +261,38 @@ export const updateUser = async (updatedUser) => {
...
+  verified,
+  certified,
+  companyName,
+}) => {
   const userOptions = [
-    { $match: { role: role ? parseInt(role, 10) : { $exists: true } } },
+    {
+      $match: {
+        $and: [
+          { role: role ? parseInt(role, 10) : { $exists: true } },
```
andela-hpopoola, line 282: Can you create a loop so that we add the property only if it is needed. 

Another pointer is to group all the filters into different data types and use a loop to create the object.

For example:

```js
// Disclaimer: This is just a thought process, this code is not intended as the final code so it is not safe or tested. Please feel free to change everything or the whole implementation into something better

// all queries are saved here
let filterQuery = [];

// we would have boolean keys, time keys e.t.c.
// any property that doesn't satisfy any of this group, we would handle them individually
const stringKeys = [
'firstName',
'lastName',
'email'
'phone'
'referralCode',
];
// loop that will be repeated for each group
  stringKeys.forEach((key) => {
    if (query[key]) { // query is destructed from above comment
      filterQuery.push { [key]: ({ $regex: key, $options: 'i' });
    }
  });

 // same loop is done for time-based filters, boolean filters, number filters, and special filters

// After the loop, you could can use filterQuery on line 282:
  $match: {
        $and: filterQuery
  }
...

```

test/services/user.service.test.js
```diff
@@ -692,10 +692,10 @@ describe('User Service', () => {
     });
     context('when new user is added', async () => {
       before(async () => {
-        await User.create(UserFactory.build());
+        await addUser(UserFactory.build());
       });
       it('returns total users plus one', async () => {
-        const users = await getAllRegisteredUsers();
+        const users = await getAllUsers({});
```
andela-hpopoola, line 698: Can we set a default object for the service so that we don't edit this test.

test/controllers/user.controller.test.js
```diff
@@ -3113,21 +3131,67 @@ describe('User Controller', () => {
...
+              result: 1,
+              total: 1,
+              totalPage: 1,
+            });
+            expect(res.body.result[0].role).to.be.eql(USER_ROLE.ADMIN);
+            done();
+          });
+      });
+    });
+
```
andela-hpopoola, line 3173: We need a comprehensive test for the filters
1. A looped test to test all the filters one by one
2. Another test to test combined filters and ensure they work as expected.

andela-hpopoola (changes requested):
Thank you for the updates. Added a few comments to guide you. Thank you.

server/helpers/constants.js
```diff
@@ -43,18 +43,23 @@ export const VENDOR_STEPS = ['companyInfo', 'bankDetails', 'directorInfo', 'docu
 export const addressSchema = {
   city: {
     type: String,
+    default: null,
```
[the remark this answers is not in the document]
tobieeniafe: if the default isn't null, adding `address` to the filter would return not return newly registered users that haven't added their addresses.

server/services/user.service.js
```diff
@@ -261,9 +261,38 @@ export const updateUser = async (updatedUser) => {
...
+  role,
+  activated,
+  firstName,
+  lastName,
+  email,
+  phone,
+  referralCode,
+  verified,
+  certified,
+  companyName,
```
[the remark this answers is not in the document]
tobieeniafe: adding fields that are not added on creation, breaks the filter because of the `{ $exists: true }` field which acts as a wildcard search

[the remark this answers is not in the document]
andela-hpopoola: I am not sure I understand your point. I feel we should not add { $exists: true } as a default for filters that are not given in a request. If the only property in `req.query` is `firstName` then the query should run with `firstName` only and not use `$exists: true` for other ones. 

At the end of the day the generated `and` query should be like if a user searches for the firstName only:
```js
$and: [
{ firstName: { $regex: firstName, $options: 'i' }  }
]
```

If it is `firstName` and `lastName`, then the generated query should look like this:
```js
$and: [
{ firstName: { $regex: firstName, $options: 'i' }  },
{ lastName: { $regex: lastName, $options: 'i' }  },
]
```

I believe the query should only **include what is needed**, we should not set a default for what the user is not searching for. I don't see any reason why we should use `exists: true` if a field is not given in the query.

The same thing applies to the address comment above this.

server/services/user.service.js
andela-hpopoola: Achieving the above comment is possible with the logic explained here:
https://github.com/Highrachy/ballers-api/pull/112#discussion_r567625021

server/services/user.service.js
```diff
@@ -261,24 +261,64 @@ export const updateUser = async (updatedUser) => {
...
-    {
-      $lookup: {
-        from: 'properties',
-        localField: '_id',
-        foreignField: 'assignedTo',
-        as: 'assignedProperties',
-      },
-    },
+export const getAllUsers = async ({ page = 1, limit = 10, ...query }) => {
+  const filterQuery = [{ role: query.role ? parseInt(query.role, 10) : { $exists: true } }];
```
andela-hpopoola, line 265: `role` should be added as an optional filter too. It can be added as `integerKeys`

server/services/user.service.js
```diff
@@ -261,24 +261,64 @@ export const updateUser = async (updatedUser) => {
...
+  const stringKeys = ['firstName', 'lastName', 'email', 'phone', 'referralCode'];
```
andela-hpopoola, line 266: `phone2` is missing here

server/services/user.service.js
```diff
@@ -261,24 +261,64 @@ export const updateUser = async (updatedUser) => {
...
+  const booleanKeys = ['activated'];
+  const vendorBooleanKeys = ['verified', 'certified'];
+  const vendorStringKeys = ['companyName'];
```
andela-hpopoola, line 269: `entity` and `redanNumber` are missing here

server/services/user.service.js
```diff
@@ -261,24 +261,64 @@ export const updateUser = async (updatedUser) => {
...
+
```
andela-hpopoola, line 270: Address filters are missing here.

andela-hpopoola, line 270: We are supposed to add date filters but it is more complicated. Maybe when this has finally taken shape.

server/services/user.service.js
```diff
@@ -261,24 +261,64 @@ export const updateUser = async (updatedUser) => {
...
+  stringKeys.forEach((key) => {
+    if (query[key]) {
+      filterQuery.push({ [key]: { $regex: query[key], $options: 'i' } });
```
andela-hpopoola, line 273: The next step is to create a generic function to generate for each datatype. I will explain better during standup.

test/services/user.service.test.js
andela-hpopoola: Can you please fix this before resolving or let me know the challenges involved that is preventing the fix?

test/controllers/user.controller.test.js
```diff
@@ -3104,37 +3129,30 @@ describe('User Controller', () => {
...
-        user: adminUser,
-        userId: adminUser._id,
+    context('when role is passed', () => {
+      beforeEach(async () => {
+        adminToken = await addUser(adminUser);
       });
+      [USER_ROLE.VENDOR, USER_ROLE.ADMIN, USER_ROLE.EDITOR].map((user) =>
+        it('returns the error', (done) => {
+          request()
+            [method](`${endpoint}?role=${user}`)
```
andela-hpopoola, line 3140: Nice work here, however, we need to write tests for all the filters used in the code to ensure that they work as expected.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments and would also explain more concepts that could be used to simplify the filter process during standup.

test/services/user.service.test.js
tobieeniafe: the object can be empty, the function was only adjusted due to the switch from optional parameters `page` & `limit` to an object

server/helpers/filters.js
```diff
@@ -0,0 +1,38 @@
+export const filterStringKeys = (filterQuery, stringArray, query, prefix = '') => {
+  let pre = '';
```
andela-hpopoola, line 2: The 4 lines can be converted into 1 when a ternary operator is used.

```js
  const pre = prefix.length > 0 ? `${prefix}.` : '';
```

server/helpers/filters.js
```diff
@@ -0,0 +1,38 @@
...
+  if (prefix.length > 0) {
+    pre = `${prefix}.`;
+  }
+
+  stringArray.forEach((key) => {
+    if (query[key]) {
+      filterQuery.push({ [`${pre}${key}`]: { $regex: query[key], $options: 'i' } });
```
andela-hpopoola, line 9: It is bad practice to mutate variables outside a function. It is better to return the result so that it can be used as needed.

server/helpers/filters.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+  });
+};
+
+export const filterIntegerKeys = (filterQuery, integerArray, query, prefix = '') => {
```
andela-hpopoola, line 14: These functions have lots of repetition. I think we can compress them into one. Here is the update I came up with for the filter.js helper

```js
export const FILTER_TYPE = {
  BOOLEAN: 'boolean',
  INTEGER: 'integer',
  STRING: 'string',
};

export const buildFilterQuery = (stringArray, query, type, prefix = '') => {
  const pre = prefix.length > 0 ? `${prefix}.` : '';

  return stringArray.reduce((acc, key) => {
    if (query[key]) {
      switch (type) {
        case FILTER_TYPE.BOOLEAN:
          acc.push({ [`${pre}${key}`]: query[key] === 'true' });
          break;

        case FILTER_TYPE.INTEGER:
          acc.push({ [`${pre}${key}`]: parseInt(query[key], 10) });
          break;

        case FILTER_TYPE.STRING:
          acc.push({ [`${pre}${key}`]: { $regex: query[key], $options: 'i' } });
          break;

        default:
          break;
      }
    }
    return acc;
  }, []);
};


```

server/services/user.service.js
```diff
@@ -261,24 +262,38 @@ export const updateUser = async (updatedUser) => {
...
+export const getAllUsers = async ({ page = 1, limit = 10, ...query }) => {
+  const filterQuery = [];
+  const stringKeys = ['firstName', 'lastName', 'email', 'phone', 'phone2', 'referralCode'];
+  const integerKeys = ['role'];
+  const booleanKeys = ['activated'];
+  const vendorBooleanKeys = ['verified', 'certified'];
+  const vendorStringKeys = ['companyName', 'entity', 'redanNumber'];
+  const addressStringkeys = ['city', 'country', 'state', 'street1', 'street2'];
+
+  filterStringKeys(filterQuery, stringKeys, query);
```
andela-hpopoola, line 274: For the above  comment, instead of mutating the filterQuery variable, we can get the build it with the function above:

```js
  const filterQuery = [
    ...buildFilterQuery(booleanKeys, query, FILTER_TYPE.BOOLEAN),
    ...buildFilterQuery(integerKeys, query, FILTER_TYPE.INTEGER),
    ...buildFilterQuery(stringKeys, query, FILTER_TYPE.STRING),
    ...buildFilterQuery(addressStringkeys, query, FILTER_TYPE.STRING, 'address'),
    ...buildFilterQuery(vendorBooleanKeys, query, FILTER_TYPE.BOOLEAN, 'vendor'),
    ...buildFilterQuery(vendorStringKeys, query, FILTER_TYPE.STRING, 'vendor'),
  ];
```

Even though they look similar, this will be easier to debug since the function is a pure function. When you are chanced, you can try to read more on mutations and pure functions.

test/controllers/user.controller.test.js
```diff
@@ -3117,36 +3159,50 @@ describe('User Controller', () => {
...
+    });
+    context('when multiple filters are used', () => {
+      it('returns matched user', (done) => {
+        request()
+          [method](
+            `${endpoint}?firstName=${dummyVendor1.firstName}&role=${dummyVendor1.role}&activated=${dummyVendor1.activated}&verified=${dummyVendor1.vendor.verified}&companyName=${dummyVendor1.vendor.companyName}&country=${dummyVendor1.address.country}`,
+          )
+          .set('authorization', adminToken)
+          .end((err, res) => {
+            expectsPaginationToReturnTheRightValues(res, {
```
andela-hpopoola, line 3195: Tests for single filters are still missing.

andela-hpopoola (changes requested):
Thank you for working on this. This is really taking shape and we are almost through with the implementation. Apart from the single tests, every other thing looks legit and almost done. Nice work.

test/services/user.service.test.js
tobieeniafe: the function takes in an object parameter, which contains two default values `page` & `limit`. Calling the function without the object in it makes `page` & `limit` undefined. Thereby causing an error.

server/helpers/filters.js
```diff
@@ -1,38 +1,31 @@
...
+  INTEGER: 'integer',
+  STRING: 'string',
 };
 
-export const filterIntegerKeys = (filterQuery, integerArray, query, prefix = '') => {
-  let pre = '';
-  if (prefix.length > 0) {
-    pre = `${prefix}.`;
-  }
+export const buildFilterQuery = (arrayWithKeys, query, type, prefix = '') => {
```
andela-hpopoola, line 7: Lets use `filterKeys` here, `arrayWithKeys` is not descriptive.

```suggestion
export const buildFilterQuery = (filterKeys, query, type, prefix = '') => {
```

test/controllers/user.controller.test.js
```diff
@@ -3184,12 +3192,19 @@ describe('User Controller', () => {
         }),
       );
     });
+
     context('when multiple filters are used', () => {
+      const filteredUrl = JSON.stringify(dummyVendor1Details)
```
andela-hpopoola, line 3197: I think it is better to use the queryString to generate the filter params. Also, the name should be filteredParams as it contains a list of parameters not the entire URL.

```suggestion
      const filteredUrl = querystring.stringify(dummyVendor1Details);
```

You need to import querystring at the top of the test file to use the above

```js
import querystring from 'querystring';
```

Please take a minute to go through it in the documentation since it contains lots of valuable information to cater for query strings.

https://nodejs.org/api/querystring.html

andela-hpopoola (changes requested):
Thank you for the updates. We are almost getting there 💪🏽

test/services/user.service.test.js
andela-hpopoola: I think I understand you now but I still dont think you understood my point. You could have fixed this by using a default parameter for getAllUsers.

For example, adding a default parameter here (= {}) should fix the issue:

```js
export const getAllUsers = async ({ page = 1, limit = 10, ...query } = {}) => {
```

andela-hpopoola:
Can we update the title of the PR to reflect the functionality being implemented in it.

```
implement filter functionality for `GET users/all` endpoint
```

test/controllers/user.controller.test.js
```diff
@@ -3220,5 +3232,77 @@ describe('User Controller', () => {
           });
       });
     });
+
+    context('sending single parameter', () => {
```
andela-hpopoola, line 3236: Instead of repeating this statement, you could have grouped similar tests under just one heading.

test/controllers/user.controller.test.js
```diff
@@ -3220,5 +3232,77 @@ describe('User Controller', () => {
...
+      beforeEach(async () => {
+        adminToken = await addUser(adminUser);
+      });
+      Object.keys(filterToReturnOneResult).map((field) =>
+        it('returns matched user', (done) => {
+          request()
+            [method](`${endpoint}?${field}=${filterToReturnOneResult[field]}`)
+            .set('authorization', adminToken)
+            .end((err, res) => {
+              expectsPaginationToReturnTheRightValues(res, {
```
andela-hpopoola, line 3246: I think you will also need to test that the right value is returned. For example, For `firstName` as `vendor1`, it is better to test that the returned object has a firstName of `vendor1`.

test/controllers/user.controller.test.js
```diff
@@ -3220,5 +3232,77 @@ describe('User Controller', () => {
...
+                total: 1,
+                totalPage: 1,
+              });
+              done();
+            });
+        }),
+      );
+    });
+
+    context('sending single parameter', () => {
```
andela-hpopoola, line 3260: I think this should be used be `when sending multiple parameters`.

test/controllers/user.controller.test.js
andela-hpopoola, line 3236: I think it is better to define all the fields of a single user here and use a loop to confirm that the exact field is returned. I noticed that you defined just 2 fields for `filterToReturnOneResult`; there should be no compromise here, we need to define everything used for the filter.

For this part, You will create a user containing all the parts we used in filter and send the request to confirm that it is returned.

Also, it is better for the user to be defined in this block to make it easier to understand to our future selves or anyone going reading the code in the future. Don't forget the code is over 3000 lines, it can be very difficult to find variables especially when it is not written by you.

test/controllers/user.controller.test.js
andela-hpopoola: For this, you can just add two users with the same multiple parameters and use a single test to confirm that the two users are returned. We don't need a loop here since the test for a single parameter already ensures that all parameters work as expected.

Also, apart from pagination, you want to also test that the right values are returned.

test/controllers/user.controller.test.js
```diff
@@ -3220,5 +3232,77 @@ describe('User Controller', () => {
...
+                total: 2,
+                totalPage: 1,
+              });
+              done();
+            });
+        }),
+      );
+    });
+
+    context('sending single parameter', () => {
```
andela-hpopoola, line 3284: This is not needed. The above test already covers this. Instead we should write test for when no parameter is satisfied, it should return an empty result.

test/controllers/user.controller.test.js
```diff
@@ -3142,6 +3145,19 @@ describe('User Controller', () => {
       country: dummyVendor1.address.country,
     };
 
+    const filterToReturnOneResult = {
```
andela-hpopoola, line 3148: Definitions like this are better defined inside their tests assertion so that they are easier to locate.

server/helpers/filters.js
```diff
@@ -4,10 +4,10 @@ export const FILTER_TYPE = {
...
 
-export const buildFilterQuery = (arrayWithKeys, query, type, prefix = '') => {
+export const buildFilterQuery = (filterKeys, query, type, prefix = '') => {
   const pre = prefix.length > 0 ? `${prefix}.` : '';
 
-  return arrayWithKeys.reduce((acc, key) => {
+  return filterKeys.reduce((acc, key) => {
     if (query[key]) {
       switch (type) {
         case FILTER_TYPE.BOOLEAN:
```
andela-hpopoola, line 13: It looks like you have not added the DATE datatype.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few more comments, especially on the tests arrangement and to ensure the tests are actually touching everypart needed for filter so that they will be easier to duplicate in the future.

Nice work and thank you for making adequate efforts to work on the udpates.

server/services/user.service.js
```diff
@@ -269,14 +269,18 @@ export const getAllUsers = async ({ page = 1, limit = 10, ...query }) => {
   const vendorBooleanKeys = ['verified', 'certified'];
   const vendorStringKeys = ['companyName', 'entity', 'redanNumber'];
   const addressStringkeys = ['city', 'country', 'state', 'street1', 'street2'];
+  const dateKeys = ['createdAt'];
+  const vendorDateKeys = ['verifiedOn'];
```
andela-hpopoola, line 274: I think the following are still needed for the filter:
```
- banned status - necessary to search to banned user if an admin wants to unban
- VendorInfo status - necessary to search for pending vendor verification based on some criteria
```

server/services/user.service.js
```diff
@@ -269,14 +269,18 @@ export const getAllUsers = async ({ page = 1, limit = 10, ...query }) => {
...
 
   const filterQuery = [
     ...buildFilterQuery(booleanKeys, query, FILTER_TYPE.BOOLEAN),
     ...buildFilterQuery(integerKeys, query, FILTER_TYPE.INTEGER),
     ...buildFilterQuery(stringKeys, query, FILTER_TYPE.STRING),
+    ...buildFilterQuery(dateKeys, query, FILTER_TYPE.DATE),
```
andela-hpopoola, line 279: Since this is sorted alphabetically, dateKeys should come after booleanKeys. Same with the one below too 😏 .

test/controllers/user.controller.test.js
```diff
@@ -3184,63 +3163,48 @@ describe('User Controller', () => {
...
-            expect(res.body.result[0]._id).to.be.eql(dummyVendor1._id.toString());
-            done();
-          });
-      });
-    });
 
-    context('sending single parameter', () => {
-      beforeEach(async () => {
-        adminToken = await addUser(adminUser);
+      context('sending single parameter', () => {
```
andela-hpopoola, line 3180: I dont think this is needed, not sure why we need to separate or test role only here when we have a test that loops through every single parameter.

test/controllers/user.controller.test.js
```diff
@@ -3251,20 +3215,22 @@ describe('User Controller', () => {
                 total: 1,
                 totalPage: 1,
               });
+              expect(res.body.result[0]._id).to.be.eql(dummyVendor1._id.toString());
```
andela-hpopoola, line 3218: We also want to add tests to ensure the filtered parameters are returned.

test/controllers/user.controller.test.js
```diff
@@ -3275,34 +3241,69 @@ describe('User Controller', () => {
...
+
+      context('sending single parameter', () => {
+        const singleParameterFilers = {
+          firstName: dummyVendor1.firstName,
+          role: dummyVendor1.role,
+          activated: dummyVendor1.activated,
+        };
+
+        Object.keys(singleParameterFilers).map((field) =>
+          it('returns matched users', (done) => {
```
andela-hpopoola, line 3294: You also need to test that it is only one result that is returned.

test/controllers/user.controller.test.js
```diff
@@ -3275,34 +3241,69 @@ describe('User Controller', () => {
...
               });
               done();
             });
-        }),
-      );
+        });
+      });
+
+      context('sending single parameter', () => {
+        const singleParameterFilers = {
```
andela-hpopoola, line 3287: This should contain all the properties in the filter parameter as recommended in previous comments.

All of this

```js
  const stringKeys = ['firstName', 'lastName', 'email', 'phone', 'phone2', 'referralCode'];
  const integerKeys = ['role'];
  const booleanKeys = ['activated'];
  const vendorBooleanKeys = ['verified', 'certified'];
  const vendorStringKeys = ['companyName', 'entity', 'redanNumber'];
  const addressStringkeys = ['city', 'country', 'state', 'street1', 'street2'];
```

You can define all the values in an Object and use the Object to create the user.

andela-hpopoola (changes requested):
Thank you for working on this. The tests still need some perfection to increase our confidence level, so I dropped a few comments on the best way to go about it. Thank you.

andela-hpopoola:
To be more explicit, the test format should satisfy the following:

- When no filter is used, it returns all users in the database (necessary to ensure that no filter is not blocking the gell all users service). This has already been implemented in the `User pagination` test case
- When an unknown filter is used, it still returns all users in the database
- When a known filter with no result is used, it returns 0 user
- When a single filter is used, it returns the user that satisfied the filter. To ease this, we can use one user with predefined filters and loop over each filter. This is the most important part of the test as this will test all the filters we have used to ensure they are working as expected
- When a single filter is used that satisfies multiple users, it returns the multiple users. The test for the roles satisfies this, we only need to properly explain it in the test.
- When multiple filters are used, it returns the users that satisfy the multiple filters are returned.

Note that I used `When` to start all of my statements above to denote the `context` block. Similar contexts should be grouped together. Thank you.

server/services/user.service.js
tobieeniafe: adding banned would require us to change the user model from `banned.status` to `banned.banned` to make it usable in the filter function as passing `status` alone as a parameter isn't descriptive

andela-hpopoola: Changing the user model because of filter is not a good idea.  Are you going to change for other models too because they don't conform to the function we have written. We can always improve the function to cater for this. 

For now, lets treat `banned` as a special variable, so we don't need to include it among the keys used for the function but create another function to generate its own filter. 

Also, we might need to change the filter function to support this type of parameter in the future. Maybe when the PR sounds good to go, I will rewrite the function to cater for cases like this. 

For now, change the `filterQuery` variable to `let` instead of `const`, after generating all the filter queries with the `buildFilterQuery` functions, check if `banned` is included in `req.query`, and push the generated `banned` filter query into the `filterQuery` variable.

test/controllers/user.controller.test.js
```diff
@@ -3113,9 +3127,11 @@ describe('User Controller', () => {
         firstName: 'vendor1',
         address: { country: 'Ghana' },
         vendor: {
-          companyName: 'Dangote PLC',
+          companyName: 'dangote plc',
```
andela-hpopoola, line 3130: Not sure why this is changed but can we use the proper casing 🙏🏽

test/controllers/user.controller.test.js
```diff
@@ -3125,9 +3141,9 @@ describe('User Controller', () => {
       activated: true,
       address: { country: 'Nigeria' },
       vendor: {
-        companyName: 'Google Nigeria',
-        verified: true,
-        certified: true,
+        companyName: 'google nigeria',
```
andela-hpopoola, line 3144: Same comment as above.

test/controllers/user.controller.test.js
```diff
@@ -3216,15 +3231,23 @@ describe('User Controller', () => {
...
             });
         });
       });
 
-      context('sending multiple parameters', () => {
+      context('when multiple filters are used', () => {
         const filterToReturnTwoResults = {
           role: 2,
-          verified: true,
+          companyName: 'go',
```
andela-hpopoola, line 3250: Lets leave this as this to ensure that the filter is case insensitive. Remember, the admin doesn't even know the case the used to save the data, so we must ensure that this is case insensitive.

test/controllers/user.controller.test.js
```diff
@@ -3283,22 +3306,137 @@ describe('User Controller', () => {
...
+                total: 18,
+                totalPage: 2,
+              });
+              done();
+            });
+        });
+      });
+
+      context('when sending single parameter', () => {
+        const singleParameterFilters = {
```
andela-hpopoola, line 3334: All of this can be placed in their own context too for easier grouping.

```js
     context('when sending single parameter', () => {
         context('when root parameters are sent', () => {
            ...
         })
         context('when banned parameters are sent', () => {
            ...
         })
         context('when vendor parameters are sent', () => {
            ...
         })
     })

```

andela-hpopoola (changes requested):
Wow. We are 98% there. Thank you for the hard work. This looks good and almost ready to launch. The pending stuff is to rewrite the buildFilterQuery function and ensure that the filter is case insensitive. 💪🏽

test/controllers/user.controller.test.js
```diff
@@ -3331,116 +3345,155 @@ describe('User Controller', () => {
...
+            lastName: dummyEditor[0].lastName,
+            email: dummyEditor[0].email,
+            phone: dummyEditor[0].phone,
+            phone2: dummyEditor[0].phone2,
+            referralCode: dummyEditor[0].referralCode,
+            role: dummyEditor[0].role,
+            activated: dummyVendor1.activated,
+          };
+
+          Object.keys(singleParameterFilters).map((field) =>
```
andela-hpopoola, line 3360: Can we consolidate all of the single tests and use `USER_FILTERS` to filter through all of the filters defined for `Users` as discussed. This has the advantages of ensuring that we don't miss any of the defined filters, makes it easier to update in the future, and makes writing other tests a piece of cake.

test/controllers/user.controller.test.js
```diff
@@ -3331,116 +3345,155 @@ describe('User Controller', () => {
...
-          lastName: dummyEditor[0].lastName,
-          email: dummyEditor[0].email,
-          phone: dummyEditor[0].phone,
-          phone2: dummyEditor[0].phone2,
-          referralCode: dummyEditor[0].referralCode,
-          role: dummyEditor[0].role,
-          activated: dummyVendor1.activated,
-        };
+        context('when root parameters are sent', () => {
+          const singleParameterFilters = {
```
andela-hpopoola, line 3349: Can we use a single `Vendor` user instead of an editor  as discussed on Friday to make using the loop possible?

test/services/user.service.test.js
andela-hpopoola: You have not fixed this too? ☝🏽

andela-hpopoola (changes requested):
Thank you for working on this. We are like 99.5% done. The only thing remaining is to ensure you update the tests to use a single loop for a single filter.

I should approve when this has been done. Thank you and Nice work.

test/controllers/user.controller.test.js
tobieeniafe: I don't understand?

andela-hpopoola: We discussed this on Friday. We agreed to use just one loop for the single filters test. We discussed removing one vendor from the roles test to enable us to use vendor to test for everything involving user filters (by looping through USER_FILTERS). 

I mentioned on the call that this would enable you to understand how to use the `Object.entries` loop. Apart from the fact that it makes the test easier (as everything will be in one place), the best part is that it forces us to test for all the filters in case we add a new filter in the future and we forget to add a test to it.

USER_FILTERS already has the necessary `keys` and `filter` fields needed for this test, we don't need to split the single test into multiple contexts when we can use just one loop to achieve the same thing in a better way. I hope you understand now.

tobieeniafe: Oh yes, I do. Thanks

tobieeniafe: Could you remind me of the code snippet? Can't seem to remember it ☹️

andela-hpopoola: Sure. The code snippets should be similar to this

https://github.com/Highrachy/ballers-api/blob/9fbca5d3f3e9f060aab0c1dff55e9aa627160c32/server/helpers/filters.js#L11

Here is the algorithm (but using similar variable names from the above link).

1. Get the entries of USER_FILTER
2. Get the `queryKey` from the USER_FILTER entry, and get the **value** of the` queryKey` from the `Vendor` object via the `filterKey`. So the result will be like `[method](`${endpoint}?${queryKey}=${vendorUser[filterKey]}`)`. Not 100% about `vendorUser[filterKey]` though, let me know if you have problems with it.
3. The step above should be used to build the request you are sending to the database
4. Test that the result is equal to Vendor[filterKey] 

I didn't test this, but it should be helpful enough to guide you through the implementation.

tobieeniafe:
I was able to attempt this and make it run at once, but it passes all except the vendor verification steps.

server/helpers/filters.js
```diff
@@ -0,0 +1,80 @@
...
+  street1: { key: 'address.street1', type: FILTER_TYPE.STRING },
+  street2: { key: 'address.street2', type: FILTER_TYPE.STRING },
+};
+
+export const USER_FILTERS = {
+  ...ADDRESS_FILTERS,
+  activated: { type: FILTER_TYPE.BOOLEAN },
+  activationDate: { type: FILTER_TYPE.DATE },
+  banned: { key: 'banned.status', type: FILTER_TYPE.BOOLEAN },
+  bankDetailsStatus: { key: 'vendor.verification.bankDetails.status', type: FILTER_TYPE.STRING },
```
andela-hpopoola, line 58: Updated this and similar queryKeys to `bankDetailsStatus` to make it more explicit and understandable.

test/controllers/user.controller.test.js
```diff
@@ -40,8 +47,14 @@ useDatabase();
 
 let adminToken;
 let userToken;
+const futureDate = add(new Date(), { days: 5 });
```
andela-hpopoola, line 50: Updating other dates to a future date to enable us to filter via today's date. Not 100% sure about the date implementation in the filter though, we can always fix it later.

test/controllers/user.controller.test.js
```diff
@@ -3331,157 +3359,27 @@ describe('User Controller', () => {
...
-                    result: 1,
-                    total: 1,
-                    totalPage: 1,
-                  });
-                  expect(res.body.result[0].vendor.verification[field].status).to.be.eql(
-                    singleVendorVerificationFilters[field],
-                  );
-                  done();
+        Object.entries(USER_FILTERS).map(([queryKey, { key }]) => {
+          const processNestedObject = (parentObject, objPath) =>
```
andela-hpopoola, line 3363: This enables us to read nested property from the object. Necessary for properties like `vendor.companyName`. I didn't place this in a different function because I believe we can make this test a function to test for single properties for other models too.

test/controllers/user.controller.test.js
```diff
@@ -3331,157 +3359,27 @@ describe('User Controller', () => {
...
+            objPath.split('.').reduce((acc, value) => acc[value], parentObject);
+          const filterKey = key || queryKey;
+          return it(`returns matched user for ${queryKey}`, (done) => {
+            request()
+              [method](`${endpoint}?${queryKey}=${processNestedObject(dummyVendor, filterKey)}`)
+              .set('authorization', adminToken)
+              .end((err, res) => {
+                expect(res.body.result[0]._id.toString()).to.be.eql(dummyVendor._id.toString());
```
andela-hpopoola, line 3371: This ensures the right user is returned for all the tests.

andela-hpopoola (approved):
Thank you for your hard work. I have updated the test for the single user and I think everything should be good to go. 

The only thing expected from you is to update the details in the Pull Request description as you didn't explain how the filter will work in the description (you only listed the parameters).

Feel free to merge when ready. 👍🏽

PR #115, opened 2021-02-02
#176735767 add timestamp to log date

tobieeniafe:
#### What does this PR do?
Add timestamp to log date

Add timestamp to log date

#176735767

### DIFF
--- a/server/helpers/dates.js
+++ b/server/helpers/dates.js
@@ -1,7 +1,10 @@
 import format from 'date-fns/format';
+import formatISO from 'date-fns/formatISO';
 
 export const getTodaysDateInWords = () => format(new Date(), 'd MMM yyyy');
 
 export const getTodaysDateShortCode = () => format(new Date(), 'ddMMyyyy');
 
 export const getTodaysDateStandard = () => format(new Date(), 'yyyy-MM-dd');
+
+export const getDateWithTimestamp = () => formatISO(new Date(), 'yyyy-MM-dd');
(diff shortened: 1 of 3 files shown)


server/helpers/dates.js
```diff
@@ -1,7 +1,10 @@
 import format from 'date-fns/format';
+import formatISO from 'date-fns/formatISO';
 
 export const getTodaysDateInWords = () => format(new Date(), 'd MMM yyyy');
 
 export const getTodaysDateShortCode = () => format(new Date(), 'ddMMyyyy');
 
 export const getTodaysDateStandard = () => format(new Date(), 'yyyy-MM-dd');
+
+export const getDateWithTimestamp = () => formatISO(new Date(), 'yyyy-MM-dd');
```
andela-hpopoola, line 10: I don't think 'yyyy-MM-dd' has no effect on this function. 

```suggestion
export const getDateWithTimestamp = () => formatISO(new Date());
```

server/services/user.service.js
```diff
@@ -577,7 +577,7 @@ export const generateLog = ({ updatedFields, updatedVendor, user }) => {
     }
   });
 
-  const updateDate = { updatedAt: getTodaysDateStandard() };
+  const updateDate = { updatedAt: getDateWithTimestamp() };
```
andela-hpopoola, line 580: Even if we are not testing for the exact value (dates requires a special way to test), Let's add a test to ensure that `updatedAt` is always present in this function so that we don't mistakingly distrupt it in the future.

https://github.com/Highrachy/ballers-api/blob/946713b57646b72b6cc3dcb95880298119c93a97/test/controllers/user.controller.test.js#L2187

Something like 

```js
        expect(res.body.user.vendor.logs[0]).to.have.property('updatedAt');
```
should do the trick

andela-hpopoola (approved):
Thank you for working on this. I left a few comments, they are nothing major. Feel free to merge after fixing. Thank you.

PR #117, opened 2021-02-05
#176735760 load additional information when user is viewing a single property

tobieeniafe:
#### What does this PR do?
Return visitation and enquiry when viewing a single property for a user

Return visitation and enquiry when viewing a single property for a user

`GET /api/v1/property/:id`

#176735760

### DIFF
--- a/server/services/property.service.js
+++ b/server/services/property.service.js
@@ -132,8 +132,8 @@ export const getAllProperties = async (user) => {
   return properties;
 };
 
-export const getOneProperty = async (propertyId) =>
-  Property.aggregate([
+export const getOneProperty = async (propertyId, user = {}) => {
+  const propertyOptions = [
     { $match: { _id: ObjectId(propertyId) } },
     {
       $lookup: {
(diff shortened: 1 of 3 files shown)


server/services/property.service.js
```diff
@@ -160,7 +160,67 @@ export const getOneProperty = async (propertyId) =>
         ...PROJECTED_VENDOR_INFO,
       },
     },
-  ]);
+  ];
+
+  if (user && user.role === USER_ROLE.USER) {
```
andela-hpopoola, line 165:
```suggestion
  if (user?.role === USER_ROLE.USER) {
```

andela-hpopoola (changes requested):
Thanks for working on this. The only thing is to write the tests and merge this asap. Thank you.

server/services/property.service.js
andela-hpopoola: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

andela-hpopoola (approved):
Thank you for working on this. This looks good to me. Feel free to merge when the test passes.

PR #118, opened 2021-02-07
#176735770 paginate & implement filter functionality for `GET /api/v1/property/all` endpoint

tobieeniafe:
#### What does this PR do?
Implement filter functionality for `GET property/all` endpoint

-Paginate `GET /api/v1/property/all`
-Implement filter functionality for `GET property/all` endpoint.
-Add `bathrooms` to property model

`GET /api/v1/property/all?filterKey=value`

```
DATE = yyyy-mm-dd
INTEGER = number
OBJECT_ID = valid mongo id
STRING = matching text (case insensitive)
```

Example `/api/v1/property/all?createdAt=2021-01-26&bathrooms=3&houseType=penthouse`

```
addedBy: OBJECT_ID 
bathrooms: INTEGER 
bedrooms: INTEGER 
city: STRING 
country: STRING 
createdAt: DATE 
houseType: STRING 
name: STRING 
price: INTEGER 
toilets: INTEGER 
units: INTEGER 
state: STRING 
street1: STRING 
street2: STRING 
```

#176735770 #176842947


test/controllers/property.controller.test.js
```diff
@@ -1310,50 +1326,30 @@ describe('Property Controller', () => {
...
-              expect(res.body.properties[0]).to.have.property('address');
-              expect(res.body.properties[0]).to.have.property('mainImage');
-              expect(res.body.properties[0]).to.have.property('gallery');
-              expect(res.body.properties[0]).to.have.property('price');
-              expect(res.body.properties[0]).to.have.property('houseType');
-              expect(res.body.properties[0]).to.have.property('description');
-              expectResponseToExcludeSensitiveVendorData(res.body.properties[0].vendorInfo);
-              expectResponseToContainNecessaryVendorData(res.body.properties[0].vendorInfo);
+              expectsPaginationToReturnTheRightValues(res, defaultPaginationResult);
+              expectResponseToContainNecessaryPropertyData(res.body.result[0]);
```
andela-hpopoola, line 1330: Nice one here 🙌🏽

andela-hpopoola (changes requested):
Thank you for working on this. I did a quick look and this looks good to me. Waiting for the filter so that I can approve asap. Nice work.

tobieeniafe:
@andela-hpopoola the property filter seems to be a little bit more complex compared to the user filter. I'm considering deprecating `/api/v1/property/added-by/:id` and `/api/v1/property/search` endpoints, since this endpoint would be able to cater for them all. But I'm also considering user access level rights. Eg a vendor using `/api/v1/property/all?addedBy=anotherVendorId` and other related cases. Thoughts?

andela-hpopoola:
Nice observations. Thank you for raising this. 

> I'm considering deprecating /api/v1/property/added-by/:id and /api/v1/property/search endpoints, since this endpoint would be able to cater for them all

Let's do this in another PR when we have sorted out the filter functionality

>  But I'm also considering user access level right issues. Eg a vendor or a user using /api/v1/property/all?addedBy=anotherVendorId and other related cases. Thoughts?

I don't think this should be a major issue. We will define different Filters constant based on the access level. That means we will have 3 filters constant to cater to the different access level. Even if a vendor tries use `addedBy`, as long as it is not defined for the vendor role in the vendor filter constant, it will be naturally ignored and will give us no security issues. We should also add tests to verify that this works as expected after implementation.

Thank you for raising this and asking 👍🏽

tobieeniafe:
@andela-hpopoola the test for `createdAt` seems to be failing as it's generating a new property `id` can't seem to figure out why. Perhaps a fresh set of eyes might. Also, I added the `bathrooms` to the property model in this PR as I needed to test it with the filter.

andela-hpopoola (changes requested):
Thank you for working on this. As you have noticed, all our hard work paid off for filters and the filter functionality looks fun to work with 😉.

I left a few comments on this to start with. We still have some edge cases we need to consider but we can do those ones gradually.


Next time, I think it would be better to do the bathroom inclusion in a later PR. We can merge that PR first while you include the filter implementation here. You are able to get away with it, but it might make the PR more complex in the future.

We shouldn't deprecate user search for the `GET /property/all` as the endpoint is not targeted for users. Also, looking through the search service, the implementation is tailored towards a more customized search (involving areas, price, and others) which we didn't cover here.

I believe that we should be able to deprecate the `addedBy` vendor endpoint in a later PR.

Also, we need to implement a functionality to cater for greater than, less than, and equal to for some query parameters.

test/controllers/property.controller.test.js
```diff
@@ -1270,151 +1281,302 @@ describe('Property Controller', () => {
...
+          street1: 'miracle street',
+          street2: 'miracle street',
+          city: 'ilorin',
+          state: 'kwara',
+          country: 'ghana',
+        },
+        addedBy: vendorUser._id,
+        bathrooms: 1,
+        bedrooms: 1,
+        createdAt: futureDate,
```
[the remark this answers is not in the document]
tobieeniafe: the issue is the format of `futureDate`. The endpoint expects the date in `yyyy-mm-dd` format.  `futureDate` returns the date in words causing an invalid date on the endpoint.

test/controllers/property.controller.test.js
```diff
@@ -1270,151 +1281,302 @@ describe('Property Controller', () => {
...
+                });
+                done();
+              });
+          });
+        });
+
+        itReturnsErrorForUnverifiedVendor({
+          endpoint,
+          method,
+          user: vendorUser,
```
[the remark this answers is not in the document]
tobieeniafe: the vendor becomes unverified in the helper file. (see `test/helpers.js` line `346`

server/services/property.service.js
```diff
@@ -121,15 +126,28 @@ export const getAllProperties = async (user) => {
...
+      },
+    },
   ];
 
+  if (filterQuery.length < 1) {
+    propertiesOptions.shift();
+  }
+
   if (user.role === USER_ROLE.VENDOR) {
     propertiesOptions.unshift({ $match: { addedBy: ObjectId(user._id) } });
```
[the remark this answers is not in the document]
tobieeniafe: Tests to cover this already exist. see `property.controller.test.js` lines `1501` & `1518`

test/helpers.js
```diff
@@ -386,14 +386,66 @@ export const expectResponseToContainNecessaryVendorData = (data) => {
...
 
-export const expectResponseToContainNecessaryPropertyData = (data) => {
-  expect(data).to.have.property('name');
-  expect(data).to.have.property('address');
-  expect(data).to.have.property('mainImage');
-  expect(data).to.have.property('gallery');
-  expect(data).to.have.property('price');
-  expect(data).to.have.property('houseType');
-  expect(data).to.have.property('description');
+export const expectResponseToContainNecessaryPropertyData = (data, object) => {
```
andela-hpopoola, line 389: With the current naming, it is difficult to differentiate the response data from the actual property data.
```suggestion
export const expectResponseToContainNecessaryPropertyData = (response, property) => {
```

test/controllers/property.controller.test.js
```diff
@@ -1292,24 +1324,26 @@ describe('Property Controller', () => {
...
+          street2: 'sesame street',
           city: 'ilorin',
           state: 'kwara',
           country: 'ghana',
         },
         addedBy: vendorUser._id,
         bathrooms: 1,
         bedrooms: 1,
-        createdAt: futureDate,
+        createdAt: '2021-12-21',
```
andela-hpopoola, line 1346: Can we revert this to `futureDate` to ensure that it works as expected. If the problem is with future date, then it is better to fix it than having to use a workaround.

test/helpers.js
```diff
@@ -386,14 +386,66 @@ export const expectResponseToContainNecessaryVendorData = (data) => {
...
+  useExistingUser = false,
+}) => {
+  let token;
+
+  beforeEach(async () => {
+    if (useExistingUser) {
+      const loggedInUser = await loginUser(user);
+      token = loggedInUser.token;
+    } else {
+      token = await addUser({ ...user, vendor: { ...user.vendor, verified: false } });
```
andela-hpopoola, line 423: Not sure why you are making all the users here unverified. If we need an unverified vendor, it is better to pass them into the function than forcing all users to be unverified. 

```suggestion   
   token = await addUser(user)
```

test/controllers/property.controller.test.js
```diff
@@ -1452,7 +1492,7 @@ describe('Property Controller', () => {
         await addProperty(vendorProperty);
       });
 
-      context('when unknown filter is used', () => {
+      context('Unknown Filters', () => {
```
andela-hpopoola, line 1495: This should be a `describe` block with the updates.

```suggestion
	      describe('Unknown Filters', () => {
```

test/controllers/property.controller.test.js
```diff
@@ -1292,24 +1324,26 @@ describe('Property Controller', () => {
     const vendorProperties = PropertyFactory.buildList(12, {
       addedBy: vendorUser._id,
       updatedBy: vendorUser._id,
+      createdAt: new Date(),
```
andela-hpopoola, line 1327: Can we add this to the property factory instead?

andela-hpopoola (changes requested):
Thank you for the hard work on this. Just a few comments, this should be ready to go after the fix.

test/controllers/property.controller.test.js
tobieeniafe: No we can't. Joi would flag it.

andela-hpopoola: Thanks for pointing this out.

andela-hpopoola (approved):
This looks good to go. Thank you for ensuring that this works as expected.

PR #119, opened 2021-02-09
#176843012 add more processes for visitation

tobieeniafe:
#### What does this PR do?
Add routes to reschedule and resolve visitations

Add routes to reschedule and resolve visitations

`PUT /api/v1/visitation/cancel`
`PUT /api/v1/visitation/reschedule`
`PUT /api/v1/visitation/resolve`

#176843012

### DIFF
--- a/mailer/index.js
+++ b/mailer/index.js
@@ -60,7 +60,7 @@ const EMAIL_CONTENT = {
   RESCHEDULE_VISIT: {
     subject: 'Visitation Rescheduled',
     contentTop:
-      'Your visitation for property [property name], has been recheduled for [new date]. Visit your dashboard for more information.',
+      'The visitation to [property name] on [previous visitation date], has been rescheduled for [new date] by [user that rescheduled]. [Reason]. Visit your dashboard for more information.',
   },
   RESOLVE_CONCERN: {
     subject: 'Concern Resolved',
--- a/mailer/index.js
+++ b/mailer/index.js
@@ -13,7 +13,7 @@ const EMAIL_CONTENT = {
   CANCEL_VISIT: {
     subject: 'Visitation Cancelled',
     contentTop:
-      'The visitation to [property name] on [previous visitation date], has been cancelled by [user that rescheduled]. [Reason]. Visit your dashboard for more information.',
+      'The visitation to [property name] on [previous visitation date], has been cancelled by [user]. [Reason]. Visit your dashboard for more information.',
   },
   CERTIFY_VENDOR: {
     subject: 'Account Certified',
--- a/mailer/index.js
+++ b/mailer/index.js
@@ -57,6 +57,11 @@ const EMAIL_CONTENT = {
     contentBottom: `Click on the link below to accept their invitation`,
     buttonText: 'Join BALLERS',
   },
+  RESCHEDULE_VISIT: {
+    subject: 'Visitation Rescheduled',
+    contentTop:
+      'Your visitation for property [property name], has been recheduled for [new date]. Visit your dashboard for more information.',
+  },
   RESOLVE_CONCERN: {
     subject: 'Concern Resolved',
     contentTop: `Your raised concern has been resolved. Details below <br /> <strong>Question: </strong> [Question]<br /> <strong>Response: </strong>[Response]`,
@@ -69,6 +74,10 @@ const EMAIL_CONTENT = {
       "You (or someone pretending to be you) requested a password reset for your account. If you didn't made this request you can ignore this email.",
     subject: 'Password Reset',
   },
+  SCHEDULE_VISIT: {
+    subject: 'You have a new inspection request',
+    contentTop: 'A new propery visit has been requested',
+  },
   UNBAN_USER: {
     subject: 'Account Unlocked',
     contentTop:
@@ -78,10 +87,6 @@ const EMAIL_CONTENT = {
     subject: 'Account Verified',
     contentTop:
       'Congratulaions, your information has been reviewed and your account verified. Your dashboard is now fully accessible.',
-  },
-  SCHEDULE_VISIT: {
-    subject: 'You have a new inspection request',
-    contentTop: 'A new propery visit has been requested',
   },
   WELCOME_MESSAGE: {
     subject: 'Welcome to BALLERS',
--- a/mailer/index.js
+++ b/mailer/index.js
@@ -10,6 +10,11 @@ const EMAIL_CONTENT = {
     contentTop:
       'Your account has been locked for security reasons. Kindly contact Ballers Support for more information.',
   },
+  CANCEL_VISIT: {
+    subject: 'Visitation Cancelled',
+    contentTop:
+      'The visitation to [property name] on [previous visitation date], has been cancelled by [user that rescheduled]. [Reason]. Visit your dashboard for more information.',
+  },
   CERTIFY_VENDOR: {
     subject: 'Account Certified',
     contentTop:
--- a/server/controllers/visitation.controllers.js
+++ b/server/controllers/visitation.controllers.js
@@ -3,6 +3,7 @@ import {
   getAllVisitations,
   rescheduleVisitation,
   resolveVisitation,
+  cancelVisitation,
 } from '../services/visitation.service';
 import EMAIL_CONTENT from '../../mailer';
 import { sendMail } from '../services/mailer.service';
@@ -68,6 +69,21 @@ const VisitationController = {
         res
           .status(httpStatus.OK)
           .json({ success: true, message: 'Visitation rescheduled', visitation });
+      })
+      .catch((error) => next(error));
+  },
+
+  cancelVisitation(req, res, next) {
+    const visitationInfo = req.locals;
+    const { user } = req;
+    cancelVisitation({ visitationInfo, userId: user._id })
+      .then(({ visitation, mailDetails, property }) => {
+        const contentTop = `The visitation to ${property.name} on ${visitation.visitDate}, has been cancelled by ${user.lastName} ${user.firstName}. Reason: ${visitation.rescheduleLog[0].reason}. Visit your dashboard for more information.`;
+
+        sendMail(EMAIL_CONTENT.CANCEL_VISIT, mailDetails, { contentTop });
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Visitation cancelled', visitation });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/visitation.controllers.js
+++ b/server/controllers/visitation.controllers.js
@@ -7,7 +7,9 @@ import {
 import EMAIL_CONTENT from '../../mailer';
 import { sendMail } from '../services/mailer.service';
 import httpStatus from '../helpers/httpStatus';
-import { convertDateToWords } from '../helpers/dates';
+import { convertDateToLongHumanFormat } from '../helpers/dates';
+import getUserName from '../helpers/funtions';
+import { PROCESS_VISITATION_ACTION } from '../helpers/constants';
 
 const VisitationController = {
   book(req, res, next) {
@@ -55,14 +57,14 @@ const VisitationController = {
   rescheduleVisitation(req, res, next) {
     const visitationInfo = req.locals;
     const { user } = req;
-    const action = 'reschedule';
+    const action = PROCESS_VISITATION_ACTION.RESCHEDULE;
     processVisitation({ user, visitationInfo, action })
       .then(({ visitation, mailDetails, property }) => {
-        const contentTop = `The visitation to ${property.name} on ${convertDateToWords(
+        const contentTop = `The visitation to ${property.name} on ${convertDateToLongHumanFormat(
           visitation.rescheduleLog[0].rescheduleFrom,
-        )}, has been rescheduled for ${convertDateToWords(visitation.visitDate)} by ${
-          user.vendor.companyName ? user.vendor.companyName : `${user.firstName} ${user.lastName}`
-        }. Reason: ${
+        )}, has been rescheduled for ${convertDateToLongHumanFormat(
+          visitation.visitDate,
+        )} by ${getUserName(user)}. Reason: ${
           visitation.rescheduleLog[0].reason
         }. Visit your dashboard for more information.`;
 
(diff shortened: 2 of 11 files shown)


mailer/index.js
```diff
@@ -57,6 +57,11 @@ const EMAIL_CONTENT = {
     contentBottom: `Click on the link below to accept their invitation`,
     buttonText: 'Join BALLERS',
   },
+  RESCHEDULE_VISIT: {
+    subject: 'Visitation Rescheduled',
+    contentTop:
+      'Your visitation for property [property name], has been recheduled for [new date]. Visit your dashboard for more information.',
```
andela-hpopoola, line 63: The visitation to [property name] on [previous visitation date], has been rescheduled for [new date] by [user that rescheduled]. [Reason]. Visit your dashboard for more information.

server/controllers/visitation.controllers.js
```diff
@@ -33,6 +38,32 @@ const VisitationController = {
...
+      })
+      .catch((error) => next(error));
+  },
+
+  rescheduleVisitation(req, res, next) {
+    const visitationInfo = req.locals;
+    const { user } = req;
+    rescheduleVisitation({ user, visitationInfo })
+      .then(({ visitation, response }) => {
+        const contentTop = `Your visitation for property ${response.property.name}, has been recheduled for ${visitation.visitDate}. Visit your dashboard for more information.`;
```
andela-hpopoola, line 59: I have updated this, so you might want to include the previous date to make it more clear to the user.

server/services/visitation.service.js
```diff
@@ -59,3 +61,64 @@ export const getAllVisitations = async (user, page = 1, limit = 10) => {
...
+  }
+
+  if (user.role === USER_ROLE.VENDOR) {
+    mailDetails = await getUserById(visitation.userId);
+  } else {
+    mailDetails = await getUserById(visitation.vendorId);
+  }
+  const property = await getPropertyById(visitation.propertyId);
+
+  const response = { mailDetails, property };
```
andela-hpopoola, line 111: This is not needed since they are not reused, it is easier to read if we write the output out directly instead of assigning them to response.

server/services/visitation.service.js
```diff
@@ -59,3 +61,64 @@ export const getAllVisitations = async (user, page = 1, limit = 10) => {
...
+
+  try {
+    const rescheduledVisitation = await Visitation.findByIdAndUpdate(
+      visitation.id,
+      { $set: { visitDate: visitationInfo.visitDate } },
+      { new: true },
+    );
+
+    return { visitation: rescheduledVisitation, response };
```
andela-hpopoola, line 120:
```suggestion
    return { visitation: rescheduledVisitation, mailDetails, property };
```

server/controllers/visitation.controllers.js
```diff
@@ -33,6 +38,32 @@ const VisitationController = {
...
+          .json({ success: true, message: 'Visitation resolved', visitation });
+      })
+      .catch((error) => next(error));
+  },
+
+  rescheduleVisitation(req, res, next) {
+    const visitationInfo = req.locals;
+    const { user } = req;
+    rescheduleVisitation({ user, visitationInfo })
+      .then(({ visitation, response }) => {
```
andela-hpopoola, line 58: I don't think we need response here. Can't think of an added advantage of using it over getting the variables directly 🤸🏽‍♂️

server/services/visitation.service.js
```diff
@@ -59,3 +61,64 @@ export const getAllVisitations = async (user, page = 1, limit = 10) => {
...
+      visitation.id,
+      { $set: { status: VISITATION_STATUS.RESOLVED } },
+      { new: true },
+    );
+  } catch (error) {
+    throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error resolving visitation', error);
+  }
+};
+
+export const rescheduleVisitation = async ({ user, visitationInfo }) => {
```
andela-hpopoola, line 88: Let's add a reschedule reason here to make the other party know why the reschedule was necessary. This might not look necessary but from experience, you will want to keep as much information as possible for cases when users complain a vendor is not around or a vendor complains that a user is not serious. This will allow the admin to know the real fact.

This should be an array that contains the following:

```
reschedule : [{
  reason: ..., // the reason the reschedule was made
  rescheduleFrom: .. // previous date in the db
  rescheduleTo: // new date the reschedule moved to
  date: ...,  // date the reschedule was made. Should be datetime format
  rescheduleBy: // String => Vendor / User. (we can get their ID from the visitation id, so no need to use an id here.
}]
```

andela-hpopoola (changes requested):
Thank you for working on this. The implementation looks solid and you have made lots of thoughts in making this as concise as possible. Thank you for the awesome job.

However, my evil twin decided to make it a little bit more complex 👻  to ensure we have logs of all rescheduled that has been made. In real life scenerio, you never can tell when this will be handy, it is better to have them then to ignore them.

Also, after implementation and adding tests, I believe this should  be good to go. Can't really see a big blocker on this. Thank you and Nice work.

test/controllers/visitation.controller.test.js
```diff
@@ -422,4 +423,223 @@ describe('Visitation Controller', () => {
...
+        expect(res.body.success).to.be.eql(true);
+        expect(res.body.message).to.be.eql('Visitation rescheduled');
+        expect(res.body.visitation._id).to.be.eql(visitation._id.toString());
+        expect(res.body.visitation.rescheduleLog[0].reason).to.be.eql(body.reason);
+        expect(res.body.visitation.rescheduleLog[0].rescheduleTo).to.have.string(body.visitDate);
+        expect(res.body.visitation.rescheduleLog[0].rescheduleFrom).to.have.string(
+          visitation.visitDate,
+        );
+        expect(sendMailStub.callCount).to.eq(1);
+        expect(sendMailStub).to.have.be.calledWith(EMAIL_CONTENT.RESCHEDULE_VISIT);
```
andela-hpopoola, line 544: Can we test that the mail is sent to the right person - user or vendor. We don't want to mix this up, so it is better to have a test to ensure this sends the mail to the right person.

andela-hpopoola (changes requested):
Thank you for working on this. This looks very good and thoughtful. 

However, when reviewing the code, I noticed that it might be better to allow a user to cancel a visitation too so that the reason can be stated in the rescheduleLog (not sure if we should give the vendor the same power so you might have to confirm this with Nnamdi 😜 ). Thank you.

test/controllers/visitation.controller.test.js
tobieeniafe: How do you propose we go about this? Adding the `vendor` or `user` object throws an error.

server/services/visitation.service.js
```diff
@@ -139,3 +146,48 @@ export const rescheduleVisitation = async ({ user, visitationInfo }) => {
...
+  }
+
+  if (visitation.status === VISITATION_STATUS.RESOLVED) {
+    throw new ErrorHandler(
+      httpStatus.PRECONDITION_FAILED,
+      'The visitation has already been resolved',
+    );
+  }
+
+  const mailDetails = await getUserById(visitation.vendorId);
```
andela-hpopoola, line 169: Lets update this to `vendorDetails` or something similar

mailer/index.js
```diff
@@ -10,6 +10,11 @@ const EMAIL_CONTENT = {
     contentTop:
       'Your account has been locked for security reasons. Kindly contact Ballers Support for more information.',
   },
+  CANCEL_VISIT: {
+    subject: 'Visitation Cancelled',
+    contentTop:
+      'The visitation to [property name] on [previous visitation date], has been cancelled by [user that rescheduled]. [Reason]. Visit your dashboard for more information.',
```
andela-hpopoola, line 16:
```suggestion
      'The visitation to [property name] on [previous visitation date], has been cancelled by [user]. [Reason]. Visit your dashboard for more information.',
```

server/controllers/visitation.controllers.js
```diff
@@ -71,6 +72,21 @@ const VisitationController = {
       })
       .catch((error) => next(error));
   },
+
+  cancelVisitation(req, res, next) {
+    const visitationInfo = req.locals;
+    const { user } = req;
+    cancelVisitation({ visitationInfo, userId: user._id })
+      .then(({ visitation, mailDetails, property }) => {
+        const contentTop = `The visitation to ${property.name} on ${visitation.visitDate}, has been cancelled by ${user.lastName} ${user.firstName}. Reason: ${visitation.rescheduleLog[0].reason}. Visit your dashboard for more information.`;
```
andela-hpopoola, line 81: Not sure if we should use the `LastName FirstName` naming convention. In the international market, it is always presented as `FirstName LastName`. Maybe we should have an helper function to generate someone's full name so that it will be easier to update irrespective of the format.

```suggestion
        const contentTop = `The visitation to ${property.name} on ${visitation.visitDate}, has been cancelled by ${user.firstName} ${user.lastName}. Reason: ${visitation.rescheduleLog[0].reason}. Visit your dashboard for more information.`;
```

test/controllers/visitation.controller.test.js
```diff
@@ -590,7 +586,28 @@ describe('Visitation Controller', () => {
             .end((err, res) => {
               expect(res).to.have.status(412);
               expect(res.body.success).to.be.eql(false);
-              expect(res.body.message).to.be.eql('You cannot reschedule a resolved visitation');
+              expect(res.body.message).to.be.eql('The visitation has already been resolved');
```
andela-hpopoola, line 589: Can we add `mailStubCount` to these tests to be sure that the mails are not sent if there are errors?

server/controllers/visitation.controllers.js
```diff
@@ -33,6 +39,54 @@ const VisitationController = {
...
+      .catch((error) => next(error));
+  },
+
+  rescheduleVisitation(req, res, next) {
+    const visitationInfo = req.locals;
+    const { user } = req;
+    rescheduleVisitation({ user, visitationInfo })
+      .then(({ visitation, mailDetails, property }) => {
+        const contentTop = `The visitation to ${property.name} on ${
+          visitation.rescheduleLog[0].rescheduleFrom
```
andela-hpopoola, line 61: This should be converted into a human-readable date.

server/services/visitation.service.js
```diff
@@ -59,3 +62,132 @@ export const getAllVisitations = async (user, page = 1, limit = 10) => {
...
+      { new: true },
+    );
+
+    return { visitation: rescheduledVisitation, mailDetails, property };
+  } catch (error) {
+    throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error rescheduling visitation', error);
+  }
+};
+
+export const cancelVisitation = async ({ userId, visitationInfo }) => {
```
andela-hpopoola, line 150: The` rescheduleVisitation` service and the `cancelVisitation` service have so much in common that I feel it is better to combine them together into one function like `processVisitation` service that takes an action. They should have their different endpoints to ensure that `cancelVisitation` is only accessible by the user.

```js
export const processVisitation = async ({ userId, visitationInfo, action }) => {
```
`action` would be used to determine if the service should set the log as RESCHEDULED OR CANCELLED.

server/services/visitation.service.js
```diff
@@ -59,3 +62,132 @@ export const getAllVisitations = async (user, page = 1, limit = 10) => {
...
+  }
+};
+
+export const rescheduleVisitation = async ({ user, visitationInfo }) => {
+  let mailDetails;
+  const visitation = await getVisitationById(visitationInfo.visitationId).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (!visitation) {
```
andela-hpopoola, line 95: Not too important (for improved coverage), can we add a test for this?

andela-hpopoola (changes requested):
Thank you for working on this and making the necessary updates. I left a few comments, the most important is combining the services for `cancelVisitation` and `rescheduleVisitation`. Once this has been fixed, we should be able to merge this with no issues. Thank you.

test/controllers/visitation.controller.test.js
andela-hpopoola: Feel free to ignore, currently too busy to troubleshoot this now.

server/controllers/visitation.controllers.js
```diff
@@ -55,12 +55,13 @@ const VisitationController = {
...
+    processVisitation({ user, visitationInfo, action })
       .then(({ visitation, mailDetails, property }) => {
-        const contentTop = `The visitation to ${property.name} on ${
-          visitation.rescheduleLog[0].rescheduleFrom
-        }, has been rescheduled for ${visitation.visitDate} by ${
-          user.vendor.companyName ? user.vendor.companyName : `${user.lastName} ${user.firstName}`
+        const contentTop = `The visitation to ${property.name} on ${convertDateToWords(
+          visitation.rescheduleLog[0].rescheduleFrom,
+        )}, has been rescheduled for ${convertDateToWords(visitation.visitDate)} by ${
+          user.vendor.companyName ? user.vendor.companyName : `${user.firstName} ${user.lastName}`
```
andela-hpopoola, line 64: Lets create a function for this so that it would be easier to update the name format in the future. 

```suggestion
          user.vendor.companyName ? user.vendor.companyName : `${getUserName(user)}`
```

server/services/visitation.service.js
```diff
@@ -126,68 +127,33 @@ export const rescheduleVisitation = async ({ user, visitationInfo }) => {
   const rescheduleLog = [
     {
       reason: visitationInfo.reason,
-      rescheduleFrom: visitation.visitDate,
-      rescheduleTo: visitationInfo.visitDate,
+      rescheduleFrom: action === 'reschedule' ? visitation.visitDate : null,
```
andela-hpopoola, line 130: Lets define the actions in a constant to make it consistent and save us from unintended bug 🙏🏽 

```js
const PROCESS_VISITATION_ACTION = {
   CANCEL: 0,
   RESCHEDULE: 1,
}
```

server/helpers/dates.js
```diff
@@ -8,3 +8,5 @@ export const getTodaysDateShortCode = () => format(new Date(), 'ddMMyyyy');
 export const getTodaysDateStandard = () => format(new Date(), 'yyyy-MM-dd');
 
 export const getDateWithTimestamp = () => formatISO(new Date());
+
+export const convertDateToWords = (date) => format(date, 'd MMM yyyy');
```
andela-hpopoola, line 12: We should use the Long format in emails so that it is more explicit. The short one can be used for other API messages that require a human-readable format.
```suggestion
export const convertDateToLongHumanFormat  = (date) => format(date, 'dddd, Do MMMM YYYY');
export const convertDateToShortHumanFormat = (date) => format(date, 'ddd, MMM D, YYYY');
```

server/services/visitation.service.js
```diff
@@ -86,8 +86,9 @@ export const resolveVisitation = async ({ visitationId, vendorId }) => {
   }
 };
 
-export const rescheduleVisitation = async ({ user, visitationInfo }) => {
+export const processVisitation = async ({ user, visitationInfo, action }) => {
   let mailDetails;
+  let update;
```
andela-hpopoola, line 91:
```suggestion
  let visitationInfo;
```

andela-hpopoola (approved):
Thank you for your hard work on this. I left a few comments but feel free to merge when you resolve them. Awesome job bro, you deserve a :hi5:.

PR #125, opened 2021-02-16
#176955810 add range to filters

tobieeniafe:
#### What does this PR do?
Add range to filters

Add range to filters

equals `/api/v1/route/all?parameter=value`
to `/api/v1/route/all?parameter=0_:_value`
from `/api/v1/route/all?parameter=value_:_0`
range `/api/v1/route/all?parameter=from_:_to`

#176955810

### DIFF
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -32,11 +32,11 @@ const processRangeSeparator = (key) => {
       return RANGE_TYPE.RANGE;
     }
 
-    if (parseInt(rangeTo, 10) === 0) {
+    if (parseInt(rangeFrom, 10) === 0) {
       return RANGE_TYPE.TO;
     }
 
-    if (parseInt(rangeFrom, 10) === 0) {
+    if (parseInt(rangeTo, 10) === 0) {
       return RANGE_TYPE.FROM;
     }
   }
@@ -120,7 +120,7 @@ export const buildFilterQuery = (allFilters, query) => {
 
               case RANGE_TYPE.TO:
                 integerQueryValue = {
-                  $gte: parseInt(rangeTo, 10),
+                  $lte: parseInt(rangeTo, 10),
                 };
                 break;
 
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -1,4 +1,4 @@
-import { parse, add } from 'date-fns';
+import { add } from 'date-fns';
 import mongoose from 'mongoose';
 
 const { ObjectId } = mongoose.Types.ObjectId;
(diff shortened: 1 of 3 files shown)


server/helpers/filters.js
```diff
@@ -16,22 +16,61 @@ export const buildFilterQuery = (allFilters, query) => {
...
-              $gte: parse(query[queryKey], 'yyyy-MM-dd', new Date()),
+          {
+            let dateQueryValue = {
+              $gte: add(new Date(query[queryKey]), { days: 0 }),
               $lt: add(new Date(query[queryKey]), { days: 1 }),
-            },
-          });
+            };
+
+            if (shouldSort) {
```
andela-hpopoola, line 35: This looks good but it is a little bit difficult to read. I am finding it difficult to know which query works for what and this might be difficult to update in the future:

Here is the rewritten version, please remove comments 🙏🏽 :

```js
import { add } from 'date-fns';
import mongoose from 'mongoose';

const { ObjectId } = mongoose.Types.ObjectId;
// since the date format we support is in YYYY-MM-DD, 
// we can use `:` only and change later if we want to use include time for filters.
const RANGE_SEPARATOR = ':';

export const FILTER_TYPE = {
  BOOLEAN: 'boolean',
  DATE: 'date',
  INTEGER: 'integer',
  OBJECT_ID: 'objectId',
  STRING: 'string',
};

// ensures we know the range type we are dealing with
const RANGE_TYPE = {
  EQUALS: 1,
  TO: 2,
  FROM: 3,
  RANGE: 4,
};

// function to get the type of range from the key
// This makes us know if it is equal, to, from or a range
const processRangeSeparator = (key) => {
  if (key.includes(RANGE_SEPARATOR)) {
    const [rangeFrom = 0, rangeTo = 0] = key.split(RANGE_SEPARATOR);

    if (parseInt(rangeFrom, 10) !== 0 && parseInt(rangeTo, 10) !== 0) {
      return RANGE_TYPE.RANGE;
    }

    if (parseInt(rangeFrom, 10) === 0) {
      return RANGE_TYPE.FROM;
    }
    if (parseInt(rangeTo, 10) === 0) {
      return RANGE_TYPE.TO;
    }
  }

  return RANGE_TYPE.EQUALS;
};

export const buildFilterQuery = (allFilters, query) => {
  return Object.entries(allFilters).reduce((acc, [queryKey, { key, type }]) => {
    const filterKey = key || queryKey;

    if (query[queryKey]) {
      const [rangeFrom = 0, rangeTo = 0] = query[queryKey].split(RANGE_SEPARATOR);

      switch (type) {
        case FILTER_TYPE.BOOLEAN:
          acc.push({ [filterKey]: query[queryKey] === 'true' });
          break;

        case FILTER_TYPE.DATE:
          {
            let dateQueryValue;
            // get the range type here, from the query key
            const rangeType = processRangeSeparator(query[queryKey]);
            
            // I favoured switch here over if since it makes it easier to arrange the code in a logical manner
            switch (rangeType) {
              case RANGE_TYPE.EQUALS: 
                dateQueryValue = {
                  $gte: add(new Date(query[queryKey]), { days: 0 }),
                  $lt: add(new Date(query[queryKey]), { days: 1 }),
                };
                break;

              case RANGE_TYPE.RANGE:
                dateQueryValue = {
                  $gte: add(new Date(rangeFrom), { days: 0 }),
                  $lt: add(new Date(rangeTo), { days: 1 }),
                };
                break;

              case RANGE_TYPE.FROM:
                dateQueryValue = {
                  $gte: add(new Date(rangeFrom), { days: 0 }),
                };
                break;

              case RANGE_TYPE.TO:
                dateQueryValue = {
                  $lt: add(new Date(rangeTo), { days: 1 }),
                };
                break;

              default:
                break;
            }
            acc.push({ [filterKey]: dateQueryValue });
          }
          break;

        case FILTER_TYPE.INTEGER:
          {
            let integerQueryValue;
            const rangeType = processRangeSeparator(query[queryKey]);

            switch (rangeType) {
              case RANGE_TYPE.EQUALS:
                integerQueryValue = parseInt(query[queryKey], 10);
                break;

              case RANGE_TYPE.RANGE:
                integerQueryValue = {
                  $gte: parseInt(rangeFrom, 10),
                  $lte: parseInt(rangeTo, 10),
                };
                break;

              case RANGE_TYPE.FROM:
                integerQueryValue = {
                  $gte: parseInt(rangeFrom, 10),
                };
                break;

              case RANGE_TYPE.TO:
                integerQueryValue = {
                  $gte: parseInt(rangeTo, 10),
                };
                break;

              default:
                break;
            }
            acc.push({ [filterKey]: integerQueryValue });
          }
          break;

        case FILTER_TYPE.OBJECT_ID:
          acc.push({ [filterKey]: ObjectId(query[queryKey]) });
          break;

        case FILTER_TYPE.STRING:
          acc.push({ [filterKey]: { $regex: query[queryKey], $options: 'i' } });
          break;

        default:
          break;
      }
    }
    return acc;
  }, []);
};

// ...
// didn't copy the filter definitions over since nothing changed
// export const ADDRESS_FILTERS = {

```

server/helpers/filters.js
```diff
@@ -16,22 +16,61 @@ export const buildFilterQuery = (allFilters, query) => {
...
+            });
+          }
           break;
 
         case FILTER_TYPE.INTEGER:
-          acc.push({ [filterKey]: parseInt(query[queryKey], 10) });
+          {
+            let integerQueryValue = parseInt(query[queryKey], 10);
+            if (shouldSort) {
+              if (parseInt(rangeTo, 10) === 0) {
```
andela-hpopoola, line 61: One of the filter is meaning here; we should have another if statement testing for 'rangeFrom' 🤷🏽‍♂️

andela-hpopoola (changes requested):
Thank you for working on this. We need to write a few tests to ensure that they work as expected:

1. We need to test for the four filters separately (please create a separate test for them and don't update the existing tests as this will be the only test we would use for `range` test)
2. The four filters above should be in 2 parts:
    a. One part should test for Date (equals, to, from, and range)
    b. The second part should test for integer (equals, to, from, and range)

We only need to write the above test for the user model, once they work as expected, we can be rest assured that they would work as expected for other models too since we would have at lease on filter based on them in our test.

andela-hpopoola (approved):
Thank you for helping out with the tests. I noticed lots of bugs with the code, I wasn't too bothered with them since I know the tests will force us to correct them. Thank you for making the corrections. Please merge when ready.

PR #126, opened 2021-02-16
#176842968 add sort to `/user/all` endpoint

tobieeniafe:
#### What does this PR do?
The purpose of this PR is to...

Have the following working...

`GET /api/v1/user/?sortBy=[parameter]&sortDirection=[asc/desc]`

#176842968

### DIFF
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -13,7 +13,7 @@ export const FILTER_TYPE = {
 
 const SORT_DIRECTION = {
   ASC: 1,
-  DES: -1,
+  DESC: -1,
 };
 
 export const buildFilterQuery = (allFilters, query) => {
@@ -55,25 +55,23 @@ export const buildFilterQuery = (allFilters, query) => {
   }, []);
 };
 
-export const buildSortQuery = (sortFilter, modelFilter, query) => {
-  return Object.entries(sortFilter).reduce((acc, [queryKey, { type }]) => {
-    if (query[queryKey]) {
-      switch (type) {
-        case 'key':
-          acc.key = modelFilter[query[queryKey]].key || query[queryKey];
-          break;
+export const buildSortQuery = (modelFilter, query) => {
+  const { sortBy, sortDirection } = query;
 
-        case 'value':
-          acc.value = SORT_DIRECTION[query[queryKey].toUpperCase()];
-          break;
+  if (sortBy && Object.keys(modelFilter).includes(sortBy)) {
+    const direction =
+      sortDirection?.toUpperCase() === 'DESC' ? SORT_DIRECTION.DESC : SORT_DIRECTION.ASC;
 
-        default:
-          break;
-      }
-    }
-    return acc;
-  }, {});
+    const key = modelFilter[sortBy].key || sortBy;
+    return { [key]: direction };
+  }
+  return {};
 };
+
+export const buildFilterAndSortQuery = (allFilters, query) => ({
+  filterQuery: buildFilterQuery(allFilters, query),
+  sortQuery: buildSortQuery(allFilters, query),
+});
 
 export const ADDRESS_FILTERS = {
   city: { key: 'address.city', type: FILTER_TYPE.STRING },
@@ -176,9 +174,4 @@ export const VISITATION_FILTERS = {
   visitorEmail: { type: FILTER_TYPE.STRING },
   visitorName: { type: FILTER_TYPE.STRING },
   visitorPhone: { type: FILTER_TYPE.STRING },
-};
-
-export const SORT_FILTERS = {
-  sortBy: { type: 'key' },
-  sortDirection: { type: 'value' },
 };
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -11,6 +11,11 @@ export const FILTER_TYPE = {
   STRING: 'string',
 };
 
+const SORT_DIRECTION = {
+  ASC: 1,
+  DES: -1,
+};
+
 export const buildFilterQuery = (allFilters, query) => {
   return Object.entries(allFilters).reduce((acc, [queryKey, { key, type }]) => {
     const filterKey = key || queryKey;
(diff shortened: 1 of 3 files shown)


server/helpers/filters.js
```diff
@@ -50,6 +55,26 @@ export const buildFilterQuery = (allFilters, query) => {
   }, []);
 };
 
+export const buildSortQuery = (sortFilter, modelFilter, query) => {
+  return Object.entries(sortFilter).reduce((acc, [queryKey, { type }]) => {
```
andela-hpopoola, line 59: I believe this is overkill and not necessary. In fact, you made it unnecessarily complex. It is better to try to understand the purpose of what is expected of you instead of copying and pasting previous code.

Here is my updated code

```js
export const buildSortQuery = (modelFilter, query) => {
  const { sortBy, sortDirection } = query;

  // check if sortBy is a valid filter value
// if it is not valid, then return an empty object {}
  if (sortBy && Object.keys(modelFilter).includes(sortBy)) {
   // check direction and return DESC only if user defines DESC
  // this forces the direction to be ASC even with invalid input or when the user doesn't give a direction parameter
    const direction =
      sortDirection?.toUpperCase === 'DESC' ? SORT_DIRECTION.DESC : SORT_DIRECTION.ASC;
 // return the sort query
    return {
      [sortBy]: direction,
    };
  }
  return {};
};

```

server/helpers/filters.js
```diff
@@ -152,3 +177,8 @@ export const VISITATION_FILTERS = {
   visitorName: { type: FILTER_TYPE.STRING },
   visitorPhone: { type: FILTER_TYPE.STRING },
 };
+
+export const SORT_FILTERS = {
```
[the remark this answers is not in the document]
andela-hpopoola: Based on the updated code, this is not needed.

server/services/user.service.js
```diff
@@ -264,9 +264,18 @@ export const updateUser = async (updatedUser) => {
 
 export const getAllUsers = async ({ page = 1, limit = 10, ...query } = {}) => {
   const filterQuery = buildFilterQuery(USER_FILTERS, query);
+  let sortQuery;
```
andela-hpopoola, line 267: No longer needed.

server/services/user.service.js
```diff
@@ -264,9 +264,18 @@ export const updateUser = async (updatedUser) => {
...
+
+  const sortKeys = buildSortQuery(SORT_FILTERS, USER_FILTERS, query);
```
andela-hpopoola, line 269:
```suggestion
    const sortQuery = buildSortQuery(USER_FILTERS, query);
```

andela-hpopoola: Now that the 2 functions have the same number of parameters, we can combine them together to make them easier to work with.

```suggestion
  const { filterQuery, sortQuery } = buildFilterAndSortQuery(USER_FILTERS, query);
```

Then in filters.js, you will define the new function as:

```js
export const buildFilterAndSortQuery = (allFilters, query) => ({
  filterQuery: buildFilterQuery(allFilters, query),
  sortQuery: buildSortQuery(allFilters, query),
});
```

server/services/user.service.js
```diff
@@ -264,9 +264,18 @@ export const updateUser = async (updatedUser) => {
...
+  if (Object.keys(sortKeys).length === 0) {
```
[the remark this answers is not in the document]
andela-hpopoola: Since I have added it to the updated function, it is safe to remove this.

andela-hpopoola (changes requested):
Thank you for working on this. I have updated the sort function into a simpler format to ensure that it is easier for everyone of us to understand.

Kindly help with the test so that we can merge this asap. Thank you and well done.

andela-hpopoola:
Also, can you update the title of the Pull Request, this is `Sort` not `Filter`

test/controllers/user.controller.test.js
```diff
@@ -3372,5 +3372,91 @@ describe('User Controller', () => {
         useExistingUser: true,
       });
     });
+
+    describe('User Sorting', () => {
```
andela-hpopoola, line 3376: For this, I think it is better you move it outside this test block and create its own users for it. You can try to mix it up to test for all the filter type that we have defined:

```
  BOOLEAN: 'boolean',
  DATE: 'date',
  INTEGER: 'integer',
  OBJECT_ID: 'objectId', (this is not used in user filter, might have to write a test for it in property when we implement sort for property)
  STRING: 'string',
```

That means we would have at least one context to test for each datatype.

To ensure the test covers all edge cases, we need to ensure we create users for each test cases. For example, we should create 3 new users and sort ASC and DESC and test that the 3 users are returned as expected.

test/controllers/user.controller.test.js
```diff
@@ -3372,5 +3372,91 @@ describe('User Controller', () => {
...
+      context('when sorted by role', () => {
+        context('in ascending order', () => {
+          const sortFilters = {
+            limit: 100,
+            sortBy: 'role',
+            sortDirection: 'asc',
```
andela-hpopoola, line 3382: For ASC, lets skip the sortDirection to be certain that it works as expected when it is not given.

test/controllers/user.controller.test.js
```diff
@@ -3372,5 +3372,91 @@ describe('User Controller', () => {
...
+          };
+          const filteredParams = querystring.stringify(sortFilters);
+
+          it('returns matched users', (done) => {
+            request()
+              [method](`${endpoint}?${filteredParams}`)
+              .set('authorization', adminToken)
+              .end((err, res) => {
+                expect(res.body.result[0].role).to.be.eql(3);
+                expect(res.body.result[res.body.result.length - 1].role).to.be.eql(0);
```
andela-hpopoola, line 3412: I will prefer we test each of the returned value to be sure that they work as expected. Like

```js
// user1, user2, user3 are only used for explaination purposes only 
// and all the three users should be tested this way in ASC and DESC to confirm sort works as expected
expect(res.body.result[0]._id).to.be.eql(user1._id);
expect(res.body.result[1]._id).to.be.eql(user2._id);
expect(res.body.result[2]._id).to.be.eql(user3._id);
```

test/controllers/user.controller.test.js
andela-hpopoola: Also, after writting the test for single users, we should also write the same test for nested values to ensure that they work as expected. 

Once we have completed all the tests for the user sorting, it will be safe to ignore the test for other models when we implement their sort functionality.

andela-hpopoola (changes requested):
Thank you for the updates. I left a few comments on ensuring we improve the test cases, I will provide more guidance during standup.

To ensure that we can skip the sorting tests for other models, we need to make the sorting test here very comprehensive.

test/controllers/user.controller.test.js
andela-hpopoola: Lastly, we need another test for unknown sort filters

andela-hpopoola (approved):
Thank you for working on this. This looks good to me. Feel free to merge.

PR #127, opened 2021-02-18
#176955705 generate next payment dates for offer

tobieeniafe:
#### What does this PR do?
Generate next payment dates for offer

Generate next payment dates for offer

`POST /api/v1/offer/create`

#176955705

### DIFF
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -1,5 +1,5 @@
 import mongoose from 'mongoose';
-import { format, add } from 'date-fns';
+import { add } from 'date-fns';
 import Offer from '../models/offer.model';
 import Enquiry from '../models/enquiry.model';
 import { ErrorHandler } from '../helpers/errorHandler';
@@ -279,20 +279,23 @@ export const generatePaymentSchedules = (offer) => {
     paymentFrequency,
     handOverDate,
   } = offer;
-  const paymentDates = [{ date: format(offer.handOverDate, 'yyyy-MM-dd'), amount: initialPayment }];
+  const paymentDates = [{ date: offer.handOverDate, amount: initialPayment }];
 
   const numberOfPaymentsToBeMade = (totalAmountPayable - initialPayment) / periodicPayment;
 
-  if (numberOfPaymentsToBeMade > 0 && numberOfPaymentsToBeMade < 1) {
-    const paymentDate = format(add(handOverDate, { days: paymentFrequency }), 'yyyy-MM-dd');
-    paymentDates.push({ date: paymentDate, amount: totalAmountPayable - initialPayment });
-  } else {
-    for (let i = 1; i <= numberOfPaymentsToBeMade; i += 1) {
-      const paymentDate = format(add(handOverDate, { days: paymentFrequency * i }), 'yyyy-MM-dd');
-      paymentDates.push({ date: paymentDate, amount: periodicPayment });
-    }
+  const fractionPayment = (totalAmountPayable - initialPayment) % periodicPayment;
+
+  for (let i = 1; i <= numberOfPaymentsToBeMade; i += 1) {
+    const paymentDate = add(handOverDate, { days: paymentFrequency * i });
+    paymentDates.push({ date: paymentDate, amount: periodicPayment });
   }
 
+  if (fractionPayment > 0) {
+    const paymentDate = add(paymentDates[paymentDates.length - 1].date, {
+      days: paymentFrequency,
+    });
+    paymentDates.push({ date: paymentDate, amount: fractionPayment });
+  }
   return paymentDates;
 };
 
@@ -366,9 +369,6 @@ export const acceptOffer = async (offerToAccept) => {
   }
 
   const vendor = await getUserById(offer[0].vendorId);
-
-  // const paymentDates = generatePaymentSchedules(offer[0]);
-  // code to add dates to payment schema
 
   try {
     await assignPropertyToUser({
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -1,4 +1,5 @@
 import mongoose from 'mongoose';
+import { format, add } from 'date-fns';
 import Offer from '../models/offer.model';
 import Enquiry from '../models/enquiry.model';
 import { ErrorHandler } from '../helpers/errorHandler';
@@ -270,6 +271,25 @@ export const getOffer = async (offerId, user) => {
   return offer;
 };
 
+const generatePaymentDates = (offer) => {
+  const paymentDates = [];
+
+  const numberOfPaymentsToBeMade =
+    (offer.totalAmountPayable - offer.initialPayment) / offer.periodicPayment;
+
+  const { paymentFrequency } = offer;
+
+  for (let i = 0; i < numberOfPaymentsToBeMade; i += 1) {
+    const paymentDate = format(
+      add(offer.handOverDate, { days: paymentFrequency * i }),
+      'yyyy-MM-dd',
+    );
+    paymentDates.push(paymentDate);
+  }
+
+  return paymentDates;
+};
+
 export const createOffer = async (offer) => {
   const enquiry = await getEnquiryById(offer.enquiryId).catch((error) => {
     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
@@ -291,6 +311,8 @@ export const createOffer = async (offer) => {
   });
 
   const referenceCode = await generateReferenceCode(enquiry.propertyId);
+  const paymentDates = generatePaymentDates(offer);
+
   const user = await getUserById(enquiry.userId).catch((error) => {
     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
   });
@@ -305,6 +327,7 @@ export const createOffer = async (offer) => {
       userId: enquiry.userId,
       propertyId: enquiry.propertyId,
       referenceCode,
+      paymentDates,
     }).save();
     await approveEnquiry({ enquiryId: enquiry._id, vendor });
     const offerInfo = await getOffer(newOffer._id, user);
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -1,5 +1,5 @@
 import mongoose from 'mongoose';
-import { format, add } from 'date-fns';
+import { format, add, parseISO } from 'date-fns';
 import Offer from '../models/offer.model';
 import Enquiry from '../models/enquiry.model';
 import { ErrorHandler } from '../helpers/errorHandler';
@@ -273,17 +273,14 @@ export const getOffer = async (offerId, user) => {
 
 const generatePaymentDates = (offer) => {
   const paymentDates = [];
+  const { totalAmountPayable, initialPayment, periodicPayment, paymentFrequency } = offer;
 
-  const numberOfPaymentsToBeMade =
-    (offer.totalAmountPayable - offer.initialPayment) / offer.periodicPayment;
-
-  const { paymentFrequency } = offer;
+  const numberOfPaymentsToBeMade = (totalAmountPayable - initialPayment) / periodicPayment;
 
   for (let i = 0; i < numberOfPaymentsToBeMade; i += 1) {
-    const paymentDate = format(
-      add(offer.handOverDate, { days: paymentFrequency * i }),
-      'yyyy-MM-dd',
-    );
+    const handOverDate =
+      offer.handOverDate === 'string' ? parseISO(offer.handOverDate) : offer.handOverDate;
+    const paymentDate = format(add(handOverDate, { days: paymentFrequency * i }), 'yyyy-MM-dd');
     paymentDates.push(paymentDate);
   }
 
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -1,5 +1,5 @@
 import mongoose from 'mongoose';
-import { format, add, parseISO } from 'date-fns';
+import { format, add } from 'date-fns';
 import Offer from '../models/offer.model';
 import Enquiry from '../models/enquiry.model';
 import { ErrorHandler } from '../helpers/errorHandler';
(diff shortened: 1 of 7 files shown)


server/services/offer.service.js
```diff
@@ -270,6 +271,25 @@ export const getOffer = async (offerId, user) => {
   return offer;
 };
 
+const generatePaymentDates = (offer) => {
+  const paymentDates = [];
+
+  const numberOfPaymentsToBeMade =
+    (offer.totalAmountPayable - offer.initialPayment) / offer.periodicPayment;
+
+  const { paymentFrequency } = offer;
```
andela-hpopoola, line 280: This can be defined after paymentDates so as to get the variables for `numberOfPaymentsToBeMade`.

```suggestion
  const { paymentFrequency, totalAmountPayable, initialPayment, periodicPayment  } = offer;
```

server/services/offer.service.js
```diff
@@ -270,6 +271,25 @@ export const getOffer = async (offerId, user) => {
...
+
+  for (let i = 0; i < numberOfPaymentsToBeMade; i += 1) {
+    const paymentDate = format(
```
andela-hpopoola, line 283: We need the paymentDate with the amount they will pay. Don't forget that initial payment details should come first

server/services/offer.service.js
```diff
@@ -305,6 +327,7 @@ export const createOffer = async (offer) => {
       userId: enquiry.userId,
       propertyId: enquiry.propertyId,
       referenceCode,
+      paymentDates,
```
andela-hpopoola, line 330: On a better thought, we might not even need to save this in the database as we can always calculate it when it is needed. Also, we don't know everything we need to save now, so it is better not to waste time implementing this.

andela-hpopoola: In another PR, let's save the next payment in the `nextPayment` model. We will need to calculate the next payment to be made and save it in the model.

Here are the events I think would trigger updating the nextPayment model:

1. A user accepts an offer letter
2. A users payment is confirmed
3. Expiry date of the payment has been exceeded

These are the ones I can think of for now, there might be more but we would keep updating them later.

server/services/offer.service.js
```diff
@@ -270,6 +271,25 @@ export const getOffer = async (offerId, user) => {
   return offer;
 };
 
+const generatePaymentDates = (offer) => {
```
andela-hpopoola, line 274:
```suggestion
const generatePaymentSchedules = (offer) => {
```

This PR is already getting messy, lets conclude this PR by writing services tests to confirm that this function works as expected. Lets test it with various cases like:

1. When totalAmountPayable is equal to  initial payment
2. When totalAmountPayable is slightly bigger than initial payment (i.e different is not up to periodic payment) 
3. When totalAmountPayable is greater than initial Payment so that we can have a couple of payments to be made
4. Testing same values with other paymentFrequency to be sure they work as expected
5. Other tests you feel is necessary to ensure this works as expected.

andela-hpopoola (changes requested):
Thank you for working on this. This looks like a good start and we can always work on improving the necessary stuff in other PRs. I left a couple of comments, I want us to do this one at a time, so let's focus more on writing tests to be certain that the payment schedules work as expected.

#### Next PR implementation

The next payment should focus on how to get the next payment and resolve old payments for previously saved in the next payment model. This model is super necessary for cron jobs so we should be some thoughts on how/when to generate a new record here and the necessary fields to put. 

Since you worked on the spike, you should know the other things I might have missed. However, I am sure we will need the following 
- offerId (to connect the payment to the model)
- userId (for sending mails)
- propertyId (if we can easily get property info via offerId, then this is not necessary)
- vendorId (enables us to show the vendor the next payment they are expecting)
- amount (the next payment to be made)
- expiresOn (the date the next payment expires) We can use this to send a mail/ message before a payment expires. Sending of mails can be based on predefined days like a day before, 7 days before, 30 days before and 60 days before expiry date
- status (open or closed, a row can be closed by 2 ways, automatically or via a transaction. We should automatically close open rows that have expired, we need to  think on the best way to generate another one the next expiry date and amount if we close automatically -- maybe a cron job to close and create a new nextPayment row 🤷🏽‍♂️ )

// not sure how much of this we would need
- resolvedViaTransaction - (is it resolved by cronjob or transaction)
- transactionId - (Id of transaction used to resolve this)
- resolvedDate - (date the payment was resolved)

I am sure I might still be missing some stuff or edge cases since I didn't do a complex deep dive into this but I hope you can fill gaps and make this work as expected. 

Please note that this is still complicated to me too as your spike didn't cover much technical base, so expect a few changes even after implementing some stuff. I will prefer we take this slow and steady to totally get what we are doing. 

Thank you.

server/services/offer.service.js
```diff
@@ -270,6 +271,22 @@ export const getOffer = async (offerId, user) => {
...
+  const paymentDates = [];
+  const { totalAmountPayable, initialPayment, periodicPayment, paymentFrequency } = offer;
+
+  const numberOfPaymentsToBeMade = (totalAmountPayable - initialPayment) / periodicPayment;
+
+  for (let i = 0; i < numberOfPaymentsToBeMade; i += 1) {
+    const handOverDate =
+      offer.handOverDate === 'string' ? parseISO(offer.handOverDate) : offer.handOverDate;
```
andela-hpopoola, line 282: I think it is better to remove this since we already know the cause of the failing tests. It is an hot fix, so we should only use it if it is becoming difficult to troubleshoot the root cause.

Also, this should be 
```suggestion
      typeof offer.handOverDate === 'string' ? parseISO(offer.handOverDate) : offer.handOverDate;
```

server/services/offer.service.js
```diff
@@ -271,17 +271,26 @@ export const getOffer = async (offerId, user) => {
...
-  const { totalAmountPayable, initialPayment, periodicPayment, paymentFrequency } = offer;
+export const generatePaymentSchedules = (offer) => {
+  const {
+    totalAmountPayable,
+    initialPayment,
+    periodicPayment,
+    paymentFrequency,
+    handOverDate,
+  } = offer;
+  const paymentDates = [{ date: format(offer.handOverDate, 'yyyy-MM-dd'), amount: initialPayment }];
```
andela-hpopoola, line 282: Can we make the date be in date format instead of a string to make date comparison easier in the future.

server/services/offer.service.js
```diff
@@ -271,17 +271,26 @@ export const getOffer = async (offerId, user) => {
...
-      offer.handOverDate === 'string' ? parseISO(offer.handOverDate) : offer.handOverDate;
-    const paymentDate = format(add(handOverDate, { days: paymentFrequency * i }), 'yyyy-MM-dd');
-    paymentDates.push(paymentDate);
+  if (numberOfPaymentsToBeMade > 0 && numberOfPaymentsToBeMade < 1) {
+    const paymentDate = format(add(handOverDate, { days: paymentFrequency }), 'yyyy-MM-dd');
+    paymentDates.push({ date: paymentDate, amount: totalAmountPayable - initialPayment });
+  } else {
+    for (let i = 1; i <= numberOfPaymentsToBeMade; i += 1) {
+      const paymentDate = format(add(handOverDate, { days: paymentFrequency * i }), 'yyyy-MM-dd');
+      paymentDates.push({ date: paymentDate, amount: periodicPayment });
```
andela-hpopoola, line 292: This didn't cater for when last payment is not equal to periodic payment.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
...
+      it('returns a valid accepted offer', async () => {
+        const paymentDates = generatePaymentSchedules(offer1);
+        expect(paymentDates.length).to.eql(6);
+        expect(paymentDates[0].amount).to.eql(50000);
+        expect(paymentDates[1].amount).to.eql(10000);
+        expect(paymentDates[2].amount).to.eql(10000);
+        expect(paymentDates[3].amount).to.eql(10000);
+        expect(paymentDates[4].amount).to.eql(10000);
+        expect(paymentDates[5].amount).to.eql(10000);
+        expect(paymentDates[0].date).to.have.string('2020-12-11');
```
andela-hpopoola, line 696: This would be better as dates not string.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
...
+        updatedBy: vendor._id,
+        price: 100000,
+      },
+      { generateId: true },
+    );
+    const enquiry = EnquiryFactory.build(
+      { userId: demoUser._id, propertyId: property._id },
+      { generateId: true },
+    );
+    const offer1 = OfferFactory.build(
```
andela-hpopoola, line 638: Offer1, offer2, and offer3 are vague and don't really convene any meaning; Since these offers are used once in each test, it is better to define them in the context they are used to make the tests easier to read and update.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
       });
     });
   });
+
+  describe('#generatePaymentSchedules', () => {
+    const demoUser = UserFactory.build({ role: USER_ROLE.USER }, { generateId: true });
```
andela-hpopoola, line 625: I think you approached this like controllers tests.  This is not needed as it is not used in the service.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
...
+    const property = PropertyFactory.build(
```
andela-hpopoola, line 626: This is not needed too.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
...
+      {
+        addedBy: vendor._id,
+        updatedBy: vendor._id,
+        price: 100000,
+      },
+      { generateId: true },
+    );
+    const enquiry = EnquiryFactory.build(
```
andela-hpopoola, line 634: This is not needed too.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
...
+      { userId: demoUser._id, propertyId: property._id },
+      { generateId: true },
+    );
+    const offer1 = OfferFactory.build(
+      {
+        enquiryId: enquiry._id,
```
andela-hpopoola, line 640: enquiryId and vendorId are not needed for this service too.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
...
+        totalAmountPayable: 100000,
+        initialPayment: 95000,
+        periodicPayment: 10000,
+        paymentFrequency: 14,
+        handOverDate: parseISO('2020-12-11'),
+      },
+      { generateId: true },
+    );
+
+    beforeEach(async () => {
```
andela-hpopoola, line 677: I don't think this is needed too.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
...
+      await addUser(demoUser);
+      await addProperty(property);
+      await addEnquiry(enquiry);
+    });
+
+    context('when totalAmountPayable is greater than initial payment', () => {
+      beforeEach(async () => {
```
andela-hpopoola, line 684: This too is not necessary.

test/services/offer.service.test.js
```diff
@@ -618,4 +620,112 @@ describe('Offer Service', () => {
...
+      },
+      { generateId: true },
+    );
+
+    const offer3 = OfferFactory.build(
+      {
+        enquiryId: enquiry._id,
+        vendorId: vendor._id,
+        totalAmountPayable: 100000,
+        initialPayment: 95000,
```
andela-hpopoola, line 669: Lets make this 75000, it will make you notice an important section not covered by the code.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments but I am more concerned about ensuring that we return dates for the payment dates and not string. Thank you.

test/services/offer.service.test.js
tobieeniafe: It's needed if we want the `addEnquiry` to work see `enquiry.service.js:19`

test/services/offer.service.test.js
tobieeniafe: It's needed if we want the `addEnquiry` to work see `enquiry.service.js:19`

test/services/offer.service.test.js
andela-hpopoola: I still believe this can be removed (You might want to give a reason instead of resolving this without implementing it)

test/services/offer.service.test.js
andela-hpopoola: This too ☝🏽

test/services/offer.service.test.js
```diff
@@ -681,50 +642,91 @@ describe('Offer Service', () => {
...
+          vendorId: vendor._id,
+          totalAmountPayable: 100000,
+          initialPayment: 100000,
+          periodicPayment: 10000,
+          paymentFrequency: 30,
+          handOverDate: new Date('2021-03-01'),
+        },
+        { generateId: true },
+      );
       beforeEach(async () => {
```
andela-hpopoola, line 691: This too. This is a service with no database interaction.

test/services/offer.service.test.js
```diff
@@ -681,50 +642,91 @@ describe('Offer Service', () => {
...
-        expect(paymentDates[1].amount).to.eql(5000);
-        expect(paymentDates[0].date).to.have.string('2020-12-11');
-        expect(paymentDates[1].date).to.have.string('2020-12-25');
+        const paymentDates = generatePaymentSchedules(offer);
+        expect(paymentDates.length).to.eql(4);
+        expect(paymentDates[0].amount).to.eql(75000);
+        expect(paymentDates[1].amount).to.eql(10000);
+        expect(paymentDates[2].amount).to.eql(10000);
+        expect(paymentDates[3].amount).to.eql(5000);
+        expect(paymentDates[0].date).to.eql(new Date('2021-03-01T00:00:00.000Z'));
```
andela-hpopoola, line 726: We are already parsing the Date, so it should be safe to pass in a human readable date. Let's update this in all instance 🙏🏽 

```suggestion
        expect(paymentDates[0].date).to.eql(new Date('2021-03-01'));
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments, this should be ready to go when you have fixed the comments.

andela-hpopoola (approved):
Thank you for working on this. This looks good to me.

PR #128, opened 2021-02-19
#176955705 add next payment schema

tobieeniafe:
#### What does this PR do?
Add nextPayment schema

Add nextPayment schema

NA

#176955705

### DIFF
--- a/server/models/nextPayment.model.js
+++ b/server/models/nextPayment.model.js
@@ -1,5 +1,4 @@
 import mongoose from 'mongoose';
-import { PAYMENT_STATUS } from '../helpers/constants';
 
 /**
  * @swagger
@@ -14,7 +13,7 @@ import { PAYMENT_STATUS } from '../helpers/constants';
  *          - propertyId
  *          - resolvedDate
  *          - resolvedViaTransaction
- *          - status
+ *          - resolved
  *          - transactionId
  *          - vendorId
  *        properties:
@@ -29,9 +28,9 @@ import { PAYMENT_STATUS } from '../helpers/constants';
  *          resolvedDate:
  *            type: date
  *          resolvedViaTransaction:
- *            type: string
- *          status:
- *            type: string
+ *            type: boolean
+ *          resolved:
+ *            type: boolean
  *          transactionId:
  *            type: string
  *          vendorId:
@@ -42,8 +41,8 @@ import { PAYMENT_STATUS } from '../helpers/constants';
  *           offerId: 5f22f7f8c790039da1242381
  *           propertyId: 5f22f8aec790039da1242382
  *           resolvedDate: 2020-01-01
- *           resolvedViaTransaction: Automatic
- *           status: Open
+ *           resolvedViaTransaction: true
+ *           resolved: false
  *           transactionId: 5f2b39035a086cfc4b7fa722
  *           userId: 5f2b39035a086cfc4b7fa7f6
  *           vendorId: 5f22f7f8c790039da1242381
@@ -72,11 +71,11 @@ const NextPaymentSchema = new mongoose.Schema(
       type: Date,
     },
     resolvedViaTransaction: {
-      type: String,
+      type: Boolean,
     },
-    status: {
-      type: String,
-      default: PAYMENT_STATUS.OPEN,
+    resolved: {
+      type: Boolean,
+      default: false,
     },
     transactionId: {
       type: ObjectId,
--- a/server/models/nextPayment.model.js
+++ b/server/models/nextPayment.model.js
@@ -1,0 +1,98 @@
+import mongoose from 'mongoose';
+import { PAYMENT_STATUS } from '../helpers/constants';
+
+/**
+ * @swagger
+ *  components:
+ *    schemas:
+ *      NextPayment:
+ *        type: object
+ *        required:
+ *          - expectedAmount
+ *          - expiresOn
+ *          - offerId
+ *          - propertyId
+ *          - resolvedDate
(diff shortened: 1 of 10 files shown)


server/models/nextPayment.model.js
```diff
@@ -0,0 +1,98 @@
...
+    },
+    propertyId: {
+      type: ObjectId,
+      required: true,
+    },
+    resolvedDate: {
+      type: Date,
+    },
+    resolvedViaTransaction: {
+      type: String,
```
andela-hpopoola, line 75: Lets make this a boolean.

server/models/nextPayment.model.js
```diff
@@ -0,0 +1,98 @@
...
+    },
+    status: {
```
andela-hpopoola, line 77: We can also change this to resolved so that we can use a boolean here that defaults to false. Saves us the stress of importing the values of open and close each time.

andela-hpopoola (changes requested):
Thank you for working on this. The next step is to add next payment to the database when a user accepts an offer. Thank you.

server/models/nextPayment.model.js
```diff
@@ -29,9 +28,9 @@ import { PAYMENT_STATUS } from '../helpers/constants';
  *          resolvedDate:
  *            type: date
  *          resolvedViaTransaction:
- *            type: string
- *          status:
- *            type: string
+ *            type: boolean
+ *          resolved:
```
andela-hpopoola, line 32: Let's make this come after propertyId to stick with the alphabetical order 😼

andela-hpopoola:
Thank you for making the changes.

>  The next step is to add next payment to the database when a user accepts an offer

Can we add step above ☝🏽  to this PR to increase our confidence that we are on the right track 🙏🏽

server/services/offer.service.js
```diff
@@ -368,6 +369,20 @@ export const acceptOffer = async (offerToAccept) => {
     throw new ErrorHandler(httpStatus.PRECONDITION_FAILED, 'Offer has expired');
   }
 
+  const paymentSchedule = generatePaymentSchedules({
+    ...offer[0],
+    responseDate: add(Date.now(), { days: 0 }),
```
andela-hpopoola, line 374: Not sure if I am right here, but the initialPayment should be due on the day the offer letter expires. Maybe you need to ask Nnamdi to be 100% sure.

server/services/offer.service.js
```diff
@@ -384,6 +400,7 @@ export const acceptOffer = async (offerToAccept) => {
           signature: offerToAccept.signature,
           contributionReward,
           responseDate: Date.now(),
+          paymentSchedule,
```
andela-hpopoola, line 403: I still feel we might not need to save the `paymentSchedule` since we can regenerate it from `generatePaymentSchedules` when needed. 

Do you have a good reason why you think it is better to save it?

server/services/offer.service.js
```diff
@@ -376,6 +391,7 @@ export const acceptOffer = async (offerToAccept) => {
       userId: offer[0].userId,
       vendor,
     });
+    await addNextPayment(nextPayment);
```
andela-hpopoola, line 394: Can we  add tests to confirm this 🙏🏽

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments but I am mostly concerned about testing that the `nextPayment` has been saved as expected and with the right values. I am currently not sure if it is better to test this via services or controller test (the choice is yours as long as your tests are optimal and tests for the required stuff)

server/services/offer.service.js
```diff
@@ -277,16 +278,16 @@ export const generatePaymentSchedules = (offer) => {
     initialPayment,
     periodicPayment,
     paymentFrequency,
-    handOverDate,
+    responseDate,
```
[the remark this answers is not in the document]
tobieeniafe: The initial code was using the `handOverDate` to calculate the dates. But the current one uses the date the offer is accepted from, hence the `responseDate`. The `responseDate` was added here instead of spreading it in the `generatePaymentSchedules` method.

server/services/offer.service.js
tobieeniafe: Just spoke to Nnamdi, and the offer is supposed to contain an `initialPaymentDate` in the offer. Stating when the first payment is supposed to be made. Would make the appropriate changes to the model and tests.

server/services/offer.service.js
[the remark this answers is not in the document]
tobieeniafe: This has been resolved as Nnamdi explained that the calculation should use the `initialPaymentDate`

server/services/offer.service.js
tobieeniafe: Wanted to be sure the right dates are generated. But found a way around it without saving it

andela-hpopoola (approved):
Thank you for working on this. This looks good to me for now. Feel free to merge when ready. 

Here are the remaining stories in my opinion.
1. Handle the events when a user's payment has been confirmed. Not sure we have everything needed for this but after a user makes any payment and the payment has been approved, we should generate the next payment. Past payments should be taken into consideration when generating a new NextPayment as we don't want to overgenerate or under generate the next payment.
2. After the above story, the next story is to handle any expired next payment. When this endpoint is accessed it takes the next expired next payment in the nextPayment model, related necessary information is also retrieved to generates a new NextPayment.

3. Create an endpoint to send reminders. Let's stick with 60 days, 30days, 14days, 7days, 1 day for now. This will send a mail reminder while we integrate the SMS reminder later in the future.

Thank you.

PR #129, opened 2021-02-20
#176955814 make provisions for multiple floor plans

tobieeniafe:
#### What does this PR do?
Make provisions for multiple floor plans

Make provisions for multiple floor plans

`POST /api/v1/property/add`
`PUT /api/v1/property/update`

#176955814


andela-hpopoola (changes requested):
Thank you for working on this. However, I think we can improve your current implementation to make it an optimal solution. 

The problem here is how a user would easily delete or update a floor plan (I thought I explained this when I compared it to the gallery implementation). Most times when we are dealing with array, we should always keep in mind how the user would update or remove an entry from the array.

The floorplan model should look like this imho:

```js
floorPlans: {
[ {name: String, plan: String} ]
}
```
I believe it would be more manageable if we have a different endpoint to add, update and remove a floor plan. Unfortunately, this implementation doesn't cater for this.

andela-hpopoola:
To fix this, lets create endpoints for the floorplans to cater for these implementation:

```
POST /property/:id/floorplan
PUT /property/:id/floorplan
DELETE /property/:id/floorplan
```

The best part is that we would use the same implementation for gallery.

tobieeniafe:
> To fix this, lets create endpoints for the floorplans to cater for these implementation:
> 
> ```
> POST /property/:id/floorplan
> PUT /property/:id/floorplan
> DELETE /property/:id/floorplan
> ```
> 
> The best part is that we would use the same implementation for gallery.

The floorPlan can be updated directly from the update property route. Is there a need to create a route solely just for this purpose?

andela-hpopoola:
> The floorPlan can be updated directly from the update property route. Is there a need to create a route solely just for this purpose?

Yes, please. Updating the floor plan with the generic update property route doesn't cater to the necessary checks when updating floor plans and this is prone to multiple errors since this will have to be handled by the agent that is accessing the endpoint. The test you added caters for total replacement, not actual updates 😏 

Also, the additional endpoint saves time and makes it easier to work with floor plans without the complexities of properties themselves. To update or delete a floorplan, it is easier to send the information of the floor plan you wish to change with the id of the floor plan instead of sending the whole floor plans 🤷🏽‍♂️ 

Lastly, the implementation shouldn't be time-consuming as we have similar implementations for directors in user models.

Thank you.

test/controllers/property.controller.test.js
```diff
@@ -2797,4 +2797,368 @@ describe('Property Controller', () => {
...
+        ],
+        addedBy: vendorUser._id,
+        updatedBy: vendorUser._id,
+      },
+      { generateId: true },
+    );
+    const endpoint = `/api/v1/property/${property._id}/floorplan`;
+    const method = 'post';
+
+    const data = {
```
andela-hpopoola, line 2818: Since we are working with the floor plan endpoint, I think it is better to just send the properties without the additional floor plan.

```js
const data = {
    name: 'Second Floor',
    plan: 'https://ballers.ng/secondfloor.png',
}
```
```

File: test/controllers/property.controller.test.js [hunk 2]
```diff
@@ -2797,4 +2797,368 @@ describe('Property Controller', () => {
...
+        addedBy: vendorUser._id,
+        updatedBy: vendorUser._id,
+      },
+      { generateId: true },
+    );
+    const endpoint = `/api/v1/property/${property._id}/floorplan`;
+    const method = 'put';
+
+    const data = {
+      floorPlan: {
```
andela-hpopoola, line 2944: Same comment here to make it easier to manage.

```js
const data = {
    name: 'Bottom Floor',
    plan: 'https://ballers.ng/bottomfloor.png'
    floorPlanId: property.floorPlans[1]._id
}
```

andela-hpopoola (approved):
Thank you for working on this. My only comment is ensuring we send the data directly instead of enclosing them in the `floorPlan` object. Feel free to merge after resolving this.

PR #130, opened 2021-02-21
#176955779 update neighborhood for properties

tobieeniafe:
#### What does this PR do?
Update neighborhood for properties

Update neighborhood for properties

`POST /api/v1/property/add`

#176955779

### DIFF
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -9,6 +9,7 @@ import {
   getAvailablePropertyOptions,
   getAssignedPropertyByOfferId,
   getAssignedProperties,
+  updateNeighborhood,
 } from '../services/property.service';
 import httpStatus from '../helpers/httpStatus';
 
@@ -118,6 +119,18 @@ const PropertyController = {
           success: true,
           properties,
         });
+      })
+      .catch((error) => next(error));
+  },
+
+  updateNeighborhood(req, res, next) {
+    const updatedNeighborhood = req.locals;
+    const vendorId = req.user._id;
+    updateNeighborhood({ ...updatedNeighborhood, vendorId })
+      .then((property) => {
+        res
+          .status(httpStatus.OK)
+          .json({ success: true, message: 'Neighborhood updated', property });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -9,7 +9,9 @@ import {
   getAvailablePropertyOptions,
   getAssignedPropertyByOfferId,
   getAssignedProperties,
+  addNeighborhood,
   updateNeighborhood,
+  deleteNeighborhood,
 } from '../services/property.service';
 import httpStatus from '../helpers/httpStatus';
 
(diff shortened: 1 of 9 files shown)


test/helpers.js
```diff
@@ -400,7 +400,7 @@ export const expectResponseToContainNecessaryPropertyData = (response, property)
   expect(response.bathrooms).to.be.eql(property.bathrooms);
   expect(response.toilets).to.be.eql(property.toilets);
   expect(response.titleDocument).to.be.eql(property.titleDocument);
-  expect(response.neighborhood).to.be.eql(property.neighborhood);
+  // expect(response.neighborhood).to.be.eql(property.neighborhood);
```
andela-hpopoola, line 403: If this is problematic here, we can remove it and write the necessary tests for confirming that the neighbourhood actions are done in the right way.

andela-hpopoola (changes requested):
Thank you for working on this. I know we can use the `update property` route to manipulate the neighborhood object but it cannot be easily manipulated as you have portrayed it in the tests. 

For the neighbourhood, we would also need the CRUD endpoints to handle. Trying to update this from the property update route is too complicated and would be prone to error (it is also difficult to test).

In the frontend (or anywhere the endpoint is consumed), we would want a user to click on an `x` icon to cater for deleting a neighborhood entry, sending the whole neighbourhood object back is not an ideal solution. Also adding a new neighbourhood entry is easier without sending the whole object to the `property update` endpoint. 


Let's create individual endpoints to cater for this please:

```
POST /property/:id/neighborhood - the agent consuming the endpoint will need to specify the type of neighborhood they wish to add out of the 6 defined categories
PUT /property/:id/neighborhood - neighbourhood type and and actual id is needed here to know the right property we are updating
DELETE /property/:id/neighborhood - same as above
```

Thank you.

server/controllers/property.controllers.js
```diff
@@ -121,6 +122,18 @@ const PropertyController = {
       })
       .catch((error) => next(error));
   },
+
+  updateNeighborhood(req, res, next) {
```
andela-hpopoola, line 126: Also, can we add functionality to add a new entry and delete an existing entry and add tests for them too?

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on consistency and best practices. Thank you.

test/controllers/property.controller.test.js
```diff
@@ -2309,4 +2316,451 @@ describe('Property Controller', () => {
...
+          mapLocation: {
+            longitude: 123.22,
+            latitude: 123.11,
+          },
+        },
+        {
+          name: 'babcock university',
+          timeAwayFromProperty: 15,
+          mapLocation: {
+            longitude: 123.22,
```
andela-hpopoola, line 2346: Let's use different longitude and latitude for each items here 🙏🏽

test/controllers/property.controller.test.js
```diff
@@ -2309,4 +2316,451 @@ describe('Property Controller', () => {
...
+          });
+      });
+    });
+  });
+
+  describe('Update Neighborhood', () => {
+    const property = PropertyFactory.build(
+      {
+        neighborhood: {
+          schools: [
```
andela-hpopoola, line 2468: It would have been more preferable to add 2 types here. Maybe schools and shoppingMalls.

server/services/property.service.js
```diff
@@ -365,3 +365,97 @@ export const getAssignedProperties = async (userId) =>
...
+
+  if (!property) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Invalid property');
+  }
+
+  if (neighborhoodInfo.vendorId.toString() !== property.addedBy.toString()) {
+    throw new ErrorHandler(httpStatus.FORBIDDEN, 'You are not permitted to perform this action');
+  }
+
+  let neighborhood = property.neighborhood[neighborhoodInfo.type];
```
andela-hpopoola, line 446: I noticed you are always removing a delete entry yourself by using filters, do you have any important reasons for doing this.  

I think it is safe to remove this and use `pull` to remove the entry from the database:

https://docs.mongodb.com/manual/reference/operator/update/pull/

server/services/property.service.js
```diff
@@ -365,3 +365,97 @@ export const getAssignedProperties = async (userId) =>
...
+
+  neighborhood = neighborhood.filter(
+    (n) => n._id.toString() !== neighborhoodInfo.typeId.toString(),
+  );
+
+  try {
+    return Property.findByIdAndUpdate(
+      property._id,
+      { $set: { [`neighborhood.${neighborhoodInfo.type}`]: neighborhood } },
```
andela-hpopoola, line 455:
```suggestion
 { $pull: { [`neighborhood.${neighborhoodInfo.type}`]: { _id: neighborhoodInfo.typeId } } },
```

server/schemas/property.schema.js
```diff
@@ -63,3 +86,19 @@ export const searchPropertySchema = Joi.object({
   minPrice: optionalNumber('Property Minimum Price'),
   maxPrice: optionalNumber('Property Maximum Price'),
 });
+
+export const addNeighborhoodSchema = Joi.object({
+  type: neighborhoodType,
+  neighborhood: neighborhoodInfo,
```
andela-hpopoola, line 92: I think the endpoint should be used to add one neighborhood at a time (like the updateSchema below). The purpose of this endpoint is to do a one-off simple transaction not multiple.

andela-hpopoola (approved):
Thank you for working on this. This looks good to me and should be safe to merge when you fix the comments. Thank you and Nice work.

server/schemas/property.schema.js
tobieeniafe: That was initially the idea. But a situation where the vendor decides to add multiple neighborhoods, this reduces the number of requests to be made. As one neighborhood can also be added through this same endpoint.

PR #132, opened 2021-02-28
#176955705 generate next payment when user payment is confirmed

tobieeniafe:
#### What does this PR do?
Generate next payment when user payment is confirmed

Generate next payment when user payment is confirmed

`POST /api/v1/transaction/add`

#176955705


andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to ensure we get this right. Thank you.

andela-hpopoola (changes requested):
Thank you for working on this. I believe your implementation is still lacking some indepths, so I will discuss this with you during standups. Thank you.

server/services/nextpayment.service.js
```diff
@@ -1,13 +1,83 @@
...
+
+  return { validSchedules, expectedTotal };
+};
+
+export const generateNextPaymentDate = async ({ transactionId = null, offerId }) => {
+  const offer = await getOfferById(offerId).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  const totalPaid = await getTotalPaidOnOffer(offerId).catch((error) => {
```
andela-hpopoola, line 57:
```suggestion
  const totalPaid = await getTotalTransactionByOfferId(offerId).catch((error) => {
```

server/services/nextpayment.service.js
```diff
@@ -1,13 +1,83 @@
...
+const calculateExpectedTotal = (schedule) => {
+  const today = new Date();
+  const validSchedules = schedule.reduce((acc, val) => {
+    if (val.date <= today) {
+      acc.push(val);
+    }
+    return acc;
+  }, []);
+
+  const expectedTotal = validSchedules.reduce((a, b) => {
```
andela-hpopoola, line 45: We should be able to add this to the function above but no worries, let's do that after the test.

server/services/nextpayment.service.js
```diff
@@ -1,13 +1,83 @@
...
+
+  const totalPaid = await getTotalPaidOnOffer(offerId).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  const paymentSchedule = generatePaymentSchedules(offer);
+
+  const { validSchedules, expectedTotal } = calculateExpectedTotal(paymentSchedule);
+
+  const nextPayments = await getNextPaymentByOfferId(offerId).catch((error) => {
```
andela-hpopoola, line 65: We can actually write a service to get the last row in the `nextPayments` model. Something like

```js
// this is for illustration purposes only, I didn't test it but it should work 🤷🏽‍♂️ 
NextPayment.find({ offerId: ObjectId(offerId) }).sort({_id:-1}).limit(1);

```

andela-hpopoola, line 65:
```suggestion
  const pendingNextPayments = await getNextPaymentByOfferId(offerId).catch((error) => {
```

server/services/nextpayment.service.js
```diff
@@ -1,13 +1,83 @@
...
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  const pendingPayment = nextPayments[nextPayments.length - 1];
```
andela-hpopoola, line 69: This willl no longer be needed if we get the last row for the nextPayment above

andela-hpopoola (changes requested):
Thank you so much for this. The code changes look so great and close to perfection. Though there are a few code smells (I can help remove them when you have written the tests).

I understand that the tests might be a little bit messy and difficult, no worries, take it easy but ensure you cover most of the edge cases we discussed. 

Thank you for this 🙌🏽

server/services/nextpayment.service.js
```diff
@@ -1,8 +1,18 @@
...
 import httpStatus from '../helpers/httpStatus';
+// eslint-disable-next-line import/no-cycle
+import { getOfferById, generatePaymentSchedules } from './offer.service';
+// eslint-disable-next-line import/no-cycle
+import { getTotalTransactionByOfferId } from './transaction.service';
 
-export const getNextPaymentById = async (id) => NextPayment.findById(id).select();
+const { ObjectId } = mongoose.Types.ObjectId;
+
+export const getLastPendingPayment = async (offerId) =>
```
andela-hpopoola, line 12: Not sure of the number of payments we would have later so let's try to be more explicit by using NextPayment for all payments involving the next payment service.
```suggestion
export const getLastPendingNextPayment = async (offerId) =>
```

server/services/nextpayment.service.js
```diff
@@ -11,3 +21,61 @@ export const addNextPayment = async (payment) => {
...
+  const nextPayment = {
+    expectedAmount: expectedTotal - totalPaid,
+    expiresOn: paymentSchedule[validSchedules.length].date,
+    offerId,
+    propertyId: offer.propertyId,
+    userId: offer.userId,
+    vendorId: offer.vendorId,
+  };
+
+  await resolvePendingPayment(pendingPayment[0]._id, transactionId);
```
andela-hpopoola, line 78: Can we check that this exists first before resolving. This should also fix some weird failing tests.

test/controllers/property.controller.test.js
```diff
@@ -39,6 +39,8 @@ import AddressFactory from '../factories/address.factory';
 import VisitationFactory from '../factories/visitation.factory';
 import { scheduleVisitation } from '../../server/services/visitation.service';
 import { PROPERTY_FILTERS } from '../../server/helpers/filters';
+import { addNextPayment } from '../../server/services/nextPayment.service';
```
andela-hpopoola, line 42: This can be removed as it is not needed here. Rather we should fix why the test is failing rather than adding a quick fix.

test/controllers/transaction.controller.test.js
```diff
@@ -25,6 +25,8 @@ import {
   expectsPaginationToReturnTheRightValues,
   itReturnsAnErrorWhenServiceFails,
 } from '../helpers';
+import { addNextPayment } from '../../server/services/nextPayment.service';
```
andela-hpopoola, line 28: Same comment as above.

test/controllers/user.controller.test.js
```diff
@@ -47,6 +47,8 @@ import {
 import AddressFactory from '../factories/address.factory';
 import VendorFactory from '../factories/vendor.factory';
 import { USER_FILTERS } from '../../server/helpers/filters';
+import { addNextPayment } from '../../server/services/nextPayment.service';
```
andela-hpopoola, line 50: Same comment as above

test/factories/nextPayment.factory.js
```diff
@@ -0,0 +1,14 @@
...
+import mongoose from 'mongoose';
+
+export default new Factory()
+  .option('generateId', false)
+  .after((nextPayment, options) =>
+    options.generateId ? { _id: mongoose.Types.ObjectId(), ...nextPayment } : nextPayment,
+  )
+  .attr('expectedAmount', 50000)
+  .attr('expiresOn', '2090-09-08')
+  .attr('offerId', '5f2b39035a086cfc4b7fa7f6')
```
andela-hpopoola, line 11: Can we use slightly different Ids for all of the ids defined here. This is to ensure that we don't test for unintended bugs and improve the integrity of our tests.

test/services/transaction.service.test.js
```diff
@@ -17,6 +17,8 @@ import { addUser } from '../../server/services/user.service';
 import { createOffer } from '../../server/services/offer.service';
 import { addEnquiry } from '../../server/services/enquiry.service';
 import { USER_ROLE } from '../../server/helpers/constants';
+import { addNextPayment } from '../../server/services/nextPayment.service';
```
andela-hpopoola, line 20: I believe this is also not needed.

test/services/user.service.test.js
```diff
@@ -39,6 +39,8 @@ import { addTransaction } from '../../server/services/transaction.service';
 import { addReferral } from '../../server/services/referral.service';
 import Referral from '../../server/models/referral.model';
 import { USER_ROLE } from '../../server/helpers/constants';
+import { addNextPayment } from '../../server/services/nextPayment.service';
```
andela-hpopoola, line 42: I believe this is also not needed.

andela-hpopoola (changes requested):
Thank you for working on this. I noticed that you added some quick fixes on failing tests involving next payment. Unfortunately, this is bad practises and in the future, it is better to find the exact reason why the test is failing instead of creating a hotfix.

I didn't test the code but I suspect that the issue would be comming from this

```js
await resolvePendingPayment(pendingPayment[0]._id, transactionId);
```

This didn't cater for situations when next payment is not available in the database. Even if the problem is not from here, it is better to troubleshoot the root cause of the failing tests. Remember, other models are expected to work effectively without the next payment (hence, the hotfixes are not needed).

test/services/nextPayment.service.test.js
```diff
@@ -0,0 +1,59 @@
...
+      referenceCode: 'HIG/P/OLP/02/28022021',
+      propertyId: mongoose.Types.ObjectId(),
+      vendorId: mongoose.Types.ObjectId(),
+      userId: mongoose.Types.ObjectId(),
+    },
+    { generateId: true },
+  );
+
+  describe('#generateNextPaymentDate', () => {
+    let countedNextPayments;
```
andela-hpopoola, line 25: This is not needed here since the service is about generating the next payment date not about getting the total of next payment.

server/services/nextpayment.service.js
```diff
@@ -1,8 +1,18 @@
+import mongoose from 'mongoose';
```
andela-hpopoola, line 1: The file name used here is slightly wrong, the `p` should be capitalized. It should be `nextPayment.service.js`

test/services/nextPayment.service.test.js
andela-hpopoola: Rather we should add tests to cater for that the 
- first generated nextPayment is valid
- test for different dates to confirm that the right next payment is generated based on date
- test that different amount paid in transaction results in the right nextPayment on the same date
- test that the right amount is returned if a user has pending next payment
- test with no transaction id
- ensure that a new next payment is not generated when a user has completed his payment or have overpaid in total

Well, I am sure you have some tests in mind too, feel free to add them too. Don't forget that we also need controller test to ensure that all the test cases work as expected (very important)

andela-hpopoola (changes requested):
Thank you for working on this. The sinon fake timer is bae, looks perfect (thanks for getting it right on first trial 😄 ). Let's try to wrap up the remaining tests based on my last comments so that we can wrap up this asap. Thank you.

server/services/nextpayment.service.js
andela-hpopoola: Can you try to fix this as it could cause unintended bugs when accessing the file  🙏🏽

test/services/nextPayment.service.test.js
```diff
@@ -0,0 +1,153 @@
...
+import NextPayment from '../../server/models/nextPayment.model';
+import Offer from '../../server/models/offer.model';
+import Transaction from '../../server/models/transaction.model';
+import { generateNextPaymentDate } from '../../server/services/nextPayment.service';
+import OfferFactory from '../factories/offer.factory';
+import NextPaymentFactory from '../factories/nextPayment.factory';
+import TransactionFactory from '../factories/transaction.factory';
+
+describe('NextPayment Service', () => {
+  const offer = OfferFactory.build(
```
andela-hpopoola, line 12: We can separate numbers in javascript with `_` to make them more readable:

  const offer = OfferFactory.build(
    {
      totalAmountPayable: 10_000_000,
      initialPayment: 50_000,
      periodicPayment: 10_000,
      paymentFrequency: 30,
      ...
    },
  );

test/services/nextPayment.service.test.js
```diff
@@ -0,0 +1,153 @@
...
+    {
+      totalAmountPayable: 100000,
```
andela-hpopoola, line 14: Lets make this 10 million so that we don't easily exhaust the price with a single transaction
```suggestion
      totalAmountPayable: 10_000_000,
```

test/services/nextPayment.service.test.js
```diff
@@ -0,0 +1,153 @@
...
+        await Offer.create(offer);
+        fakeDate = sinon.useFakeTimers({
+          now: new Date('2021-03-20'),
+        });
+        await NextPayment.create(nextPayment);
+        await Transaction.create({ ...transaction, amount: 100000 });
+        await generateNextPaymentDate({ offerId: offer._id, transactionId: transaction._id });
+      });
+
+      afterEach(() => {
```
andela-hpopoola, line 60: Since you have already defined an `afterEach` above, I think this is no longer needed (unless it is giving you some weird errors and using it as a hotfix)

test/services/nextPayment.service.test.js
```diff
@@ -0,0 +1,153 @@
...
+  const nextPayment = NextPaymentFactory.build({ offerId: offer._id }, { generateId: true });
+
+  describe('#generateNextPaymentDate', () => {
+    let fakeDate;
+
+    afterEach(() => {
+      fakeDate.restore();
+    });
+
+    context('when nextPayment and transaction exists', () => {
```
andela-hpopoola, line 49: Let's enclose this in a date to be better grouped together. This will make it easier for us to define the expected expired date as a constant.

Also, instead of when `transaction exists`, we can split this up again and making the first context `when nextPayment exists` and the inner context `with an initial transaction of ${put amount here}` to make the test block more understandable.

test/services/nextPayment.service.test.js
```diff
@@ -0,0 +1,153 @@
...
+      beforeEach(async () => {
+        await Offer.create(offer);
+        fakeDate = sinon.useFakeTimers({
+          now: new Date('2021-03-20'),
```
andela-hpopoola, line 53: Can we use the year `2020` instead of `2021` to make the tests more bug proof. I am currently afraid that if this is not properly stubbed, it would still work as expected for now and start failing when we are no longer in March 2021. However, using year `2020` now would make us spot that asap.

test/services/nextPayment.service.test.js
```diff
@@ -0,0 +1,153 @@
...
+        expect(matchedNextPayments[0].resolved).to.eql(true);
+        expect(matchedNextPayments[0].resolvedViaTransaction).to.eql(true);
+        expect(matchedNextPayments[0].transactionId).to.eql(transaction._id);
+        expect(matchedNextPayments[1].resolved).to.eql(false);
+        expect(matchedNextPayments[1].expectedAmount).to.eql(-50000);
+        expect(matchedNextPayments[1].expiresOn).to.eql(new Date('2021-03-31'));
+      });
+    });
+
+    context('where previous nextPayment does not exist', () => {
```
andela-hpopoola, line 75: This should be when. Please fix for other tests too 🙏🏽 

```suggestion
    context('when previous nextPayment does not exist', () => {
```

test/services/nextPayment.service.test.js
```diff
@@ -0,0 +1,153 @@
...
+        fakeDate.restore();
+      });
+
+      it('adds a new nextPayment', async () => {
+        const matchedNextPayments = await NextPayment.find({ offerId: offer._id });
+        expect(matchedNextPayments[0].resolved).to.eql(true);
+        expect(matchedNextPayments[0].resolvedViaTransaction).to.eql(true);
+        expect(matchedNextPayments[0].transactionId).to.eql(transaction._id);
+        expect(matchedNextPayments[1].resolved).to.eql(false);
+        expect(matchedNextPayments[1].expectedAmount).to.eql(-50000);
```
andela-hpopoola, line 70: We should also add tests when the overall payment has been exceeded to be sure that the next payment is not generated for the offer.

andela-hpopoola (changes requested):
This is a good way forward, thank you for your hard work on this. The tests are getting in shape, we just it to wrap it up and get started with other stories asap. Thank you.

andela-hpopoola (changes requested):
Thank you for this. Your implementation is good but I think we can actually make this better by separating each test by the payment schedule.

Something like this makes the tests more readable imho:

```
 Payment schedule is in the first cycle
   - when a previous nextPayment record exists
   - when a previous nextPayment record doesn't exist
   - when a user pays less than initial payment
   - when user pays more than initial payment
   - when user pays exact amount
   - when a user makes multiple payment
   - when user completes his total payment here

Payment schedule is in the second cycle
   -  when a previous nextPayment record exists (user has not made payment) - paid for the first cycle, returns exact periodic payment 
   -  when a previous nextPayment record doesn't exist (user has not made payment)  - not completed payment for first cycle, returns periodic payment + pending balance
   -  when a previous nextPayment record exists (user has made payment) - paid for the first cycle, returns exact periodic payment 
   -  when a previous nextPayment record doesn't exist (user has made payment)  - not completed payment for first cycle, returns periodic payment + pending balance
   -  when user pays exact amount
   -  when user underpays for the month
   -  when user overpays for the month
   -  when user completes his total payment here

Payment schedule is in the last cycle
   -  * we assume has made some previous payments in transaction for previous cycles
   -  when user pays exact remaining amount for the whole property
   -  when user underpays for the whole property
   -  when a user overpays for the whole property

Payment schedule is past the last cycle
  - the user has outstanding payments (we need to ensure that this is regarded as overdue)
```

The implementation would look something like this

```js
 describe('For the first payment schedule cycle', () => {
     ...
     context('when a previous nextPayment record exists', () => {
          ...
     });
});

 describe('For the second payment schedule cycle', () => {
     ...
     context('when a previous nextPayment record exists', () => {
          ...
     });
});
```

I hope that helps. Thank you for the hardwork in implementing this to this stage. I know it can be very boring working on the same thing over and over again, so thank you for your patience and understanding. You rock. 


Thank you.

test/services/nextPayment.service.test.js
```diff
@@ -49,130 +49,380 @@ describe('NextPayment Service', () => {
...
-        it('does not add a new nextPayment', async () => {
+        it('returns 2 million as expected balance', async () => {
           const matchedNextPayments = await NextPayment.find({ offerId: offer._id });
           expect(matchedNextPayments.length).to.eql(1);
-          expect(matchedNextPayments[0].resolved).to.eql(true);
-          expect(matchedNextPayments[0].resolvedViaTransaction).to.eql(true);
-          expect(matchedNextPayments[0].transactionId).to.eql(transaction._id);
+          expect(matchedNextPayments[0].resolved).to.eql(false);
+          expect(matchedNextPayments[0].expectedAmount).to.eql(2_000_000);
+          expect(matchedNextPayments[0].expiresOn).to.eql(new Date('2020-03-31'));
```
andela-hpopoola, line 87: Since this would always the constant for first cycle, we can add it in a constant

```js
const expiresOnForFirstCycle = new Date('2020-03-31')
```

test/services/nextPayment.service.test.js
```diff
@@ -49,130 +49,380 @@ describe('NextPayment Service', () => {
...
+      context('when a user pays less than initial payment', () => {
         beforeEach(async () => {
-          await Transaction.create({ ...transaction, amount: 10_000_000 });
-          await generateNextPaymentDate({ offerId: offer._id, transactionId: transaction._id });
+          await Transaction.create(transaction);
+          await generateNextPaymentDate({ offerId: offer._id });
         });
 
-        it('does not add a new nextPayment', async () => {
+        it('returns 1 million as expected balance', async () => {
```
andela-hpopoola, line 97:
```suggestion
        it('returns the remaining balance', async () => {
```

test/services/nextPayment.service.test.js
```diff
@@ -49,130 +49,380 @@ describe('NextPayment Service', () => {
...
-      beforeEach(async () => {
-        fakeDate = sinon.useFakeTimers({
-          now: new Date('2020-03-20'),
+      context('when user pays more than initial payment', () => {
+        beforeEach(async () => {
+          await Transaction.create({ ...transaction, amount: 2_500_000 });
+          await generateNextPaymentDate({ offerId: offer._id });
+        });
+
+        it('returns minus 500 thousand as expected balance', async () => {
```
andela-hpopoola, line 112:
```suggestion
        it('returns a negative number as the expected balance', async () => {
```

test/services/nextPayment.service.test.js
```diff
@@ -49,130 +49,380 @@ describe('NextPayment Service', () => {
...
+          beforeEach(async () => {
+            await Transaction.create({
+              ...transaction,
+              amount: 750_000,
+              _id: mongoose.Types.ObjectId(),
+            });
+            await generateNextPaymentDate({ offerId: offer._id });
+          });
+
+          it('returns minus 250 thousand', async () => {
```
andela-hpopoola, line 160: The user has not completed his payment, so there should be no minus in the description 😏 

```suggestion
      it('returns the remaining balance', async () => {
```

test/services/nextPayment.service.test.js
```diff
@@ -49,130 +49,380 @@ describe('NextPayment Service', () => {
...
+
+      context('when user completes his total payment here', () => {
+        beforeEach(async () => {
+          await Transaction.create({ ...transaction, amount: 4_000_000 });
+          await generateNextPaymentDate({ offerId: offer._id });
+        });
+
+        it('returns no next payment', async () => {
+          const matchedNextPayments = await NextPayment.find({ offerId: offer._id });
+          expect(matchedNextPayments.length).to.eql(0);
```
andela-hpopoola, line 180: Awesome 💯

test/services/nextPayment.service.test.js
```diff
@@ -49,130 +49,380 @@ describe('NextPayment Service', () => {
...
-      context('when previous nextPayment does not exist', () => {
-        context('when service runs in second payment cycle', () => {
+    describe('Payment schedule is in the second cycle', () => {
+      beforeEach(async () => {
+        fakeDate = sinon.useFakeTimers({
+          now: new Date('2020-04-05'),
+        });
+      });
+
+      context('when a previous nextPayment record exists (user has not made payment)', () => {
```
andela-hpopoola, line 192: I believe we can break this down into 2 different contexts here.

```js
context('when a previous nextPayment record exists', () => {
   context('when user has not made any payments', () => {
   ...

   context('when user has made previous payments', () => {
   ...
```

test/services/nextPayment.service.test.js
```diff
@@ -49,130 +49,380 @@ describe('NextPayment Service', () => {
...
+        beforeEach(async () => {
+          await NextPayment.create(nextPayment);
+          await generateNextPaymentDate({ offerId: offer._id });
+        });
+
+        it('returns 2.5 million as expected balance', async () => {
```
andela-hpopoola, line 198: Same comments as previous related comments. Please fix for other ones, I will ignore other related comments from now on.

test/services/nextPayment.service.test.js
```diff
@@ -49,130 +49,380 @@ describe('NextPayment Service', () => {
...
+          const matchedNextPayments = await NextPayment.find({ offerId: offer._id });
+          expect(matchedNextPayments.length).to.eql(2);
+          expect(matchedNextPayments[0]._id).to.eql(nextPayment._id);
+          expect(matchedNextPayments[0].resolved).to.eql(true);
+          expect(matchedNextPayments[0].resolvedViaTransaction).to.eql(false);
+          expect(matchedNextPayments[1].resolved).to.eql(false);
+          expect(matchedNextPayments[1].expectedAmount).to.eql(2_500_000);
+          expect(matchedNextPayments[1].expiresOn).to.eql(new Date('2020-04-30'));
```
andela-hpopoola, line 206: Lets put this too in a constant as discussed above. I will ignore similar comments too for this.

andela-hpopoola (changes requested):
Thank you for working on this. Looking almost ready to ship.

The tests look good enough for me. The new test structure looks better and I am having more confidence in the next payment generation.  

However, for some of the tests I wish we could simulate the date two different dates. For example, in the second cycle (and last cycle), I wish we could simulate that the user paid in the first cycle to make simulate the real life scenerio. Please don't waste time on this (you can give a trial though), I might look into that later but for now, this looks good to me. 

The pending stuff is for you to fix the remaining comments so that I can give it one more review and fix some code smells so that we can merge asap.

Thank you.

andela-hpopoola (commented):
Thank you for working on this. This looks good to me but I will still love to made a few changes and cross-check the tests to we cover all edge cases. Also, We will also discuss this during sprint planning, lets take note of all feedback so that we can implement in another PR. 

Thank you.

andela-hpopoola (changes requested):
Please see the updated graphical analysis here. Kindly fix all affected tests as illustrated in the Figma file below. Please take time to look through everything and fix it as explained during the call (let's not waste time on this so that I can accept the PR asap). Thank you.

https://www.figma.com/file/bNPXrVpnXxDDgIbcyK7oT4/Payment-schedule-Analysis?node-id=1%3A429

test/services/nextPayment.service.test.js
```diff
@@ -118,7 +134,12 @@ describe('NextPayment Service', () => {
           expect(matchedNextPayments.length).to.eql(1);
           expect(matchedNextPayments[0].resolved).to.eql(false);
           expect(matchedNextPayments[0].expectedAmount).to.eql(-500_000);
-          expect(matchedNextPayments[0].expiresOn).to.eql(expiresOnForFirstCycle);
+          expect(matchedNextPayments[0].dueDate).to.eql(expiresOnForFirstCycle);
```
andela-hpopoola, line 137: Did you pay attention to the feedback Nnamdi gave us or study the figma file I linked? 

According to the feedback from Nnamdi, the expected Amount should not be negative and the dueDate should be the next date once the payment for a cycle has been completed.

test/services/nextPayment.service.test.js
```diff
@@ -133,7 +154,12 @@ describe('NextPayment Service', () => {
           expect(matchedNextPayments.length).to.eql(1);
           expect(matchedNextPayments[0].resolved).to.eql(false);
           expect(matchedNextPayments[0].expectedAmount).to.eql(0);
-          expect(matchedNextPayments[0].expiresOn).to.eql(expiresOnForFirstCycle);
+          expect(matchedNextPayments[0].dueDate).to.eql(expiresOnForFirstCycle);
```
andela-hpopoola, line 157: Same comment as above.

test/services/nextPayment.service.test.js
```diff
@@ -324,7 +395,12 @@ describe('NextPayment Service', () => {
           expect(matchedNextPayments.length).to.eql(1);
           expect(matchedNextPayments[0].resolved).to.eql(false);
           expect(matchedNextPayments[0].expectedAmount).to.eql(-200_000);
-          expect(matchedNextPayments[0].expiresOn).to.eql(expiresOnForSecondCycle);
+          expect(matchedNextPayments[0].dueDate).to.eql(expiresOnForSecondCycle);
```
andela-hpopoola, line 398: Same comment as above here. It seems like you didn't pay attention to what you were expected to do during standup and you also refused to ask questions. I will stop reviewing this here and let's discuss this offline to understand your thought process.

andela-hpopoola (changes requested):

I am sincerely not sure of what you have done and it seems totally different from what is expected. I would give you an ample opportunity to explain your thought process during standup.

andela-hpopoola (approved):

> The ones that require returning the date of the first cycle when no payment has been made

Based on this message you sent on slack, if this is still a problem, can you add the correct the tests and make sure that the tests have the expected values (i.e return the first cycle when no payment has been made). 

Since the tests would fail, add `.skip` to each of the failing tests to ensure that we are not running the test till we skip all failing tests and the remaining ones passes. This is to serve as a reminder that we have stuffs to fix in the tests.

Even if this is not a problem or you have resolved it, feel free to merge either ways, I will take a fresh look at everything when I am back.

Please focus on the stories partaining to nextPayment. I expect you to finish all the stories pertaining to next Payment this week so that I can review them by weekend when I am back. Thank you.

PR #133, opened 2021-03-01
#177149496 Add features property to property model

andela-hpopoola:
#### What does this PR do?
The purpose of this PR is to add `features` property to the property model

Have the following working
* Add `features` property
* Update `timeAwayFromProperty` to `distance`

GET `/api/v1/property/add`

#177149496

### DIFF
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -3,8 +3,10 @@ import mongoose from 'mongoose';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 const RANGE_SEPARATOR = ':';
+const ARRAY_SEPARATOR = ',';
 
 export const FILTER_TYPE = {
+  ARRAY: 'array',
   BOOLEAN: 'boolean',
   DATE: 'date',
   INTEGER: 'integer',
@@ -52,6 +54,10 @@ export const buildFilterQuery = (allFilters, query) => {
       const [rangeFrom = 0, rangeTo = 0] = query[queryKey].split(RANGE_SEPARATOR);
 
       switch (type) {
+        case FILTER_TYPE.ARRAY:
+          acc.push({ [filterKey]: { $all: query[queryKey].split(ARRAY_SEPARATOR) } });
+          break;
+
         case FILTER_TYPE.BOOLEAN:
           acc.push({ [filterKey]: query[queryKey] === 'true' });
           break;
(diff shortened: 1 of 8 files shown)


server/models/property.model.js
```diff
@@ -132,6 +132,9 @@ const PropertySchema = new mongoose.Schema(
       type: String,
       required: true,
     },
+    features: {
+      type: String,
```
tobieeniafe, line 136: Shouldn't the features be an array of strings?

server/schemas/property.schema.js
```diff
@@ -69,6 +69,7 @@ export const addPropertySchema = Joi.object({
   mapLocation,
   neighborhood,
   mainImage: nonRequiredString('Property main image'),
+  features: nonRequiredString('Features'),
```
tobieeniafe, line 72: This should also be `optionalArray('Features')` since we would be having multiple features

server/schemas/property.schema.js
```diff
@@ -89,6 +90,7 @@ export const updatePropertySchema = Joi.object({
...
```
tobieeniafe, line 93: Same comment as above

test/helpers.js
```diff
@@ -404,51 +404,51 @@ export const expectResponseToContainNecessaryPropertyData = (response, property)
...
     property.neighborhood.schools[0].mapLocation,
   );
   expect(response.neighborhood.shoppingMalls[0].name).to.be.eql(
     property.neighborhood.shoppingMalls[0].name,
   );
-  expect(response.neighborhood.shoppingMalls[0].timeAwayFromProperty).to.be.eql(
-    property.neighborhood.shoppingMalls[0].timeAwayFromProperty,
+  expect(response.neighborhood.shoppingMalls[0].distance).to.be.eql(
+    property.neighborhood.shoppingMalls[0].distance,
   );
```
tobieeniafe, line 452: We should also add the tests for `features`

tobieeniafe (changes requested):
This looks good. I just have concerns about the data type of the property's `features`. Thank you for the update 🙌🏾

server/models/property.model.js
andela-hpopoola: Thank you for catching this. I used a string separated by commas to support filtering via features by default. 

https://github.com/Highrachy/ballers-api/blob/7ddfc0457a2ca15db4c7d6bc547f74e9a9d6ffa4/test/factories/property.factory.js#L13

However, converting this to an `array of strings` sounds like an awesome idea 💡. I have to update the filter type to include an array of strings. Thank you for pointing it out 🙌🏽

server/helpers/filters.js
```diff
@@ -52,6 +54,10 @@ export const buildFilterQuery = (allFilters, query) => {
       const [rangeFrom = 0, rangeTo = 0] = query[queryKey].split(RANGE_SEPARATOR);
 
       switch (type) {
+        case FILTER_TYPE.ARRAY:
+          acc.push({ [filterKey]: { $all: query[queryKey].split(ARRAY_SEPARATOR) } });
```
andela-hpopoola, line 58: I used `$all` here to support multiple entries when filtering array.

https://docs.mongodb.com/manual/tutorial/query-arrays/

tobieeniafe (approved):
The multiple entry implementation for the array filter was 💯. Thanks for the update.

PR #134, opened 2021-03-10
#176955807 allow user to report property

tobieeniafe:
#### What does this PR do?
Add routes to add, resolve & view reported properties

Add routes to add, resolve & view reported properties

`POST /api/v1/report-property/report`
`PUT /api/v1/report-property/resolve`
`GET /api/v1/report-property/all`

#176955807

### DIFF
--- a/test/controllers/reportedProperty.controller.test.js
+++ b/test/controllers/reportedProperty.controller.test.js
@@ -1,0 +1,296 @@
+import { expect, request, sinon } from '../config';
+import ReportedProperty from '../../server/models/reportedProperty.model';
+import UserFactory from '../factories/user.factory';
+import PropertyFactory from '../factories/property.factory';
+import ReportedPropertyFactory from '../factories/reportedProperty.factory';
+import { reportProperty } from '../../server/services/reportedProperty.service';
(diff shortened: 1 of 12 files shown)


test/controllers/reportedProperty.controller.test.js
```diff
@@ -0,0 +1,296 @@
...
+        reason: 'fraudulent vendor',
+        reportedBy: regularUser._id,
+        resolved: {
+          status: false,
+        },
+      },
+      { generateId: true },
+    );
+
+    describe('Report Pagination', () => {
```
andela-hpopoola, line 252: You need to add tests to ensure that the right values are returned here. All necessary `report property` fields like the reason, reported date (i.e createdAt), the property information (we don't want to return the id only), the user information (the tatafo that reported the property - `reportedBy`) and other important information are returned here. 

If possible you can return the information of the admin that resolved the comment (but feel free to ignore if it is proving difficult).

andela-hpopoola (approved):
Thank you for working on this. I left a few comments,  this looks good to me but feel free to merge after resolving the comments. 

Thank you.

PR #135, opened 2021-03-15
#176955812 admin to flag property

tobieeniafe:
#### What does this PR do?
Add route to allow admin flag a reported or suspicious property

Add route to allow admin flag a reported or suspicious property

`POST /api/v1/property/flag`

#176955812

### DIFF
--- a/server/models/property.model.js
+++ b/server/models/property.model.js
@@ -189,19 +189,32 @@ const PropertySchema = new mongoose.Schema(
       required: true,
     },
     flagged: {
-      by: {
-        type: ObjectId,
-      },
-      date: {
-        type: Date,
-      },
       status: {
         type: Boolean,
         default: false,
       },
-      reportId: {
-        type: ObjectId,
-      },
+      case: [
+        {
+          flaggedBy: {
+            type: ObjectId,
+          },
+          flaggedDate: {
+            type: Date,
+          },
+          flaggedReason: {
+            type: String,
+          },
+          unflaggedBy: {
+            type: ObjectId,
+          },
+          unflaggedDate: {
+            type: Date,
+          },
+          unflaggedReason: {
+            type: String,
+          },
+        },
+      ],
     },
   },
   { timestamps: true },
--- a/server/models/property.model.js
+++ b/server/models/property.model.js
@@ -188,6 +188,21 @@ const PropertySchema = new mongoose.Schema(
       type: ObjectId,
       required: true,
     },
+    flagged: {
+      by: {
+        type: ObjectId,
+      },
+      date: {
+        type: Date,
+      },
+      status: {
+        type: Boolean,
+        default: false,
+      },
+      reportId: {
+        type: ObjectId,
+      },
+    },
   },
   { timestamps: true },
 );
--- a/server/routes/property.routes.js
+++ b/server/routes/property.routes.js
@@ -20,6 +20,7 @@ import {
   addImageSchema,
   updateGallerySchema,
   deleteImageSchema,
+  flagPropertySchema,
 } from '../schemas/property.schema';
 import PropertyController from '../controllers/property.controllers';
 
@@ -567,6 +568,35 @@ router.delete(
   isVendor,
   schemaValidation(deleteImageSchema),
   PropertyController.deleteGallery,
+);
+
+/**
+ * @swagger
+ * /property/flag:
+ *   post:
+ *     tags:
+ *       - Property
+ *     description: Allows an admin to flag a property
+ *     produces:
+ *       - application/json
+ *     requestBody:
+ *      content:
+ *        application/json:
+ *          schema:
+ *            $ref: '#/components/schemas/Property'
+ *      description: Allows an admin to flag a property
+ *     responses:
+ *      '200':
+ *        description: Properties flagged
+ *      '500':
+ *       description: Internal server error
+ */
+router.post(
+  '/flag',
+  authenticate,
+  isAdmin,
+  schemaValidation(flagPropertySchema),
+  PropertyController.flagProperty,
 );
 
 module.exports = router;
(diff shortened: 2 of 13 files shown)


server/models/property.model.js
```diff
@@ -188,6 +188,21 @@ const PropertySchema = new mongoose.Schema(
       type: ObjectId,
       required: true,
     },
+    flagged: {
```
andela-hpopoola, line 191: Can we use the same schema pattern used for `banned` users

https://github.com/Highrachy/ballers-api/blob/473f54577d072df37324429a6d9291bf12a53463/server/models/user.model.js#L101-L113

server/routes/property.routes.js
```diff
@@ -569,4 +570,33 @@ router.delete(
...
+ *          schema:
+ *            $ref: '#/components/schemas/Property'
+ *      description: Allows an admin to flag a property
+ *     responses:
+ *      '200':
+ *        description: Properties flagged
+ *      '500':
+ *       description: Internal server error
+ */
+router.post(
```
andela-hpopoola, line 594: We would also need another endpoint to unflag

server/schemas/property.schema.js
```diff
@@ -142,3 +142,9 @@ export const updateGallerySchema = Joi.object({
 export const deleteImageSchema = Joi.object({
   imageId: requiredObjectId('Image id'),
 });
+
+export const flagPropertySchema = Joi.object({
+  propertyId: requiredObjectId('Property id'),
+  reportId: requiredObjectId('Report id'),
```
andela-hpopoola, line 148: I believe `ReportId` should be optional. A property can be flag even without a report. Not sure if the `notes` should be compulsory so that we would know the exact reason why the admin has flagged the property.

server/models/property.model.js
andela-hpopoola, line 191: Don't forget to also add `flagged.status` to property filter to ensure that all flagged properties can be gotten with no issues.

Also, we need to return all flagged information for a property, don't forget to add tests to ensure that it is returned when a we get a property.

andela-hpopoola (changes requested):

However, the purpose of flagging is to stop people from seeing the property (either via search or recommendations) because there is an issue that needs to be resolved with the property. 

To ensure that this works as expected, flagged should support many cases (please take note of each sentence below and you can create them as multiple PRs if you wish)
  - someone complained the property doesn't align with the image uploaded (reason) 
  - when the admin sees this, he flags the property, vendor receives a mail
  -  vendor updates the image and request that admin should unflag
  - when user searches for property, the flagged property is not found
  - admin sees that vendor has made the correction and unflag property (the previous reason why the property was flagged is still saved in the database, the new reason is also saved)
  - user searches for property, user can now see property as it is unflagged
  - user noticed that the number of bedrooms is wrong and reports property again
  -  admin flags again (he can see that the property has been previously flagged and other info) 
  - vendor receives mail again, user cannot see flagged property, vendor fixes, admin unflag ... 
  - the cycle continues

I hope you get the gist. Thank you.

server/services/property.service.js
```diff
@@ -620,3 +633,96 @@ export const deleteGallery = async (imageDetails) => {
...
+    }
+
+    if (report.propertyId.toString() !== property._id.toString()) {
+      throw new ErrorHandler(httpStatus.FORBIDDEN, 'Property does not match report property');
+    }
+
+    const reportInfo = {
+      id: propertyInfo.reportId,
+      resolvedBy: propertyInfo.adminId,
+      notes: propertyInfo.notes || propertyInfo.reason,
```
andela-hpopoola, line 663: It seems like you resolved `reason` and `notes` to notes here. It doesn't feel right, as `property.notes` will be used instead of reason if notes is available in the request. The confusing stuff is that `property.notes` is optional and shouldn't override `property.reason` which is compulsory.

If we want to use `reason` and `notes`, let us resolve them separately as they are different entities else lets stick with `property.reason` alone and remove `propertyInfo.notes`.

test/controllers/property.controller.test.js
```diff
@@ -3665,4 +3751,335 @@ describe('Property Controller', () => {
...
+      reportId: report._id,
+      reason: 'fraudulent property flagged',
+    };
+
+    beforeEach(async () => {
+      await addProperty(property);
+      await reportProperty(report);
+    });
+
+    context('with a valid token & id', () => {
```
andela-hpopoola, line 3793: Can we also add a test to confirm that the report Id is resolved when a property is flagged?

test/controllers/property.controller.test.js
```diff
@@ -3665,4 +3751,335 @@ describe('Property Controller', () => {
...
+              expect(res.body.success).to.be.eql(false);
+              expect(res.body.message).to.be.eql('Validation Error');
+              expect(res.body.error).to.be.eql('"Property id" is not allowed to be empty');
+              expect(sendMailStub.callCount).to.eq(0);
+              done();
+            });
+        });
+      });
+
+      context('when reportId is empty', () => {
```
andela-hpopoola, line 3831: Can we also add tests to confirm that if the reportId is missing the property is flagged appropriately?

andela-hpopoola (approved):
Thank you for helping out with this. I have approved this PR but left a few comments that need to be resolved.

PS: I might miss a few things as I am not 100% back, please bear with me if I point out things here in the future that needs to be resolved.

Overall, this is a great code and you covered all the bases. Thank you.

test/controllers/property.controller.test.js
tobieeniafe: Already done `property.service.test.js` line `220`

andela-hpopoola: No wahala. That fixes this comment.

server/services/property.service.js
tobieeniafe: In this case, it's for resolving the report made. So where there isn't a note specifically for why the report was resolved, we'd use the reason for flagging the property.

PR #136, opened 2021-03-20
#176955705 get all next payment dates for user and vendor

tobieeniafe:
#### What does this PR do?
Get all next payment dates for user and vendor

Get all next payment dates for user and vendor

`GET /api/v1/next-payment/all`

#176955705

### DIFF
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -32,5 +32,5 @@ export default (app) => {
   app.use('/api/v1/area', area);
   app.use('/api/v1/content-property', contentProperty);
   app.use('/api/v1/report-property', reportProperty);
-  app.use('/api/v1/next-payments', nextPayment);
+  app.use('/api/v1/next-payment', nextPayment);
 };
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -14,6 +14,7 @@ import welcome from './welcome.routes';
 import area from './area.routes';
 import contentProperty from './contentProperty.routes';
 import reportProperty from './reportedProperty.routes';
+import nextPayment from './nextPayment.routes';
 
 export default (app) => {
   app.use('/api/v1/docs', swaggerUi.serve, swaggerUi.setup(swaggerSpecs, { explorer: true }));
@@ -31,4 +32,5 @@ export default (app) => {
   app.use('/api/v1/area', area);
   app.use('/api/v1/content-property', contentProperty);
   app.use('/api/v1/report-property', reportProperty);
+  app.use('/api/v1/next-payments', nextPayment);
 };
--- a/server/routes/user.routes.js
+++ b/server/routes/user.routes.js
@@ -22,7 +22,6 @@ import {
   isAdmin,
   isUnverifiedVendor,
   hasValidObjectId,
-  isUserOrVendor,
 } from '../helpers/middleware';
 import UserController from '../controllers/user.controllers';
 import Upload, { UploadController } from '../helpers/uploadImage';
@@ -852,27 +851,6 @@ router.put(
   schemaValidation(unbanUserSchema),
   UserController.unbanUser,
 );
-
-/**
- * @swagger
- * path:
- *  /user/:id:
- *    get:
- *      parameters:
- *        - in: query
- *          name: token
- *          schema:
- *            type: string
- *          description: verifies user access
- *      summary: Gets a user based by its ID
- *      tags: [User]
- *      responses:
- *        '200':
- *          description: Next payments found
- *        '500':
- *          description: Internal server error
- */
-router.get('/next-payments', authenticate, isUserOrVendor, UserController.getNextPayments);
 
 /**
  * @swagger
--- a/server/routes/user.routes.js
+++ b/server/routes/user.routes.js
@@ -22,6 +22,7 @@ import {
   isAdmin,
   isUnverifiedVendor,
   hasValidObjectId,
+  isUserOrVendor,
 } from '../helpers/middleware';
 import UserController from '../controllers/user.controllers';
 import Upload, { UploadController } from '../helpers/uploadImage';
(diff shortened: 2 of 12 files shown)


server/services/nextPayment.service.js
```diff
@@ -113,7 +113,9 @@ export const generateNextPaymentDate = async ({ transactionId = null, offerId })
     await resolvePendingPayment(pendingPayment[0]._id, transactionId);
   }
 
-  if (totalPaid < offer.totalAmountPayable) {
+  if (totalPaid > offer.totalAmountPayable) {
```
andela-hpopoola, line 116: Not sure why you changed this here when we have tests to cater for this? If it is important you change it, please add tests to catch why we missed the bug.

server/services/user.service.js
```diff
@@ -827,3 +828,19 @@ export const unbanUser = async ({ adminId, userId, caseId, reason }) => {
     throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error unbanning user', error);
   }
 };
+
+export const getNextPayments = async (user) => {
+  const activeOffers = await getActiveOffers(user._id).catch((error) => {
```
andela-hpopoola, line 833: Nice concept here but I feel this is a wrong approach to get next payments.

We have the next payments table in the database and this is based on active offers; a user cannot have a next payment record unless they have a transaction/ active offer, they cannot make a transaction unless they have an active offer, so it is safe to assume that next payment would only be generated for active offers.

This is how I will implement mine:

1. For users: Get all next payments in  with the userId that has not been resolved (`resolved is false`) and aggregate it to get the offer details, property details and other necessary information
2. For vendors: Get all next payments in  with the vendorId that has not been resolved (`resolved is false`) and aggregate it to get the offer details, property details and other necessary information
3. For admins: Get all next payments that has `resolved is false`.

Pagination, filters and others can then be included when I have implemented the above functionality.

I hope this helps.

andela-hpopoola (changes requested):
Thank you for working on this. The implementation is still in progress so I left a few comments to unblock you. I hope that helps, please let me know if you need more help so that we can proceed a little bit faster. Thank you.

server/services/nextPayment.service.js
tobieeniafe: This was done in error. The reason the tests are failing. Thanks for pointing it out.

server/services/nextPayment.service.js
```diff
@@ -113,9 +116,73 @@ export const generateNextPaymentDate = async ({ transactionId = null, offerId })
...
+    accountType = {
+      matchKey: 'userId',
+      localField: 'vendorId',
+      as: 'vendorInfo',
+      unwind: '$vendorInfo',
+    };
+  }
+
+  const nextPaymentOptions = [
+    { $match: { [accountType.matchKey]: ObjectId(user._id) } },
```
andela-hpopoola, line 147: We want the admin to get the next payments too, can we update this to accommodate an admin too. Admins should get all unresolved next payments (it is better to add it now than later).

server/routes/user.routes.js
```diff
@@ -873,4 +853,46 @@ router.put(
...
+ *          description: verifies user access
+ *      summary: Gets a user based by its ID
+ *      tags: [User]
+ *      responses:
+ *        '200':
+ *          description: Next payments found
+ *        '500':
+ *          description: Internal server error
+ */
+router.get('/next-payments', authenticate, isUserOrVendor, UserController.getNextPayments);
```
andela-hpopoola, line 875: This should be under nextpayments route `/all` as we have always done for others 😏

test/controllers/user.controller.test.js
```diff
@@ -3126,6 +3129,130 @@ describe('User Controller', () => {
...
+        vendorId: vendorUser._id,
+        resolved: false,
+      });
+      const validNextPayments2 = NextPaymentFactory.buildList(10, {
+        propertyId: vendor2Property._id,
+        userId: regularUser._id,
+        vendorId: vendorUser2._id,
+        resolved: false,
+      });
+      const inValidNextPayments = NextPaymentFactory.buildList(3, {
```
andela-hpopoola, line 3176: These are not invalid, rather they are `resolved`

test/controllers/user.controller.test.js
```diff
@@ -3126,6 +3129,130 @@ describe('User Controller', () => {
         });
       });
     });
+
+    describe('Get All Next Payments', () => {
```
andela-hpopoola, line 3133: Test should be in nextPayments controller

test/controllers/user.controller.test.js
```diff
@@ -3126,6 +3129,130 @@ describe('User Controller', () => {
...
+            ]);
+          });
+
+          itReturnsTheRightPaginationValue({
+            endpoint,
+            method,
+            user: regularUser,
+            useExistingUser: true,
+          });
+
```
andela-hpopoola, line 3217: I have always mentioned that the pagination test does not cover testing that the right things are returned. Can you please keep this in mind to prevent mentioning this over and over again? 

We need to test that the offer information, user information (for both user and vendor), and the right property information are returned.

andela-hpopoola (changes requested):
Thank you for working on this, this looks almost done. I left a few comments on ensuring that an admin can also get the next payments (as discussed in our call) and adding appropriate tests to ensure that the right values are returned. Thank you.

server/routes/index.js
```diff
@@ -31,4 +32,5 @@ export default (app) => {
   app.use('/api/v1/area', area);
   app.use('/api/v1/content-property', contentProperty);
   app.use('/api/v1/report-property', reportProperty);
+  app.use('/api/v1/next-payments', nextPayment);
```
andela-hpopoola, line 35: I noticed that we used singular naming for all routes here. Can we do the same for next payments too 🙏🏽 

```suggestion
  app.use('/api/v1/next-payment', nextPayment);
```

server/services/nextPayment.service.js
```diff
@@ -127,7 +127,7 @@ export const generateNextPaymentDate = async ({ transactionId = null, offerId })
 export const getUnresolvedNextPayments = async (user, { page = 1, limit = 10 } = {}) => {
   let accountType;
 
-  if (user.role === USER_ROLE.VENDOR) {
+  if (user.role === USER_ROLE.VENDOR || user.role === USER_ROLE.ADMIN) {
```
andela-hpopoola, line 130: I believe we can simplify this by making the `user.role` decision come first and use an else statement for this:

```js
...
} else { // this caters for vendor and admin 
...
}

```

andela-hpopoola (approved):
Thank you for working on this. I left a few comments but overall this looks good to me. Thank you for helping with this. Feel free to merge after fixing the comments.

PR #138, opened 2021-03-24
#177431730 return next payment when viewing portfolio

tobieeniafe:
#### What does this PR do?
Return next payment when viewing portfolio

Return next payment when viewing portfolio

`GET /api/v1/property/:id`

#177431730

### DIFF
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -7,7 +7,6 @@ import {
   searchThroughProperties,
   getOneProperty,
   getAvailablePropertyOptions,
-  getAssignedPropertyByOfferId,
   getAssignedProperties,
   addNeighborhood,
   updateNeighborhood,
@@ -112,22 +111,9 @@ const PropertyController = {
       .catch((error) => next(error));
   },
 
-  getAssignedPropertyByOfferId(req, res, next) {
-    const offerId = req.params.id;
-    const { user } = req;
-    getAssignedPropertyByOfferId(offerId, user)
-      .then((property) => {
-        res.status(httpStatus.OK).json({
-          success: true,
-          property,
-        });
-      })
-      .catch((error) => next(error));
-  },
-
   getAssignedProperties(req, res, next) {
-    const userId = req.user._id;
-    getAssignedProperties(userId)
+    const { user } = req;
+    getAssignedProperties(user)
       .then((properties) => {
         res.status(httpStatus.OK).json({
           success: true,
@@ -271,11 +257,7 @@ const PropertyController = {
     const { user } = req;
     getPortfolio(offerId, user)
       .then((portfolio) => {
-        if (portfolio.length > 0) {
-          res.status(httpStatus.OK).json({ success: true, portfolio: portfolio[0] });
-        } else {
-          res.status(httpStatus.NOT_FOUND).json({ success: false, message: 'Portfolio not found' });
-        }
+        res.status(httpStatus.OK).json({ success: true, portfolio });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -112,13 +112,10 @@ const PropertyController = {
   },
 
   getAssignedProperties(req, res, next) {
-    const { user } = req;
-    getAssignedProperties(user)
-      .then((properties) => {
-        res.status(httpStatus.OK).json({
-          success: true,
-          properties,
-        });
+    const { user, query } = req;
+    getAssignedProperties(user, query)
+      .then(({ result, pagination }) => {
+        res.status(httpStatus.OK).json({ success: true, pagination, result });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -19,7 +19,7 @@ import {
   deleteGallery,
   flagProperty,
   unflagProperty,
-  getPortfolio,
+  getOnePortfolio,
 } from '../services/property.service';
 import httpStatus from '../helpers/httpStatus';
 import EMAIL_CONTENT from '../../mailer';
@@ -249,10 +249,10 @@ const PropertyController = {
       .catch((error) => next(error));
   },
 
-  getPortfolio(req, res, next) {
+  getOnePortfolio(req, res, next) {
     const offerId = req.params.id;
     const { user } = req;
-    getPortfolio(offerId, user)
+    getOnePortfolio(offerId, user)
       .then((portfolio) => {
         res.status(httpStatus.OK).json({ success: true, portfolio });
       })
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -7,7 +7,7 @@ import {
   searchThroughProperties,
   getOneProperty,
   getAvailablePropertyOptions,
-  getAssignedProperties,
+  getAllPortfolios,
   addNeighborhood,
   updateNeighborhood,
   deleteNeighborhood,
@@ -111,9 +111,9 @@ const PropertyController = {
       .catch((error) => next(error));
   },
 
-  getAssignedProperties(req, res, next) {
+  getAllPortfolios(req, res, next) {
     const { user, query } = req;
-    getAssignedProperties(user, query)
+    getAllPortfolios(user, query)
       .then(({ result, pagination }) => {
         res.status(httpStatus.OK).json({ success: true, pagination, result });
       })
--- a/server/controllers/property.controllers.js
+++ b/server/controllers/property.controllers.js
@@ -20,6 +20,7 @@ import {
   deleteGallery,
   flagProperty,
   unflagProperty,
+  getPortfolio,
 } from '../services/property.service';
 import httpStatus from '../helpers/httpStatus';
 import EMAIL_CONTENT from '../../mailer';
@@ -87,6 +88,7 @@ const PropertyController = {
       })
       .catch((error) => next(error));
   },
+
   search(req, res, next) {
     const filter = req.locals;
     const userId = req.user._id;
@@ -260,6 +262,20 @@ const PropertyController = {
         sendMail(EMAIL_CONTENT.UNFLAG_PROPERTY, vendor, { contentTop });
 
         res.status(httpStatus.OK).json({ success: true, message: 'Property unflagged', property });
+      })
+      .catch((error) => next(error));
+  },
+
+  getPortfolio(req, res, next) {
+    const offerId = req.params.id;
+    const { user } = req;
+    getPortfolio(offerId, user)
+      .then((portfolio) => {
+        if (portfolio.length > 0) {
+          res.status(httpStatus.OK).json({ success: true, portfolio: portfolio[0] });
+        } else {
+          res.status(httpStatus.NOT_FOUND).json({ success: false, message: 'Portfolio not found' });
+        }
       })
       .catch((error) => next(error));
   },
--- a/server/routes/property.routes.js
+++ b/server/routes/property.routes.js
@@ -6,6 +6,7 @@ import {
   isVendor,
   isVendorOrAdmin,
   hasValidObjectId,
+  isAdminOrUserOrVendor,
 } from '../helpers/middleware';
 import {
   addPropertySchema,
@@ -204,7 +205,7 @@ router.post(
  *   get:
  *     tags:
  *       - Property
- *     description: Get distince state and house types of all properties
+ *     description: Get distinct state and house types of all properties
  *     produces:
  *       - application/json
  *     requestBody:
@@ -212,7 +213,7 @@ router.post(
  *        application/json:
  *          schema:
  *            $ref: '#/components/schemas/Property'
- *      description: Get distince state and house types of all properties
+ *      description: Get distinct state and house types of all properties
  *     responses:
  *      '200':
  *        description: returns object of available fields
@@ -627,6 +628,33 @@ router.put(
   isAdmin,
   schemaValidation(unflagPropertySchema),
   PropertyController.unflagProperty,
+);
+
+/**
+ * @swagger
+ * path:
+ *  /property/portfolio/:id:
+ *    get:
+ *      parameters:
+ *        - in: query
+ *          name: token
+ *          schema:
+ *            type: string
+ *          description: verifies user access
+ *      summary: Get details of a property in portfolio based by its offer id
+ *      tags: [Property]
+ *      responses:
+ *        '200':
+ *          description: Property found
+ *        '500':
+ *          description: Internal server error
+ */
+router.get(
+  '/portfolio/:id',
+  hasValidObjectId,
+  authenticate,
+  isAdminOrUserOrVendor,
+  PropertyController.getPortfolio,
 );
 
 module.exports = router;
--- a/server/routes/property.routes.js
+++ b/server/routes/property.routes.js
@@ -633,7 +633,7 @@ router.get(
   hasValidObjectId,
   authenticate,
   isAdminOrUserOrVendor,
-  PropertyController.getPortfolio,
+  PropertyController.getOnePortfolio,
 );
 
 module.exports = router;
--- a/server/routes/property.routes.js
+++ b/server/routes/property.routes.js
@@ -245,7 +245,7 @@ router.get(
   '/portfolio/all',
   authenticate,
   isAdminOrUserOrVendor,
-  PropertyController.getAssignedProperties,
+  PropertyController.getAllPortfolios,
 );
 
 /**
--- a/server/routes/property.routes.js
+++ b/server/routes/property.routes.js
@@ -7,7 +7,6 @@ import {
   isVendorOrAdmin,
   hasValidObjectId,
   isAdminOrUserOrVendor,
-  isUserOrVendor,
 } from '../helpers/middleware';
 import {
   addPropertySchema,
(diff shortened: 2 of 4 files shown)


test/controllers/property.controller.test.js
```diff
@@ -1250,6 +1252,119 @@ describe('Property Controller', () => {
...
+          request()
+            .get(`/api/v1/property/${property._id}`)
+            .set('authorization', userToken)
+            .send(property)
+            .end((err, res) => {
+              expect(res).to.have.status(200);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body.property._id).to.be.eql(property._id.toString());
+              expect(res.body.property.nextPaymentsInfo.length).to.be.eql(1);
+              expect(res.body.property.nextPaymentsInfo[0].userId).to.be.eql(
```
andela-hpopoola, line 1303: Group 882

From the mockup, we would need to return everything needed to build this component. The most important stuff is:

1. Next Payment Expected Amount - from `NextPayment` model
2. Due Date of Next Payment - from `NextPayment` model
3. Equity Amount - from `offer` model
4. Amount Contributed - From `transaction` model

test/controllers/property.controller.test.js
```diff
@@ -1250,6 +1252,119 @@ describe('Property Controller', () => {
...
+          unresolvedUserNextPayment,
+          resolvedUserNextPayment,
+          unresolvedNextPayment,
+        ]);
+      });
+
+      context('when request is sent by user', () => {
+        it('returns property with nextpayment info', (done) => {
+          request()
+            .get(`/api/v1/property/${property._id}`)
```
andela-hpopoola, line 1295: We have already discussed this that this endpoint is not visible for owned portfolio and that we would need a different endpoint based on offerId for users with multiple properties. 

Can you create a different endpoint (a new service might be needed) and test that important information is returned in this endpoint (something like visitation schedule is not needed).

test/controllers/property.controller.test.js
```diff
@@ -1250,6 +1252,119 @@ describe('Property Controller', () => {
       );
     });
 
+    context('when next payment exists', () => {
```
andela-hpopoola, line 1255: We should also add a test to check what is returned if a user has completed his payment. In this case, there should be no next payment returned but `Equity Amount` and `Amount Contributed` should still be returned.

test/controllers/property.controller.test.js
```diff
@@ -1250,6 +1252,119 @@ describe('Property Controller', () => {
...
+              );
+              expect(res.body.property.nextPaymentsInfo[1]._id).to.be.eql(
+                unresolvedNextPayment._id.toString(),
+              );
+              done();
+            });
+        });
+      });
+
+      context('when request is sent by admin', () => {
```
andela-hpopoola, line 1351: Since the endpoint is based on a portfolio, I believe an admin should be able to view the next payment (might be necessary for support if a customer writes in to complain).

test/controllers/property.controller.test.js
```diff
@@ -1250,6 +1252,119 @@ describe('Property Controller', () => {
...
+        it('returns property with nextpayment info', (done) => {
+          request()
+            .get(`/api/v1/property/${property._id}`)
+            .set('authorization', vendorToken)
+            .send(property)
+            .end((err, res) => {
+              expect(res).to.have.status(200);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body.property._id).to.be.eql(property._id.toString());
+              expect(res.body.property.nextPaymentsInfo.length).to.be.eql(2);
```
andela-hpopoola, line 1327: Since this is now based on the offerId, I believe a vendor, user, and admin should see the same information regarding nextPayments. We only need to confirm that the vendor and user are the owner/custodian of the offerId before they can view it.

server/services/property.service.js
```diff
@@ -180,10 +180,34 @@ export const getOneProperty = async (propertyId, user = {}) => {
...
+    propertyOptions.splice(2, 0, {
+      $lookup: {
+        from: 'nextpayments',
+        let: { propPropertyId: '$_id' },
+        pipeline: [
+          {
+            $match: {
+              $expr: {
+                $and: [
+                  { $eq: ['$propertyId', '$$propPropertyId'] },
```
andela-hpopoola, line 198: When writing the new service, we might need to use the `offerId` to match here. This will enable us to use the same condition for the user, admin, and vendor.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on ensuring that we have a different endpoint for viewing a portfolio so that we can make the matches via the offerId.

server/services/property.service.js
```diff
@@ -726,3 +727,83 @@ export const unflagProperty = async (propertyInfo) => {
...
+  ];
+
+  const portfolio = await Offer.aggregate(portfolioOptions);
+  if (
+    portfolio.length > 0 &&
+    ((user?.role === USER_ROLE.VENDOR &&
+      user?._id.toString() !== portfolio[0].vendorId.toString()) ||
+      (user?.role === USER_ROLE.USER && user?._id.toString() !== portfolio[0].userId.toString()))
+  ) {
+    return [];
```
andela-hpopoola, line 806: Hmmm, I believe it is better to return 'Not Found' when the portfolio is not found and also return an error if the portfolio is viewed by a wrong user or a wrong vendor. 

I don't see a use case where it is better to return an [] since we are trying to retrieve a single portfolio 🤷🏽‍♂️

server/services/property.service.js
```diff
@@ -726,3 +727,83 @@ export const unflagProperty = async (propertyInfo) => {
...
+            },
+          },
+        ],
+        as: 'nextPaymentInfo',
+      },
+    },
+    { $unwind: '$userInfo' },
+    { $unwind: '$vendorInfo' },
+    { $unwind: '$propertyInfo' },
+    // {
```
andela-hpopoola, line 785: If this would be difficult to get, we can get the transaction sum with another query. 

I believe it might be possible to get it here using the last example here but if it is too tricky, it might be better to retrieve the sum separately.

andela-hpopoola (changes requested):
Thank you for making the updates. On a closer look, I discovered that we had similar endpoints, can we ensure we properly deprecate these endpoints and upgrade/reuse some of the services used there. Here are the endpoints:

https://github.com/Highrachy/ballers-api/blob/c86a20bd4dd17b691b2390304fc54c4ca4a4cfc3/server/routes/property.routes.js#L254

https://github.com/Highrachy/ballers-api/blob/c86a20bd4dd17b691b2390304fc54c4ca4a4cfc3/server/routes/property.routes.js#L228

If we are using `property/portfolio/:id` then it makes complete sense to update `property/assigned` to `property/portfolio/all` to cater for all assigned properties in the admin, vendor and user point of view.

I hope this helps. Thank you and Nice work.

server/services/property.service.js
tobieeniafe: This is handled in the controller `property.controller.js` line `277`

server/routes/property.routes.js
```diff
@@ -242,33 +243,12 @@ router.get('/available-options', authenticate, PropertyController.getDistinctPro
  *          description: Internal server error
  */
 router.get(
-  '/assigned/:id',
+  '/portfolio/all',
   authenticate,
-  hasValidObjectId,
-  PropertyController.getAssignedPropertyByOfferId,
+  isUserOrVendor,
```
andela-hpopoola, line 248: I think this endpoint should also support an `Admin` since it is the only way an Admin can have access to any assigned property in a portfolio.

server/services/property.service.js
```diff
@@ -343,27 +341,24 @@ export const getTotalAmountPaidForProperty = async (offerId) =>
     },
   ]);
 
-export const getAssignedPropertyByOfferId = async (offerId, user) => {
-  const total = await getTotalAmountPaidForProperty(offerId);
-  const totalPaid = total.length > 0 ? total[0].totalAmountContributed : 0;
-
-  const offer = await getOffer(offerId, user);
+export const getAssignedProperties = async (user) => {
```
andela-hpopoola, line 344: When getting assigned properties, we would also want to paginate, filter and sort them. Maybe in another PR 🤷🏽‍♂️

test/controllers/property.controller.test.js
```diff
@@ -2263,108 +2264,180 @@ describe('Property Controller', () => {
...
+      await addEnquiry(enquiry);
+      await createOffer(offer);
+      await acceptOffer(acceptOfferData);
       await addTransaction(transaction1);
+      await addTransaction(transaction2);
     });
 
     context('with a valid token & id', () => {
-      it('returns successful payload', (done) => {
+      beforeEach(async () => {
```
andela-hpopoola, line 2356: Can you write the reason why we are faking the date in the context so that we would understand it in the future.

I understand that this is needed for `nextPayment` so it makes more sense to explain (via description in context) the reason we are faking date. For instance, on line 2391, I understand that the tests is checking when the user has completed his payment, but here, it is difficult to understand if the payment is made on the first cycle, last cycle or you had other idea in mind.

test/controllers/property.controller.test.js
```diff
@@ -2263,108 +2264,180 @@ describe('Property Controller', () => {
...
+            .end((err, res) => {
+              expect(res).to.have.status(200);
+              expect(res.body.success).to.be.eql(true);
+              expect(res.body.portfolio._id).to.be.eql(offer._id.toString());
+              expect(res.body.portfolio.propertyInfo._id).to.be.eql(property._id.toString());
+              expect(res.body.portfolio.equityAmount).to.be.eql(offer.totalAmountPayable);
+              expect(res.body.portfolio.amountContributed).to.be.eql(
+                transaction1.amount + transaction2.amount,
+              );
+              expect(res.body.portfolio.outstandingBalance).to.be.eql(
```
andela-hpopoola, line 2380: This is not needed in the backend. Even though it might look harmless to add them, the reason why are removing them is to 
* reduce the payload we are sending for the request
* we have already sent all the needed information to generate it so it is useless to calculate it from the endpoint
* it doesn't make sense trying too hard to conform with the frontend as the frontend is not a fixed design and would change without notice. The only thing we should care about is sending the right information from the endpoint which has been done

test/controllers/property.controller.test.js
```diff
@@ -2263,108 +2264,180 @@ describe('Property Controller', () => {
...
-            expect(res.body.property.offer.propertyInfo._id).to.be.eql(property2._id.toString());
+            expect(res).to.have.status(404);
+            expect(res.body.success).to.be.eql(false);
+            expect(res.body.message).to.be.eql('Portfolio not found');
+            done();
+          });
+      });
+    });
+
+    context('without wrong vendor token', () => {
```
andela-hpopoola, line 2432: We should also test for a wrong user.  A loop can be used for the two tests.

andela-hpopoola (changes requested):
Thank you for working on this. I left more comments on ensuring we get this right. If this is getting more complex than expected, feel free to break it down into multiple PRs but please let me know the ones you are skipping in this PR so that I don't expect them.

server/controllers/property.controllers.js
```diff
@@ -263,6 +248,16 @@ const PropertyController = {
       })
       .catch((error) => next(error));
   },
+
+  getPortfolio(req, res, next) {
```
andela-hpopoola, line 252: Lets make this `getOnePortfolio` so that it can be more descriptive. Same thing with the service name below on line 255,

```suggestion
  getOnePortfolio(req, res, next) {
```

andela-hpopoola (changes requested):
Thank you for making the updates. Code looks good for now, I will review it again when you have written the tests for the new changes and fixed other pending comments. Thank you.

test/controllers/property.controller.test.js
```diff
@@ -2429,18 +2427,20 @@ describe('Property Controller', () => {
...
-          .get(`/api/v1/property/portfolio/${offer._id}`)
-          .set('authorization', newVendorToken)
-          .end((err, res) => {
-            expect(res).to.have.status(404);
-            expect(res.body.success).to.be.eql(false);
-            expect(res.body.message).to.be.eql('Portfolio not found');
-            done();
-          });
-      });
+    context('without wrong token without access', () => {
```
andela-hpopoola, line 2430: Not sure of this your statement means, can you try to rewrite in a more readable format.

```suggestion
    context('with a different user', () => {
```

test/controllers/property.controller.test.js
```diff
@@ -2429,18 +2427,20 @@ describe('Property Controller', () => {
...
+      [...new Array(2)].map((_, index) =>
+        it('returns portfolio', (done) => {
```
andela-hpopoola, line 2432:
```suggestion
        it('doesn't return the portfolio', (done) => {
```

test/controllers/property.controller.test.js
```diff
@@ -2373,13 +2380,9 @@ describe('Property Controller', () => {
               expect(res.body.success).to.be.eql(true);
               expect(res.body.portfolio._id).to.be.eql(offer._id.toString());
               expect(res.body.portfolio.propertyInfo._id).to.be.eql(property._id.toString());
-              expect(res.body.portfolio.equityAmount).to.be.eql(offer.totalAmountPayable);
               expect(res.body.portfolio.amountContributed).to.be.eql(
```
andela-hpopoola, line 2383: We need a test here to ensure that the right `totalAmountPayable` is returned. We removed `equityAmount` because it is already returned in the offer, however, we need to write a test to ensure that this is truly available in the offer.

server/controllers/property.controllers.js
```diff
@@ -7,7 +7,6 @@ import {
   searchThroughProperties,
   getOneProperty,
   getAvailablePropertyOptions,
-  getAssignedPropertyByOfferId,
   getAssignedProperties,
```
andela-hpopoola, line 10: Lets update this to `getAllPortfolios` so that we can always relate it with `getOnePortfolio` in the future.

server/services/property.service.js
```diff
@@ -342,27 +341,19 @@ export const getTotalAmountPaidForProperty = async (offerId) =>
...
-  const offer = await getOffer(offerId, user);
-
-  return {
-    totalPaid,
-    offer: offer[0],
-  };
-};
+export const getAssignedProperties = async (user, { page = 1, limit = 10, ...query } = {}) => {
+  const matchKey = user.role === USER_ROLE.USER ? 'userId' : 'vendorId';
+  const filterQuery = buildFilterQuery(OFFER_FILTERS, query);
```
andela-hpopoola, line 346: We might need to create a PORTFOLIO_FILTERS in the future but I think this works for now.

test/controllers/property.controller.test.js
```diff
@@ -2263,117 +2271,182 @@ describe('Property Controller', () => {
...
+            expect(res.body.portfolio.amountContributed).to.be.eql(
+              transaction1.amount + transaction2.amount + transaction3.amount,
             );
+            expect(res.body.portfolio.nextPaymentInfo.length).to.be.eql(0);
             done();
           });
       });
     });
 
-    context('when no transaction has been made', () => {
```
andela-hpopoola, line 2356: Can we also add a test to check that the right values are returned when no transaction has been made.

test/controllers/property.controller.test.js
```diff
@@ -2455,27 +2529,31 @@ describe('Property Controller', () => {
     });
```
andela-hpopoola, line 2530: Can we update the description here to reflect that we are getting all portfolios. Please update it to `Get All Portfolios`.

https://github.com/Highrachy/ballers-api/blob/82a918d4dd3e2961d3765f2104e05696f5eae5d6/test/controllers/property.controller.test.js#L2488

andela-hpopoola: Also, it looks like pagination tests are missing here. Can you please add the generic pagination tests just to be safe that it works when we have more than the limit?

test/controllers/property.controller.test.js
```diff
@@ -2263,117 +2271,182 @@ describe('Property Controller', () => {
...
-  describe('Load transaction sum and info of assigned property', () => {
-    const property1 = PropertyFactory.build(
-      {
-        addedBy: vendorUser._id,
-        updatedBy: vendorUser._id,
-      },
-      { generateId: true },
-    );
-    const property2 = PropertyFactory.build(
+  describe('View User Portfolio', () => {
```
andela-hpopoola, line 2274:
```suggestion
  describe('Get One Portfolio', () => {
```

andela-hpopoola (changes requested):
Thank you for helping out with this. I believe this is almost completed, I left a few comments just to be sure we are not missing anything and to fix a few nits. Thank you.

test/controllers/property.controller.test.js
tobieeniafe: Currently working on this, looking for the best way to generate them with the `OFFER_STATUS`

test/controllers/property.controller.test.js
```diff
@@ -2355,46 +2355,79 @@ describe('Property Controller', () => {
...
               done();
             });
-        }),
-      );
+        });
+      });
     });
 
-    context('when user has completed payment', () => {
+    context.skip('when no payment has been made', () => {
```
andela-hpopoola, line 2424: I believe this is related to the pending next payments fix 😼

andela-hpopoola (approved):
Thank you for working on this. Apart from the skipped test, this looks good to go. Feel free to merge 👍🏽

test/controllers/property.controller.test.js
tobieeniafe: Haha yes, it 😅. It's in the second cycle and no payment has been made, so we expect the date to be the expiry of the first cycle, but the sum of both expected payments.

PR #140, opened 2021-03-29
#177431721 cron job endpoint for next payment

tobieeniafe:
#### What does this PR do?
Add route for cronjob to process unresolved expired next payment

Add route for cronjob to process unresolved expired next payment

`GET /api/v1/next-payment/resolved-expired`

#177431721

### DIFF
--- a/server/controllers/nextPayment.controllers.js
+++ b/server/controllers/nextPayment.controllers.js
@@ -2,7 +2,7 @@ import {
   getUnresolvedNextPayments,
   sendReminder,
   generateNextPaymentDate,
-  cronjob,
+  resolvedExpiredNextPayments,
 } from '../services/nextPayment.service';
 import httpStatus from '../helpers/httpStatus';
 import EMAIL_CONTENT from '../../mailer';
@@ -58,13 +58,13 @@ const NextPaymentController = {
       .catch((error) => next(error));
   },
 
-  cronjob(req, res, next) {
-    cronjob()
-      .then((numberResolved) => {
+  resolvedExpiredNextPayments(req, res, next) {
+    resolvedExpiredNextPayments()
+      .then((resolvedNextPayments) => {
         res.status(httpStatus.OK).json({
           success: true,
-          message: `${numberResolved} next ${
-            numberResolved === 1 ? 'payment' : 'payments'
+          message: `${resolvedNextPayments.length} next ${
+            resolvedNextPayments.length === 1 ? 'payment' : 'payments'
           } processed`,
         });
       })
--- a/server/controllers/nextPayment.controllers.js
+++ b/server/controllers/nextPayment.controllers.js
@@ -2,7 +2,7 @@ import {
   getUnresolvedNextPayments,
   sendReminder,
   generateNextPaymentDate,
-  resolvedExpiredNextPayments,
+  resolveExpiredNextPayments,
 } from '../services/nextPayment.service';
 import httpStatus from '../helpers/httpStatus';
 import EMAIL_CONTENT from '../../mailer';
@@ -58,8 +58,8 @@ const NextPaymentController = {
       .catch((error) => next(error));
   },
 
-  resolvedExpiredNextPayments(req, res, next) {
-    resolvedExpiredNextPayments()
+  resolveExpiredNextPayments(req, res, next) {
+    resolveExpiredNextPayments()
       .then((resolvedNextPayments) => {
         res.status(httpStatus.OK).json({
           success: true,
--- a/server/controllers/nextPayment.controllers.js
+++ b/server/controllers/nextPayment.controllers.js
@@ -2,6 +2,7 @@ import {
   getUnresolvedNextPayments,
   sendReminder,
   generateNextPaymentDate,
+  cronjob,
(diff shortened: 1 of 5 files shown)


server/controllers/nextPayment.controllers.js
```diff
@@ -2,6 +2,7 @@ import {
   getUnresolvedNextPayments,
   sendReminder,
   generateNextPaymentDate,
+  cronjob,
```
andela-hpopoola, line 5: Can we use another name 🙏🏽 ? This service and endpoint has nothing to do with cron job. Cron job would only be used to run the service, it has nothing to do with the process itself. 

Also, we might also decide to run it manually without cron job if the need arise.

server/routes/nextPayment.routes.js
```diff
@@ -84,4 +84,27 @@ router.get(
   NextPaymentController.recalculateNextPayment,
 );
 
+/**
+ * @swagger
+ * /next-payment/cronjob:
```
andela-hpopoola, line 89: The essence of the endpoint is to resolve expired next payments. 
```suggestion
 * /next-payment/resolved-expired:
```

server/services/nextPayment.service.js
```diff
@@ -241,3 +241,19 @@ export const sendReminder = async () => {
...
+    { $match: { resolved: false } },
+    { $match: { expiresOn: { $lte: today } } },
+  ]);
+
+  await unResolvedNextPayments.forEach(async (nextPayment) => {
+    await resolvePendingPayment(nextPayment._id);
+    await generateNextPaymentDate({ offerId: nextPayment.offerId });
+  });
+
+  return unResolvedNextPayments.length;
```
andela-hpopoola, line 258: Let's return the unresolvedNextPayments instead of the length so that if we run this manually we get also get the unresolved fields.

```suggestion
  return unResolvedNextPayments;
```

server/controllers/nextPayment.controllers.js
```diff
@@ -56,6 +57,19 @@ const NextPaymentController = {
...
   },
+
+  cronjob(req, res, next) {
+    cronjob()
+      .then((numberResolved) => {
+        res.status(httpStatus.OK).json({
+          success: true,
+          message: `${numberResolved} next ${
+            numberResolved === 1 ? 'payment' : 'payments'
+          } processed`,
```
andela-hpopoola, line 68: Lets return the unresolved next payments here too, it might be useful in the future.

server/services/nextPayment.service.js
```diff
@@ -241,3 +241,19 @@ export const sendReminder = async () => {
...
+
+export const cronjob = async () => {
+  const today = new Date();
+
+  const unResolvedNextPayments = await NextPayment.aggregate([
+    { $match: { resolved: false } },
+    { $match: { expiresOn: { $lte: today } } },
+  ]);
+
+  await unResolvedNextPayments.forEach(async (nextPayment) => {
```
andela-hpopoola, line 253: To ensure that this loop can handle so much record at the same time, can you add a test that tries to resolve 100 expired next payments.

andela-hpopoola (changes requested):
Thank you for working on this. This looks good and almost ready to ship when you write the test.

server/routes/nextPayment.routes.js
```diff
@@ -105,6 +105,6 @@ router.get(
  *      '500':
  *       description: Internal server error
  */
-router.get('/cronjob', NextPaymentController.cronjob);
+router.get('/resolved-expired', NextPaymentController.resolvedExpiredNextPayments);
```
andela-hpopoola, line 108:
```suggestion
router.get('/resolve-expired', NextPaymentController.resolvedExpiredNextPayments);
```

server/services/nextPayment.service.js
```diff
@@ -242,18 +242,19 @@ export const sendReminder = async () => {
   ]);
 };
 
-export const cronjob = async () => {
+export const resolvedExpiredNextPayments = async () => {
```
andela-hpopoola, line 245:
```suggestion
export const resolveExpiredNextPayments = async () => {
```

test/controllers/nextPayment.controller.test.js
```diff
@@ -488,4 +488,98 @@ describe('Next Payments Controller', () => {
...
+    beforeEach(async () => {
+      fakeDate = sinon.useFakeTimers({
+        now: new Date('2020-03-24'),
+      });
+
+      await Offer.insertMany([...unresolvedNextPaymentsOffers, ...resolvedNextPaymentsOffers]);
+      await NextPayment.insertMany([...unresolvedNextPayments, ...resolvedNextPayments]);
+    });
+
+    context('when next payments are processed properly', () => {
```
andela-hpopoola, line 572: This is good. However, we would need to add more tests for the following scenarios:
1. When only 1 next payment is processed
2. When no next payment is processed (we need to be sure of what is returned here too)

andela-hpopoola (approved):
This looks good to me. The pending stuff is to add more tests; I am approving this with the confidence that the you would add the 2 pending test alighted in my comment below. Thank you and Nice Work.

PR #142, opened 2021-04-05
#177621927 Fix payment schedules date

andela-hpopoola:
#### What does this PR do?
The purpose of this PR is to fix the payment schedule bug

* Fix the previous payment schedule bug
* Fix all skipped tests

This can be tested by running the tests `npm test`. If the all tests passes, it means the fix is valid 😅 .


tobieeniafe (approved):
This was interesting to review. Thanks for the fix. We'd probably talk about this after tomorrow's stand-up.

PR #147, opened 2021-04-08
#177678968 remove `/transaction/add` & `/transaction/update` endpoints

tobieeniafe:
#### What does this PR do?
Remove `/transaction/add` & `/transaction/update` endpoints.

Remove `/transaction/add` & `/transaction/update` endpoints. Since resolving an offline payment request has replaced the flow.

NA

#177678968


server/controllers/transaction.controllers.js
```diff
@@ -10,18 +9,6 @@ import {
...
-    addTransaction({ ...newTransaction, addedBy: user._id, updatedBy: user._id })
-      .then((transaction) => {
-        res
-          .status(httpStatus.CREATED)
-          .json({ success: true, message: 'Transaction added', transaction });
-      })
-      .catch((error) => next(error));
-  },
-
   update(req, res, next) {
```
andela-hpopoola, line 12: I believe this endpoint too should be removed. I think it is better to restrict the admin's access to the transaction table.

andela-hpopoola (changes requested):
Thank you for working on this. I left a comment on removing the `update` endpoint too. I will approve once that has been done. Thank you.

andela-hpopoola (approved):
Thank you for helping out here. Removing code is fun, I wish writing codes could be as fast as this too 😛 . Feel free to merge when ready.

PR #154, opened 2021-04-15
#177728140 add remittance to transaction model

tobieeniafe:
#### What does this PR do?
Add remittance to transaction model

Add remittance to transaction model

`PUT /api/v1/transaction/remittance`

#177728140

### DIFF
--- a/mailer/index.js
+++ b/mailer/index.js
@@ -79,6 +79,11 @@ const EMAIL_CONTENT = {
     contentBottom: `Click on the link below to accept their invitation`,
     buttonText: 'Join BALLERS',
   },
+  REMITTANCE_PAID: {
+    subject: 'Remittance Paid',
+    contentTop:
+      'You have received [amount] as payment on one of your properties. Visit your dashboard for more information.',
+  },
   RESCHEDULE_VISIT: {
     subject: 'Visitation Rescheduled',
     contentTop:
--- a/mailer/index.js
+++ b/mailer/index.js
@@ -80,7 +80,7 @@ const EMAIL_CONTENT = {
     buttonText: 'Join BALLERS',
   },
   REMITTANCE_PAID: {
-    subject: 'Remittance Paid',
+    subject: 'Remittance Information',
     contentTop:
       'You have received [amount] as payment on one of your properties. Visit your dashboard for more information.',
   },
--- a/server/controllers/transaction.controllers.js
+++ b/server/controllers/transaction.controllers.js
@@ -4,6 +4,7 @@ import {
   getReferralRewards,
   getContributionRewards,
   getOneTransaction,
+  addRemittance,
 } from '../services/transaction.service';
 import httpStatus from '../helpers/httpStatus';
 
@@ -67,6 +68,16 @@ const TransactionController = {
             .status(httpStatus.NOT_FOUND)
             .json({ success: false, message: 'Transaction not found' });
         }
+      })
+      .catch((error) => next(error));
+  },
+
+  addRemittance(req, res, next) {
+    const remittanceInfo = req.locals;
+    const adminId = req.user._id;
+    addRemittance({ ...remittanceInfo, adminId })
+      .then((transaction) => {
+        res.status(httpStatus.OK).json({ success: true, message: 'Remittance added', transaction });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/transaction.controllers.js
+++ b/server/controllers/transaction.controllers.js
@@ -7,6 +7,8 @@ import {
   addRemittance,
 } from '../services/transaction.service';
 import httpStatus from '../helpers/httpStatus';
+import EMAIL_CONTENT from '../../mailer';
+import { sendMail } from '../services/mailer.service';
 
 const TransactionController = {
   getAll(req, res, next) {
@@ -76,7 +78,10 @@ const TransactionController = {
     const remittanceInfo = req.locals;
     const adminId = req.user._id;
     addRemittance({ ...remittanceInfo, adminId })
-      .then((transaction) => {
+      .then(({ transaction, user }) => {
+        const contentTop = `You have received ${transaction.remittance.amount} as payment on one of your properties. Visit your dashboard for more information.`;
+        sendMail(EMAIL_CONTENT.REMITTANCE_PAID, user, { contentTop });
+
         res.status(httpStatus.OK).json({ success: true, message: 'Remittance added', transaction });
       })
       .catch((error) => next(error));
--- a/server/helpers/projectedSchemaInfo.js
+++ b/server/helpers/projectedSchemaInfo.js
@@ -1,3 +1,5 @@
+import { USER_ROLE } from './constants';
+
 export const PROJECTED_PROPERTY_INFO = {
   name: 1,
   titleDocument: 1,
@@ -73,6 +75,9 @@ export const NON_PROJECTED_USER_INFO = (infoType) => {
   };
 };
 
-export const EXCLUDED_TRANSACTION_INFO = {
-  remittance: 0,
+export const GET_EXCLUDED_TRANSACTION_INFO = (userRole) => {
+  if (userRole === USER_ROLE.USER) {
+    return { remittance: 0 };
+  }
+  return {};
 };
--- a/server/helpers/projectedSchemaInfo.js
+++ b/server/helpers/projectedSchemaInfo.js
@@ -71,13 +71,10 @@ export const NON_PROJECTED_USER_INFO = (infoType) => {
     [`${infoType}.vendor.verification`]: 0,
     [`${infoType}.vendor.entity`]: 0,
     [`${infoType}.vendor.redanNumber`]: 0,
+    [`${infoType}.vendor.remittancePercentage`]: 0,
     [`${infoType}.vendor.taxCertificate`]: 0,
   };
 };
 
-export const GET_EXCLUDED_TRANSACTION_INFO = (userRole) => {
-  if (userRole === USER_ROLE.USER) {
-    return { remittance: 0 };
-  }
-  return {};
-};
+export const getExcludedTransactionInfo = (userRole) =>
+  userRole === USER_ROLE.USER ? { remittance: 0 } : {};
--- a/server/helpers/projectedSchemaInfo.js
+++ b/server/helpers/projectedSchemaInfo.js
@@ -72,3 +72,7 @@ export const NON_PROJECTED_USER_INFO = (infoType) => {
     [`${infoType}.vendor.taxCertificate`]: 0,
   };
 };
+
+export const EXCLUDED_TRANSACTION_INFO = {
+  remittance: 0,
+};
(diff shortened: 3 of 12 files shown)


server/services/transaction.service.js
```diff
@@ -270,3 +272,34 @@ export const getOneTransaction = async (transactionId, user) => {
...
+export const addRemittance = async (remittanceInfo) => {
+  const transaction = await getTransactionById(remittanceInfo.transactionId).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (!transaction) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Invalid transaction');
+  }
+
+  const vendor = await getUserById(transaction.vendorId).catch((error) => {
```
andela-hpopoola, line 285:
```suggestion
  const user = await getUserById(transaction.vendorId).catch((error) => {
```

andela-hpopoola (changes requested):
Thank you for working on this. I think this looks good to me pending the time you work on the test. 

Also, can we ensure we update the `getAllTransaction` tests to ensure that remittance is not returned when a user is getting transactions (we don't want the user to know about remittance) but this should be available when an admin/vendor get all transactions?

server/services/transaction.service.js
```diff
@@ -122,6 +122,10 @@ export const getAllTransactions = async (user, page = 1, limit = 10) => {
 
   if (user.role === USER_ROLE.USER) {
     transactionOptions.unshift({ $match: { userId: ObjectId(user._id) } });
+    transactionOptions[7].$project = {
```
andela-hpopoola, line 125: This looks like a hacky way to exclude transaction info. A better alternative is to create it as a function that returns `EXCLUDED_TRANSACTION_INFO` if the current user has a user role else it returns null or an empty object.

Something like this:

```js
{
      $project: {
        ...NON_PROJECTED_USER_INFO('userInfo'),
        ...NON_PROJECTED_USER_INFO('vendorInfo'),
        ...EXCLUDED_PROPERTY_INFO,
        ...getExcludedTransactionInfo(userRole),
      },
    },
```

server/services/transaction.service.js
```diff
@@ -177,6 +177,10 @@ export const getUserTransactionsByProperty = async (propertyId, user, page = 1,
 
   if (user.role === USER_ROLE.USER) {
     transactionOptions.unshift({ $match: { userId: ObjectId(user._id) } });
+    transactionOptions[6].$project = {
```
andela-hpopoola, line 180: Same comment as above.

server/services/transaction.service.js
```diff
@@ -267,6 +265,10 @@ export const getOneTransaction = async (transactionId, user) => {
 
   if (user.role === USER_ROLE.USER) {
     transactionOptions.unshift({ $match: { userId: ObjectId(user._id) } });
+    transactionOptions[8].$project = {
```
andela-hpopoola, line 268: Still same comment as above

test/controllers/transaction.controller.test.js
```diff
@@ -207,6 +221,7 @@ describe('Transaction Controller', () => {
               expect(res.body.result[0].offerId).to.be.eql(
                 vendor2Transactions[0].offerId.toString(),
               );
+              expect(res.body.result[0]).to.have.property('remittance');
```
andela-hpopoola, line 224: Let's test the values of remittance returned instead of checking if the `remittance` property exists here. It is better to check for the amount, by, percentage e.t.c since they are available in the object.

test/controllers/transaction.controller.test.js
```diff
@@ -776,4 +843,103 @@ describe('Transaction Controller', () => {
...
+        propertyId: property._id,
+        offerId: offer._id,
+        vendorId: vendorUser._id,
+        addedBy: adminUser._id,
+        updatedBy: adminUser._id,
+      },
+      { generateId: true },
+    );
+
+    const data = {
```
andela-hpopoola, line 863: Can we add a test with the `percentage` key defined in `data` since percentage is optional.

server/controllers/transaction.controllers.js
```diff
@@ -70,6 +71,16 @@ const TransactionController = {
       })
       .catch((error) => next(error));
   },
+
+  addRemittance(req, res, next) {
+    const remittanceInfo = req.locals;
+    const adminId = req.user._id;
+    addRemittance({ ...remittanceInfo, adminId })
```
andela-hpopoola, line 78: A mail should be sent to the vendor that he has been paid.

test/controllers/transaction.controller.test.js
```diff
@@ -776,4 +843,103 @@ describe('Transaction Controller', () => {
...
+      transactionId: transaction._id,
+      amount: 1_000_000,
+      date: '2021-02-12',
+    };
+
+    beforeEach(async () => {
+      await Transaction.create(transaction);
+    });
+
+    context('with valid token and data', () => {
```
andela-hpopoola, line 873: Can we also create a user with a different remittance (from 5) to ensure that remittance from the user model works.

server/schemas/transaction.schema.js
```diff
@@ -0,0 +1,11 @@
+import Joi from '@hapi/joi';
+import { requiredObjectId, requiredNumber, requiredDate, optionalNumber } from './helper.schema';
+
+const addRemittanceSchema = Joi.object({
+  transactionId: requiredObjectId('Transaction Id'),
+  amount: requiredNumber('Amount'),
```
andela-hpopoola, line 6: On a closer look, we shouldn't ask for the amount from the endpoint, we should rather calculate this from the supplied percentage. That way the amount will always be consistent with the percentage and would fix the bug of remitting more than the amount paid for the transaction.

server/models/user.model.js
```diff
@@ -299,6 +299,10 @@ const UserSchema = new mongoose.Schema(
           updatedAt: Date,
         },
       ],
+      remittancePercentage: {
+        type: Number,
+        default: 5,
```
andela-hpopoola, line 304: Don't forget, you need to confirm this default from Nnamdi.

server/services/transaction.service.js
```diff
@@ -265,8 +265,43 @@ export const getOneTransaction = async (transactionId, user) => {
...
+  const user = await getUserById(transaction.vendorId).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  try {
+    return Transaction.findByIdAndUpdate(
+      transaction.id,
+      {
+        $set: {
+          'remittance.amount': remittanceInfo.amount,
```
andela-hpopoola, line 296: This should be calculated from the percentage and rounded to the nearest whole number for consistency sake

andela-hpopoola (changes requested):
Thank you for making the changes. I left some more comments on ensuring that we cover more edge cases concerning this endpoint. Thank you and Nice work.

test/controllers/transaction.controller.test.js
```diff
@@ -105,6 +107,12 @@ describe('Transaction Controller', () => {
         vendorId: vendorUser._id,
         addedBy: adminUser._id,
         updatedBy: adminUser._id,
+        remittance: {
```
andela-hpopoola, line 110: Lastly, In another PR (maybe next week), we would need an endpoint to enable an admin to update a vendor's remittance percentage.

mailer/index.js
```diff
@@ -79,6 +79,11 @@ const EMAIL_CONTENT = {
     contentBottom: `Click on the link below to accept their invitation`,
     buttonText: 'Join BALLERS',
   },
+  REMITTANCE_PAID: {
+    subject: 'Remittance Paid',
```
andela-hpopoola, line 83: This will surely not be remittance paid. Please work with content team to get a nice email to send out on this.

server/helpers/projectedSchemaInfo.js
```diff
@@ -73,6 +75,9 @@ export const NON_PROJECTED_USER_INFO = (infoType) => {
   };
 };
 
-export const EXCLUDED_TRANSACTION_INFO = {
-  remittance: 0,
+export const GET_EXCLUDED_TRANSACTION_INFO = (userRole) => {
```
andela-hpopoola, line 78: If this is a function, then the naming convention shouldn't be in ALL CAPS.

```js
export const getExcludedTransactionInfo = (userRole) => {
```

server/helpers/projectedSchemaInfo.js
```diff
@@ -73,6 +75,9 @@ export const NON_PROJECTED_USER_INFO = (infoType) => {
...
+  if (userRole === USER_ROLE.USER) {
```
andela-hpopoola, line 79: This can be written as one liner:

```js
export const getExcludedTransactionInfo = (userRole) => userRole === USER_ROLE.USER ? { remittance: 0 } : {};
````

File: server/schemas/transaction.schema.js [hunk 16]
```diff
@@ -1,9 +1,8 @@
 import Joi from '@hapi/joi';
-import { requiredObjectId, requiredNumber, requiredDate, optionalNumber } from './helper.schema';
+import { requiredObjectId, requiredDate, optionalNumber } from './helper.schema';
 
 const addRemittanceSchema = Joi.object({
   transactionId: requiredObjectId('Transaction Id'),
-  amount: requiredNumber('Amount'),
   date: requiredDate('Date'),
   percentage: optionalNumber('Percentage'),
```
andela-hpopoola, line 7: Can we create an optionalPercentage function to cater for this so that we don't accept values over 100% or below 0. Lets also add test for less than 0 and greater than 100 🙏🏽

server/routes/transaction.routes.js
```diff
@@ -142,4 +145,35 @@ router.get(
...
+ *      description: Adds remittance for a specific transaction
+ *     responses:
+ *      '200':
+ *        description: Remittance added
+ *      '400':
+ *        description: Error adding remittance
+ *      '500':
+ *       description: Internal server error
+ */
+router.put(
```
andela-hpopoola, line 171: Can we change this to `post` since we are using the endpoint to add a new remittance 🙏🏽 . This makes the endpoint consistent even if we need to move the remittance model into its own model. It also saves us the stress if we wish to create an endpoint to update remittance. 

I know it is confusing but I feel post is the right thing to use here even though we are updating the transaction model.

```suggestion
router.post(
```

andela-hpopoola (approved):
Thank you for working on this. I left a few comments to perfect this. Feel free to merge once you fix them. Thank you.

server/models/user.model.js
```diff
@@ -299,6 +299,10 @@ const UserSchema = new mongoose.Schema(
           updatedAt: Date,
         },
       ],
+      remittancePercentage: {
```
andela-hpopoola, line 302: I believe this is sensitive information. Can we ensure that this is also not leaked to users too. Feel free to implement this in a new PR if it is tricky.

PR #156, opened 2021-04-18
#177638823 add rate limit to API

tobieeniafe:
#### What does this PR do?
Add rate limit to API endpoints

Add rate limit to API endpoints

NA

#177638823

### DIFF
--- a/server/config/rateLimit.js
+++ b/server/config/rateLimit.js
@@ -1,0 +1,13 @@
+import rateLimit from 'express-rate-limit';
+
+const timeFrameInHours = 24;
+const maxNoOfRequests = 5000;
+
+const rateLimiter = rateLimit({
+  windowMs: timeFrameInHours * 60 * 60 * 1000,
+  max: maxNoOfRequests,
+  message: `You have exceeded the ${maxNoOfRequests} requests in ${timeFrameInHours}hrs limit!`,
+  headers: true,
+});
+
+export default rateLimiter;
--- a/server/config/rateLimit.js
+++ b/server/config/rateLimit.js
@@ -1,12 +1,12 @@
 import rateLimit from 'express-rate-limit';
 
-const timeFrameInHours = 24;
-const maxNoOfRequests = 5000;
+const timeFrameInHours = 1;
+const maxNoOfRequests = 500;
 
 const rateLimiter = rateLimit({
   windowMs: timeFrameInHours * 60 * 60 * 1000,
   max: maxNoOfRequests,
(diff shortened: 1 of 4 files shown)


server/index.js
```diff
@@ -7,13 +7,15 @@ import { PORT, startDBConnection } from './config';
...
 import { errorMiddleware } from './helpers/errorHandler';
+import rateLimiter from './config/rateLimit';
 
 const app = express();
 
 app.use(cors());
 app.use(bodyParser.urlencoded({ extended: false }));
 app.use(bodyParser.json());
 app.use(morgan('combined', { stream: winston.stream.write }));
+app.use(rateLimiter);
```
andela-hpopoola, line 18: Can we ensure that this is not used for test environment?

server/config/rateLimit.js
```diff
@@ -0,0 +1,13 @@
+import rateLimit from 'express-rate-limit';
+
+const timeFrameInHours = 24;
```
andela-hpopoola, line 3: I think 1 hour is better, 24 hours seems to much to me.

```suggestion
const timeFrameInHours = 1;
```

server/config/rateLimit.js
```diff
@@ -0,0 +1,13 @@
...
+const maxNoOfRequests = 5000;
```
andela-hpopoola, line 4: Lets start with 500 requests for 1 hour. We can always update this as time goes.

```suggestion
const maxNoOfRequests = 500;
```

server/config/rateLimit.js
```diff
@@ -0,0 +1,13 @@
...
+
+const rateLimiter = rateLimit({
+  windowMs: timeFrameInHours * 60 * 60 * 1000,
+  max: maxNoOfRequests,
+  message: `You have exceeded the ${maxNoOfRequests} requests in ${timeFrameInHours}hrs limit!`,
```
andela-hpopoola, line 9:
```suggestion
  message: `You have exceeded the ${maxNoOfRequests} requests in ${timeFrameInHours} hour(s) limit!`,
```

andela-hpopoola (approved):
Thank you for working on this. This is a good starting point, we might need to change the number of requests later. Nice work.

server/index.js
andela-hpopoola: Also, can we make the limits applicable to the API endpoints only. Something like:

```suggestion
app.use('/api', rateLimiter);
```
Please test to ensure that this works as expected.

PR #159, opened 2021-04-20
#177755633 make expiry date close end of day

tobieeniafe:
#### What does this PR do?
Make expiry date close end of day

Make expiry date close end of day

`POST /api/v1/offer/create`
`PUT /api/v1/offer/reactivate`

#177755633

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "cloudinary": "^1.22.0",
     "cors": "^2.8.5",
     "date-fns": "^2.11.0",
+    "date-fns-tz": "^1.1.4",
     "dotenv": "^8.2.0",
     "ejs": "^3.0.2",
     "express": "^4.17.1",
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "cloudinary": "^1.22.0",
     "cors": "^2.8.5",
     "date-fns": "^2.11.0",
-    "date-fns-tz": "^1.1.4",
     "dotenv": "^8.2.0",
     "ejs": "^3.0.2",
     "express": "^4.17.1",
--- a/test/controllers/offer.controller.test.js
+++ b/test/controllers/offer.controller.test.js
@@ -150,6 +150,7 @@ describe('Offer Controller', () => {
               expect(res.body.offer).to.have.property('userInfo');
               expect(res.body.offer.vendorInfo._id).to.be.eql(vendorUser._id.toString());
               expect(res.body.offer.propertyInfo._id).to.be.eql(newProperty._id.toString());
+              expect(res.body.offer.expires).to.have.string('22:59:59.999');
               expect(res.body.offer.referenceCode).to.be.eql(
                 `HIG/LVE/OLM/01/${getTodaysDateShortCode()}`,
               );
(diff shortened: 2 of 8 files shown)


test/controllers/offer.controller.test.js
```diff
@@ -155,6 +155,7 @@ describe('Offer Controller', () => {
               expect(res.body.offer).to.have.property('userInfo');
               expect(res.body.offer.vendorInfo._id).to.be.eql(vendorUser._id.toString());
               expect(res.body.offer.propertyInfo._id).to.be.eql(newProperty._id.toString());
+              expect(res.body.offer.expires).to.have.string('22:59:59.999');
```
andela-hpopoola, line 158: Not currently sure if the converted date format (expires) is the issue but the end of the day should be `23:59:59.999` since we have 24 hours. 

Likely issue is that `expires` is supposed to be in GMT+1 format to correspond with our own date.

https://date-fns.org/v2.21.1/docs/endOfDay
```suggestion
              expect(res.body.offer.expires).to.have.string('23:59:59.999');
```

test/controllers/offer.controller.test.js
```diff
@@ -1457,6 +1458,166 @@ describe('Offer Controller', () => {
         });
       });
     });
+
+    describe('Reactivate Expired Offer', () => {
```
andela-hpopoola, line 1462: Not sure why we had so much tests for `Reactivated Expired Offer` 🤷🏽‍♂️

andela-hpopoola (changes requested):
Thank you for working on this. Once you have fixed the tests, can you update the following fields too in this models to use endOfDay:

- offer.model.js - `initialPaymentDate` (you fixed `expires` in this PR)
- enquiry.model.js - `investmentStartDate`
- nextPayment.model.js - `expiresOn`

Thank you.

test/controllers/offer.controller.test.js
tobieeniafe: This is actually correct but in a tricky way. The expiry date is set with respect to the user's timezone. So the issue with this is, the server's clock is at GMT and the machine is at GMT+1. So the date returned is correct (for the server) with respect to the user's clock.

tobieeniafe:
> Thank you for working on this. Once you have fixed the tests, can you update the following fields too in this models to use endOfDay:
> 
> * offer.model.js - `initialPaymentDate` (you fixed `expires` in this PR)
> * enquiry.model.js - `investmentStartDate`
> * nextPayment.model.js - `expiresOn`
> 
> Thank you.

I don't think  `enquiry.model.js` - `investmentStartDate` should be here

test/controllers/offer.controller.test.js
andela-hpopoola: >  the server's clock is at GMT and the machine is at GMT+1

I am sorry, I have to disagree with this ☝🏽 as the test is run on Travis CI which I believe serves as the server and machine (or whatever you mean by this). 

The returned end-of-day time in the test is `22:59:59.999`, the response doesn't have any reference to the GMT or GMT+1, so I believe this is wrong. Even if it will return `22:59:59.999`, then the GMT should be specified else the time should be in GMT+1 to ensure consistency when it is being viewed from different parts of the world.

package.json
```diff
@@ -66,6 +66,7 @@
     "cloudinary": "^1.22.0",
     "cors": "^2.8.5",
     "date-fns": "^2.11.0",
+    "date-fns-tz": "^1.1.4",
```
andela-hpopoola, line 69: Where is this used 🤷🏽‍♂️ ?

andela-hpopoola (commented):

> I don't think enquiry.model.js - investmentStartDate should be here

I added `investmentStartDate` to make it consistent with `initialPaymentDate` since the `investmentStartDate` is used to generate the `initialPaymentDate`.

package.json
tobieeniafe: I added it to try to manipulate the timezones, but it didn't work out. Forgot to remove it. My bad 🤦🏾‍♂️

andela-hpopoola:
@tobieeniafe I have created a few hotfixes for the dates and I think they are manageable for now. I have 1 pending edge case:
- the current end of the day uses the server's the end of day and not for Nigeria

However, every other thing relating to the date issues we had looks and works as expected. I will keep on testing it to be able to unravel more edge cases.

This works for all timezones for now and I think it is safe to merge. Thank you.

PR #160, opened 2021-04-22
#177810751 create notification for users

tobieeniafe:
#### What does this PR do?
1. Add service to generate notifications.
2. Create route to mark a notification as read

1. Add service to generate notifications.
2. Create route to mark a notification as read

`GET /api/v1/notification/all`
`GET /api/v1/notification/:id`
`PUT /api/v1/notification/:id`

#177810751

### DIFF
--- a/notifications/index.js
+++ b/notifications/index.js
@@ -1,6 +1,111 @@
 import { NOTIFICATION_TYPE } from '../server/helpers/constants';
 
 const NOTIFICATIONS = {
+  ACTIVATE_YOUR_ACCOUNT: {
+    description: 'Complete your registration by verifying your email',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  BAN_USER: {
+    description: 'Your account has been locked for security reasons.',
+    type: NOTIFICATION_TYPE.DANGER,
+    url: '/',
+  },
+  CANCEL_VISIT: {
+    description: 'Visitation to property cancelled',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  CERTIFY_VENDOR: {
+    description: 'Your account has been certified',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  CHANGED_PASSWORD: {
+    description: 'Password changed sucessfully',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  FLAG_PROPERTY: {
+    description: 'Your prperty has been flagged',
+    type: NOTIFICATION_TYPE.DANGER,
+    url: '/',
+  },
+  OFFER_CREATED: {
+    description: 'You have received an offer',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  OFFER_REACTIVATED: {
+    description: 'Your offer has been reactivated',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  OFFER_RESPONSE_VENDOR: {
+    description: 'Your offer has been accepted',
+    type: NOTIFICATION_TYPE.SUCCESS,
+    url: '/',
+  },
+  OFFER_RESPONSE_USER: {
+    description: 'Congratulations on signing your offer',
+    type: NOTIFICATION_TYPE.SUCCESS,
+    url: '/',
+  },
+  PAYMENT_REMINDER: {
+    description: 'You have a pending payment due',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  RAISE_CONCERN: {
+    description: 'A comment has been raised on your offer',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  RAISE_COMMENT_FOR_OFFLINE_PAYMENT: {
+    description: 'A comment has been raised on your offline payment',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  REMITTANCE_PAID: {
+    description: 'You have received payment on one of your properties',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  RESCHEDULE_VISIT: {
+    description: 'Your visitation has been rescheduled',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  RESOLVE_CONCERN: {
+    description: 'Your raised concern has been resolved.',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  RESOLVE_COMMENT_FOR_OFFLINE_PAYMENT: {
+    description: 'Your raised comment has been resolved.',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  SCHEDULE_VISIT: {
+    description: 'A new propery visit has been requested',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  UNBAN_USER: {
+    description: 'Your account has been unlocked',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  UNFLAG_PROPERTY: {
+    description: 'Your property has been unflagged',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  VERIFY_VENDOR: {
+    description: 'Your account has been verified',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
   WELCOME_MESSAGE: {
     description: `Congratulations you're officially a BALLer.`,
     type: NOTIFICATION_TYPE.INFO,
--- a/notifications/index.js
+++ b/notifications/index.js
@@ -1,0 +1,11 @@
+import { NOTIFICATION_TYPE } from '../server/helpers/constants';
+
+const NOTIFICATIONS = {
+  WELCOME_MESSAGE: {
+    description: `Congratulations you're officially a BALLer.`,
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+};
+
+export default NOTIFICATIONS;
--- a/notifications/index.js
+++ b/notifications/index.js
@@ -2,7 +2,7 @@ import { NOTIFICATION_TYPE } from '../server/helpers/constants';
 
 const NOTIFICATIONS = {
   ACTIVATE_YOUR_ACCOUNT: {
-    description: 'Complete your registration by verifying your email',
+    description: 'Your account has been successfully created',
     type: NOTIFICATION_TYPE.INFO,
     url: '/',
   },
@@ -12,13 +12,13 @@ const NOTIFICATIONS = {
     url: '/',
   },
   CANCEL_VISIT: {
-    description: 'Visitation to property cancelled',
-    type: NOTIFICATION_TYPE.INFO,
+    description: 'The visitation to [property] property has been cancelled',
+    type: NOTIFICATION_TYPE.DANGER,
     url: '/',
   },
   CERTIFY_VENDOR: {
     description: 'Your account has been certified',
-    type: NOTIFICATION_TYPE.INFO,
+    type: NOTIFICATION_TYPE.SUCCESS,
     url: '/',
   },
   CHANGED_PASSWORD: {
@@ -27,7 +27,7 @@ const NOTIFICATIONS = {
     url: '/',
   },
   FLAG_PROPERTY: {
-    description: 'Your prperty has been flagged',
+    description: 'Your property has been flagged',
     type: NOTIFICATION_TYPE.DANGER,
     url: '/',
   },
@@ -103,7 +103,7 @@ const NOTIFICATIONS = {
   },
   VERIFY_VENDOR: {
     description: 'Your account has been verified',
-    type: NOTIFICATION_TYPE.INFO,
+    type: NOTIFICATION_TYPE.SUCCESS,
     url: '/',
   },
   WELCOME_MESSAGE: {
--- a/server/controllers/notification.controllers.js
+++ b/server/controllers/notification.controllers.js
@@ -34,10 +34,12 @@ const NotificationController = {
   markAllNotificationsAsRead(req, res, next) {
     const userId = req.user._id;
     markAllNotificationsAsRead(userId)
-      .then((notifications) => {
-        res
-          .status(httpStatus.OK)
-          .json({ success: true, message: 'All notifications marked as read', notifications });
+      .then(({ noOfMarkedNotifications }) => {
+        res.status(httpStatus.OK).json({
+          success: true,
+          message: 'All notifications marked as read',
+          noOfMarkedNotifications,
+        });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/notification.controllers.js
+++ b/server/controllers/notification.controllers.js
@@ -10,10 +10,10 @@ const NotificationController = {
     const notificationId = req.params.id;
     const userId = req.user._id;
     markNotificationAsRead({ userId, notificationId })
-      .then((user) => {
+      .then((notification) => {
         res
           .status(httpStatus.OK)
-          .json({ success: true, message: 'Notification marked as read', user });
+          .json({ success: true, message: 'Notification marked as read', notification });
       })
       .catch((error) => next(error));
   },
--- a/server/controllers/notification.controllers.js
+++ b/server/controllers/notification.controllers.js
@@ -2,6 +2,7 @@ import {
   markNotificationAsRead,
   getAllNotifications,
   getOneNotification,
+  markAllNotificationsAsRead,
 } from '../services/notification.service';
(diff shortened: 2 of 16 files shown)


server/services/user.service.js
```diff
@@ -849,3 +849,41 @@ export const updateRemittancePercentage = async (remittanceInfo) => {
...
+      user._id,
+      { $set: { notifications: newNotification } },
+      { new: true, fields: '-password' },
+    );
+  } catch (error) {
+    throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error adding notification', error);
+  }
+};
+
+export const markNotificationAsRead = async ({ userId, notificationId }) => {
```
andela-hpopoola, line 879: Awesome. We would also need another endpoint to mark all notifications as READ.

server/services/user.service.js
```diff
@@ -849,3 +849,41 @@ export const updateRemittancePercentage = async (remittanceInfo) => {
...
 };
+
+export const addNotification = async (userId, notification) => {
+  const user = await getUserById(userId);
+
+  const existingNotifications = user.notifications;
+
+  const newNotification = [
+    {
+      description: notification.description,
```
andela-hpopoola, line 860: Adding a notification does not pass through any validation, so it makes sense to use the optional chaining (?.) and defaults when any of the values are missing.

server/services/user.service.js
```diff
@@ -849,3 +849,41 @@ export const updateRemittancePercentage = async (remittanceInfo) => {
...
+      type: notification.type,
```
andela-hpopoola, line 861: Notification type should be one of the following (DANGER is the default):

```js
export const NOTIFICATION_TYPE = {
  DANGER: 0,
  CONTENT: 1,
  INFO: 2,
  SUCCESS: 3,
};
```

server/services/user.service.js
andela-hpopoola, line 860: A similar file like this should be created for notifications (maybe under helpers folder)

https://github.com/Highrachy/ballers-api/blob/ft-create-notifications-177810751/mailer/index.js

This would enable us to define notifications in one place and know the content for each notification.

```js
const NOTIFICATIONS = {
  ACTIVATE_YOUR_ACCOUNT: {
    description:
      'Your ballers.ng account has been successfully created.',
    type: NOTIFICATION_TYPE.INFO,
     url: '/login'
  },
...
```

server/routes/user.routes.js
```diff
@@ -911,4 +911,37 @@ router.put(
   UserController.updateRemittancePercentage,
 );
 
+/**
+ * @swagger
+ * /user/notification/:id:
```
andela-hpopoola, line 916: On a closer look, I noticed that it would be best to create a different model for notifications instead of embedding it in the user model. My reasons are:

1. I noticed that the notifications of an ADMIN could be more than expected, so it is better to load only the needed notifications, maybe the first 20 to save bandwidth and make the application faster.
2. Saving it in its own model gives us the luxury of adding pagination, filters and sorting which is very important for long time users.

andela-hpopoola (changes requested):
Thank you for working on this. I did some quick analysis and noticed that it is better to create a different model for notifications. Please let me know if you have any further questions on this. Thank you.

server/services/user.service.js
andela-hpopoola: We would also need another endpoint to read a single Notification.

notifications/index.js
```diff
@@ -0,0 +1,11 @@
+import { NOTIFICATION_TYPE } from '../server/helpers/constants';
+
+const NOTIFICATIONS = {
+  WELCOME_MESSAGE: {
```
andela-hpopoola, line 4: In this part, you are meant to add all of the contents in the mailer field here. I believe you should be able to deduce the description from the content top. Moreover, we can always update it later even if the contents are invalid.

server/controllers/notification.controllers.js
```diff
@@ -0,0 +1,51 @@
...
+        res.status(httpStatus.OK).json({
+          success: true,
+          pagination,
+          result,
+        });
+      })
+      .catch((error) => next(error));
+  },
+
+  getOneNotification(req, res, next) {
```
andela-hpopoola, line 34: I thought we have agreed that this is no longer needed. Feel free to remove this (if you like).

server/controllers/notification.controllers.js
```diff
@@ -0,0 +1,51 @@
+import {
+  markNotificationAsRead,
+  getAllNotifications,
+  getOneNotification,
+} from '../services/notification.service';
+import httpStatus from '../helpers/httpStatus';
+
+const NotificationController = {
+  markNotificationAsRead(req, res, next) {
```
andela-hpopoola, line 9: Don't forget to add the `Mark All Notifications As Read` endpoint

server/models/notification.model.js
```diff
@@ -0,0 +1,64 @@
...
+ *          - userId
+ *          - description
+ *          - type
+ *          - URL
+ *          - status
+ *        properties:
+ *          userId:
+ *            type: string
+ *          description:
+ *            type: number
```
andela-hpopoola, line 20: I think this should be a `string` or do we want to enter the description in binary numbers 😂

server/helpers/filters.js
```diff
@@ -323,3 +323,10 @@ export const AREA_FILTERS = {
   createdAt: { type: FILTER_TYPE.DATE },
   state: { type: FILTER_TYPE.STRING },
 };
+
+export const NOTIFICATION_FILTERS = {
+  createdAt: { type: FILTER_TYPE.DATE },
+  status: { type: FILTER_TYPE.INTEGER },
```
andela-hpopoola, line 329: Since this is just 2 digits, let's change it to `read` and make it `BOOLEAN` for better understanding.

server/models/notification.model.js
```diff
@@ -0,0 +1,64 @@
...
+    },
+    description: {
+      type: String,
+      required: true,
+    },
+    type: {
+      type: Number,
+      default: NOTIFICATION_TYPE.DANGER,
+    },
+    URL: {
```
andela-hpopoola, line 51: Let's make this be in lowercase to conform with other values 🙏🏽 

```suggestion
    url: {
```

server/services/user.service.js
```diff
@@ -113,6 +114,12 @@ export const addUser = async (user) => {
     const password = await hashPassword(user.password);
     const savedUser = await new User({ ...user, password, referralCode, role }).save();
 
+    await createNotification({ userId: savedUser._id, description: 'WELCOME_MESSAGE' }).catch(
```
andela-hpopoola, line 117: This looks weird as you didn't make the best use of the constants 🤷🏽‍♂️ 

Rather, lets get the content of the notification we want to add and pass it into the function directly:

```js
await createNotification(NOTIFICATIONS.WELCOME_MESSAGE, savedUser._id)
```

See the implementation of mailer service here:

https://github.com/Highrachy/ballers-api/blob/00513ad263ab539443397e76fd85c2663eaf795c/server/services/mailer.service.js#L53

server/services/notification.service.js
```diff
@@ -0,0 +1,71 @@
...
+
+export const getOneNotification = async ({ notificationId, userId }) => {
+  return Notification.findOne({ _id: notificationId, userId });
+};
+
+export const createNotification = async (data) => {
+  try {
+    const notification = await new Notification({
+      userId: data.userId,
+      description: NOTIFICATIONS[data.description].description,
```
andela-hpopoola, line 20: This implementation looks confusing. If you pass the notification object as advised above, you should be able to use it as 

I didn't use the question mark here because I am destructuring from the parameter.

```js
export const createNotification = async ({ description, type, url}, userId) => {
 try {
    const notification = await new Notification({ description, type, userId, url}).save()
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few more comments on ensuring that we are moving in the right direction. Nice work 👍🏽

server/services/user.service.js
andela-hpopoola: Can you resolve this only when you have implemented the requested endpoint 😏

server/controllers/notification.controllers.js
tobieeniafe: But you mentioned that you needed it so you could be able to mark a notification as read. Or did I miss something?

server/services/notification.service.js
```diff
@@ -31,7 +25,7 @@ export const markNotificationAsRead = async ({ userId, notificationId }) => {
   try {
     return Notification.findOneAndUpdate(
       { _id: notificationId, userId },
-      { $set: { status: NOTIFICATION_STATUS.READ } },
+      { $set: { read: NOTIFICATION_STATUS.READ } },
```
andela-hpopoola, line 28: I feel this is more readable as true or false
```suggestion
    { $set: { read: true } },
```

server/services/notification.service.js
```diff
@@ -69,3 +63,19 @@ export const getAllNotifications = async (user, { page = 1, limit = 10, ...query
   const result = notifications[0].data;
   return { pagination, result };
 };
+
+export const markAllNotificationsAsRead = async (userId) => {
+  try {
+    return Notification.updateMany(
+      { userId },
+      { $set: { read: NOTIFICATION_STATUS.READ } },
```
andela-hpopoola, line 71: Same comment as above

server/helpers/constants.js
```diff
@@ -83,3 +83,15 @@ export const NEIGHBORHOOD_STEPS = [
...
 ];
+
+export const NOTIFICATION_TYPE = {
+  DANGER: 0,
+  CONTENT: 1,
+  INFO: 2,
+  SUCCESS: 3,
+};
+
+export const NOTIFICATION_STATUS = {
```
andela-hpopoola, line 94: This can be removed so that we can use the true and false directly.

notifications/index.js
```diff
@@ -0,0 +1,116 @@
+import { NOTIFICATION_TYPE } from '../server/helpers/constants';
+
+const NOTIFICATIONS = {
+  ACTIVATE_YOUR_ACCOUNT: {
+    description: 'Complete your registration by verifying your email',
```
andela-hpopoola, line 5:
```suggestion
    description: 'Your account has been successfully created',
```

notifications/index.js
```diff
@@ -0,0 +1,116 @@
...
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  BAN_USER: {
+    description: 'Your account has been locked for security reasons.',
+    type: NOTIFICATION_TYPE.DANGER,
+    url: '/',
+  },
+  CANCEL_VISIT: {
+    description: 'Visitation to property cancelled',
```
andela-hpopoola, line 15: We need to add the property name here.

```suggestion
    description: 'The visitation to [property] property has been cancelled',
```

notifications/index.js
```diff
@@ -0,0 +1,116 @@
...
+    type: NOTIFICATION_TYPE.INFO,
```
andela-hpopoola, line 16:
```suggestion
    type: NOTIFICATION_TYPE.DANGER,
```

notifications/index.js
```diff
@@ -0,0 +1,116 @@
...
+    url: '/',
+  },
+  CERTIFY_VENDOR: {
+    description: 'Your account has been certified',
+    type: NOTIFICATION_TYPE.INFO,
```
andela-hpopoola, line 21:
```suggestion
    type: NOTIFICATION_TYPE.SUCCESS,
```

notifications/index.js
```diff
@@ -0,0 +1,116 @@
...
+    url: '/',
+  },
+  CHANGED_PASSWORD: {
+    description: 'Password changed sucessfully',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  FLAG_PROPERTY: {
+    description: 'Your prperty has been flagged',
```
andela-hpopoola, line 30:
```suggestion
    description: 'Your property has been flagged',
```

notifications/index.js
```diff
@@ -0,0 +1,116 @@
...
+    url: '/',
+  },
+  UNFLAG_PROPERTY: {
+    description: 'Your property has been unflagged',
+    type: NOTIFICATION_TYPE.INFO,
+    url: '/',
+  },
+  VERIFY_VENDOR: {
+    description: 'Your account has been verified',
+    type: NOTIFICATION_TYPE.INFO,
```
andela-hpopoola, line 106:
```suggestion
    type: NOTIFICATION_TYPE.SUCCESS,
```

server/controllers/notification.controllers.js
andela-hpopoola: I think you missed the last part where we agreed that this might not be needed when we checked out GitHub's notification system. No worries, feel free to leave it.

andela-hpopoola (changes requested):
Thank you for working on this. I left more comments but this should be good to go when you have written the tests.

One last thing though, can we aggregate unread notifications to the returned values  in the `user/who-am-i` endpoint. This is to make it easier for users to know when they have a new notification without visiting the get all notifications endpoint. This can be done in a different PR though. Thank you.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on perfecting this so this should be ready to merge once you fixed the raised issues.

test/controllers/notification.controller.test.js
```diff
@@ -0,0 +1,371 @@
...
+          });
+      });
+    });
+
+    context('with token without valid access', () => {
+      [...new Array(3)].map((_, index) =>
+        it('returns successful payload', (done) => {
+          request()
+            [method](endpoint)
+            .set('authorization', [adminToken, vendorToken, editorToken][index])
```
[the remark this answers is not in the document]
tobieeniafe: The test is testing if other users can mark the notification of another user as read. In this case, trying to mark the notification of the `regularUser` as read.

test/services/user.service.test.js
```diff
@@ -166,7 +169,9 @@ describe('User Service', () => {
 
       it('adds a new user', async () => {
         const currentCountedUsers = await User.countDocuments({});
+        const currentCountedNotifications = await Notification.countDocuments({});
         expect(currentCountedUsers).to.eql(countedUsers + 1);
+        expect(currentCountedNotifications).to.eql(countedNotifications + 1);
```
andela-hpopoola, line 174: This looks good but counting the notification is prone to lots of errors as we might be saving the wrong notification or saving the notification for the wrong user. 

It would be better to retrieve the saved notification from the database and compare the value that was saved. At least, we need to confirm the following to be sure that the right notification was saved:
1. userId
2. type
3. description

You should import the notification file and compare the values with this. The best way is to create a test function (like pagination test functions) that we would pass the notification type and userId. The test function would get the last notification and compare every necessary detail that was passed.

andela-hpopoola (changes requested):
Thank you for the hard work. I left one comment on ensuring we tested the saved notification the right way. I believe it should take too much time, however, feel free to move to other stories if you are stuck on it instead of wasting time on it for the rest of the day. We can discuss the implementation during stand up tomorrow. The next stories are 
- loading unread notifications for the `who-am-i` endpoint
- Add all notifications to the right services.

Thank you.

test/helpers.js
```diff
@@ -591,3 +592,19 @@ export const itReturnAllResultsWhenAnUnknownFilterIsUsed = ({
...
+
+export const expectNewNotificationToBeAdded = (userId, notification) => {
+  let userLatestNotification;
+
+  beforeEach(async () => {
+    userLatestNotification = await Notification.find({ userId }).sort({ _id: -1 }).limit(1);
+  });
+
+  context('when new notification is added', () => {
+    it('adds a new notification for user', async () => {
```
andela-hpopoola, line 604: This doesn't seem like an async function

```suggestion
    it('adds a new notification for user', () => {
```

test/services/user.service.test.js
```diff
@@ -169,11 +168,11 @@ describe('User Service', () => {
 
       it('adds a new user', async () => {
         const currentCountedUsers = await User.countDocuments({});
-        const currentCountedNotifications = await Notification.countDocuments({});
         expect(currentCountedUsers).to.eql(countedUsers + 1);
-        expect(currentCountedNotifications).to.eql(countedNotifications + 1);
       });
 
+      expectNewNotificationToBeAdded(user._id, NOTIFICATIONS.WELCOME_MESSAGE);
```
andela-hpopoola, line 174: Awesome. Implemented as discussed 🙌🏽

andela-hpopoola (approved):
Thank you for working on this. This looks good to go 💯

server/services/notification.service.js
```diff
@@ -0,0 +1,76 @@
...
+  const notifications = await Notification.aggregate(notificationOptions);
+  const total = getPaginationTotal(notifications);
+  const pagination = generatePagination(page, limit, total);
+  const result = notifications[0].data;
+  return { pagination, result };
+};
+
+export const markAllNotificationsAsRead = async (userId) => {
+  try {
+    return Notification.updateMany({ userId }, { $set: { read: true } }, { new: true });
```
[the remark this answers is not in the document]
andela-hpopoola: Any luck with this ☝🏽

PR #162, opened 2021-04-29
#177903406 add notifications to services mail is sent

tobieeniafe:
#### What does this PR do?
Add notifications to services mail is sent

Add notifications to services mail is sent

`npm run test`

#177903406

### DIFF
--- a/server/services/nextPayment.service.js
+++ b/server/services/nextPayment.service.js
@@ -10,6 +10,8 @@ import { getTotalTransactionByOfferId } from './transaction.service';
 import { NON_PROJECTED_USER_INFO } from '../helpers/projectedSchemaInfo';
 import { USER_ROLE } from '../helpers/constants';
 import { generatePagination, generateFacetData, getPaginationTotal } from '../helpers/pagination';
+import { createNotification } from './notification.service';
+import NOTIFICATIONS from '../../notifications/index';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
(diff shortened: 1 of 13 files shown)


server/services/nextPayment.service.js
```diff
@@ -221,6 +223,14 @@ export const sendReminder = async () => {
     { $unwind: '$userInfo' },
     { $unwind: '$propertyInfo' },
   ]);
+
+  await Promise.all(
+    reminders.map(async (reminder) => {
+      await createNotification(NOTIFICATIONS.PAYMENT_REMINDER, reminder.userId);
```
andela-hpopoola, line 229: It looks like the services test for this is missing? Can we add it 🙏🏽

server/services/visitation.service.js
```diff
@@ -156,6 +159,12 @@ export const processVisitation = async ({ user, visitationInfo, action }) => {
       { new: true },
     );
 
+    if (action === PROCESS_VISITATION_ACTION.RESCHEDULE) {
```
andela-hpopoola, line 162: I dont think we need to define another if statement here, we can add the notification to the first statement on line 149.

andela-hpopoola (approved):
Thank you for working on this. I left a few comments but overall, this looks good to me. It looks like we are still missing notifications for PAYMENTS since they are not part of the email we send. We can do that in a different PR though.

server/services/visitation.service.js
tobieeniafe: I thought about this also, but I don't want us adding notifications, should the `findByIdAndUpdate` service fail and throw an error.

PR #163, opened 2021-04-30
#177737617 enable seeding of models

tobieeniafe:
#### What does this PR do?
Enable seeding of models for auto-population of database

Enable seeding of models for auto-population of database

`npm run seed`
`MODEL=model LIMIT=number`

#177737617


seeding/index.js
```diff
@@ -0,0 +1,34 @@
+import { startDBConnection } from '../server/config';
+import logger from '../server/config/winston';
+import { seedUsers, seedVendors } from './user.seeding';
+
+const seedDB = async () => {
+  const model = process.env.SEED;
```
andela-hpopoola, line 6: We should have a constant that defines all the models that we can seed in the database. For now, it would contain USER, RESET, UNKNOWN.
```suggestion
  const model = process.env.MODEL || MODEL.UNKNOWN;
```

seeding/index.js
```diff
@@ -0,0 +1,34 @@
...
+  const limit = process.env.LIMIT;
```
andela-hpopoola, line 7:
```suggestion
  const limit = process.env.LIMIT || 1;
```

seeding/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+  try {
+    startDBConnection();
+
+    if (model) {
+      switch (model) {
+        case 'user':
```
andela-hpopoola, line 14:
```suggestion
        case MODEL.USER:
```

seeding/index.js
```diff
@@ -0,0 +1,34 @@
...
+          seedUsers(limit);
```
andela-hpopoola, line 15: We should pass the role of the user here, we would use the role to determine the type of user we wish to seed.

seeding/index.js
```diff
@@ -0,0 +1,34 @@
...
+          break;
+        case 'vendor':
```
andela-hpopoola, line 17: This is not needed as it is not scalable since we have other roles.

seeding/index.js
```diff
@@ -0,0 +1,34 @@
...
+          seedVendors(limit);
+          break;
+
+        default:
```
andela-hpopoola, line 21: We need to add a message here to tell the user that they have input an invalid MODEL and also list the available models that are supported via seeding.

seeding/index.js
```diff
@@ -0,0 +1,34 @@
...
+import { seedUsers, seedVendors } from './user.seeding';
+
+const seedDB = async () => {
+  const model = process.env.SEED;
+  const limit = process.env.LIMIT;
+
+  try {
+    startDBConnection();
+
+    if (model) {
```
andela-hpopoola, line 12: This is also not needed. I feel it is safer for us to only seed when we know the model the user wants to seed.

seeding/index.js
```diff
@@ -0,0 +1,34 @@
...
+          break;
+        case 'vendor':
+          seedVendors(limit);
+          break;
+
+        default:
+          break;
+      }
+    } else {
+      seedUsers(limit);
```
andela-hpopoola, line 25: This is not needed as it would be defined in RESET. This should be a function when we would clear the database and create a new set of models.

seeding/user.seeding.js
```diff
@@ -0,0 +1,71 @@
...
+
+  try {
+    await User.insertMany(users);
+    logger.info(`${number} users created`);
+  } catch (error) {
+    logger.error(error);
+  }
+};
+
+export const seedVendors = async (number = 1) => {
```
andela-hpopoola, line 28: I think a new function for vendors is not needed as it can be embedded in the `seedUsers` function above. The vendor object defined below should be used if the role of the user is `vendor`. This ensures that we are maintaining only 1 function instead of 2 when we start adding more functionalities to the function. It also prevents us from creating additional functions for other roles.

package.json
```diff
@@ -70,6 +71,7 @@
     "ejs": "^3.0.2",
     "express": "^4.17.1",
     "express-rate-limit": "^5.2.6",
+    "faker": "^5.5.3",
```
andela-hpopoola, line 74: This should be installed as a `devDependencies` (see line 29 on this file) as it is only needed on the dev environment. Can you uninstall and reinstall with ` --save-dev`. Something like this:

```
npm install faker --save-dev
```

seeding/index.js
```diff
@@ -0,0 +1,34 @@
...
+      seedVendors(limit);
+    }
+    logger.info('done');
```
andela-hpopoola, line 28: Lets use `chalk` instead of logger here. This will enable us to use green colour for success, yellow for loading, blue for info, and red for errors.

https://www.npmjs.com/package/chalk

andela-hpopoola (changes requested):
Thank you for working on this. This is a good start but I left a few comments to make it even better.

package.json
tobieeniafe: This throws an error about importing a dev dependency. Can't seem to find a way around it

package.json
```diff
@@ -8,6 +8,8 @@
     "build:clean": "rm -rf build && mkdir build",
     "build:server": "babel -d ./build ./server -s",
     "coverage": "nyc report --reporter=text-lcov | coveralls",
+    "seed": "babel-node ./seeding/index.js",
+    "seed:dev": "nodemon --exec babel-node ./seeding/index.js",
```
andela-hpopoola, line 12: Not sure why you added this as seeding is only needed in development (and maybe staging). I believe we only need the one above as seeding is meant to be a 1-time thing. This can be removed 🙏🏽

seeding/index.js
```diff
@@ -0,0 +1,64 @@
...
+export const logInfo = (data) => {
+  logger.info(chalk.blueBright(data));
+};
+
+export const logError = (data) => {
+  logger.error(chalk.redBright(data));
+};
+
+const MODEL = {
+  USER: 'user',
```
andela-hpopoola, line 26: Lets ensure that that the values of the `env` variables we are getting from the user is in uppercase for consistency sake:

```suggestion
  USER: 'USER',
```

Please update the one below too for `RESET`

seeding/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+const ROLE = {
+  ADMIN: 'admin',
+  EDITOR: 'editor',
+  USER: 'user',
+  VENDOR: 'vendor',
+};
+
+const seedDB = async () => {
+  const model = process.env.MODEL;
```
andela-hpopoola, line 38: Let's convert the value of model to uppercase if a valid model is given.

seeding/index.js
```diff
@@ -0,0 +1,64 @@
...
+    switch (model) {
+      case MODEL.USER:
+        await seedUsers(limit, role);
+        break;
+      case MODEL.RESET:
+        resetDB();
+        break;
+
+      default:
+        logInfo(`Invalid model input. Available models are: ${Object.values(MODEL)}`);
```
andela-hpopoola, line 54: I think this is better as an Error.

```suggestion
        logError(`Invalid model input. Available models are: ${Object.values(MODEL)}`);
```

seeding/user.seeding.js
```diff
@@ -0,0 +1,58 @@
...
+import faker from 'faker';
+import UserFactory from '../test/factories/user.factory';
+import VendorFactory from '../test/factories/vendor.factory';
+import AddressFactory from '../test/factories/address.factory';
+import User from '../server/models/user.model';
+import { USER_ROLE } from '../server/helpers/constants';
+// eslint-disable-next-line import/no-cycle
+import { logLoading, logError } from './index';
+
+const seedUsers = async (limit, role) => {
```
andela-hpopoola, line 11: We should be able to pass in some default values here so that we can consistent results when we reset the database. So for default we should always create at least 1 user type each  for each role with the same email to avoid thinking about the email or password of the created user.

seeding/user.seeding.js
```diff
@@ -0,0 +1,58 @@
...
+        name: faker.name.findName(),
+        isSignatory: true,
+        signature: faker.image.abstract(),
+        phone: faker.phone.phoneNumber(),
+      },
+    ],
+  });
+
+  const users = UserFactory.buildList(limit, {
+    role: USER_ROLE[role.toUpperCase()],
```
andela-hpopoola, line 33: It looks like `password` field is missing here. Can you please add it and use a default password of `passworded` for all created users 🙏🏽. 

Also, can we add other fields like
- activationDate,
- phone2 (Try getting a random number from 0 to 1, if it is greater than 0.5 then you should add a phone2 else you will leave it as null)
- profileImage

andela-hpopoola (changes requested):
Thank you for working on this. I left more comments on some of the areas you missed. Hopefully, we are almost through with the first part of the seeding.

package.json
andela-hpopoola: Strange! From the docs, this shouldn't throw an error

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md

Let me know if this continues so that we can explicitly disable it from eslint.

seeding/index.js
```diff
@@ -1,57 +1,27 @@
...
-const ROLE = {
-  ADMIN: 'admin',
-  EDITOR: 'editor',
-  USER: 'user',
-  VENDOR: 'vendor',
-};
-
 const seedDB = async () => {
-  const model = process.env.MODEL;
+  const model = process.env.MODEL.toUpperCase();
```
andela-hpopoola, line 8: You might need to add chaining here in case `MODEL` is undefined to prevent an error.

```
   const model = process.env.MODEL?.toUpperCase();
```

seeding/reset.seeding.js
```diff
@@ -1,10 +1,23 @@
...
+import { logLoading, logError, logInfo } from './seed-helpers';
+import { ROLE } from './seed-constants';
+import seedUsers from './user.seeding';
+
+const resetDB = async () => {
+  clearDB();
+
+  await new Promise((resolve) => {
+    logInfo('Clearing database');
+    return setTimeout(resolve, 5000);
```
andela-hpopoola, line 11: I believe the Promise here is for testing purposes, can we remove it if it doesn't add value to the code.

seeding/reset.seeding.js
```diff
@@ -1,10 +1,23 @@
...
+  });
 
-const resetDB = () => {
   try {
-    clearDB();
+    await Promise.all(
```
andela-hpopoola, line 15: 🙌🏽 Awesome. 

Nice concept here. However, can we place it in a different function so that we don't congest these sections when we add more models?

Lets create a function called `addDefaultUsers` and call the function here.

seeding/user.seeding.js
```diff
@@ -30,26 +32,34 @@ const seedUsers = async (limit, role) => {
...
-    role: USER_ROLE[role.toUpperCase()],
+    role: roleValue,
     activated: true,
+    activationDate: new Date(),
     firstName: faker.name.firstName(),
     lastName: faker.name.lastName(),
-    email: faker.internet.email(),
+    email: defaultValues.email || faker.internet.email(),
+    password: 'passworded',
+    confirmPassword: 'passworded',
```
andela-hpopoola, line 42: This is not needed as we are inserting the user details into the database directly.

seeding/user.seeding.js
```diff
@@ -30,26 +32,34 @@ const seedUsers = async (limit, role) => {
...
   const users = UserFactory.buildList(limit, {
-    role: USER_ROLE[role.toUpperCase()],
+    role: roleValue,
     activated: true,
+    activationDate: new Date(),
     firstName: faker.name.firstName(),
     lastName: faker.name.lastName(),
-    email: faker.internet.email(),
+    email: defaultValues.email || faker.internet.email(),
+    password: 'passworded',
```
andela-hpopoola, line 41: This should be hashed. We can used the `hashPassword` helper class in `user.service.js`

andela-hpopoola (changes requested):
Looks much better, thank you for working on this.

seeding/reset.seeding.js
tobieeniafe: I noticed that when there's data in the database, there's usually a concurrent background event from `seedUsers` trying to add users throwing a duplicate error. So the promise is to delay till the database is cleared before running.

andela-hpopoola: Adding the setTimeout would not fix the problem you alighted above and looks like a hack that wouldn't work every time or on every machine. Also, we don't need the unnecessary delay (it's unproductive). 


The best fix in my own opinion is to wait for the promise clearing the database to resolve. That way we are sure that that the database has been cleared and we can continue with the addition of the data.

seeding/seed-helpers.js
```diff
@@ -0,0 +1,19 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import chalk from 'chalk';
+import logger from '../server/config/winston';
```
andela-hpopoola, line 3: Since we are making this a development stuff only, lets remove `logger` and use `console.log` for simplicity. Also, you would need to add `/* eslint-disable no-console */` at the top of the file to disable this.

seeding/user.seeding.js
```diff
@@ -0,0 +1,76 @@
...
+import VendorFactory from '../test/factories/vendor.factory';
+import AddressFactory from '../test/factories/address.factory';
+import User from '../server/models/user.model';
+import { USER_ROLE } from '../server/helpers/constants';
+import { logLoading, logError } from './seed-helpers';
+import { ROLE } from './seed-constants';
+import { hashPassword } from '../server/services/user.service';
+
+export const seedUsers = async (limit, role, defaultValues = {}) => {
+  const roleValue = Object.keys(ROLE).find((key) => ROLE[key] === role);
```
andela-hpopoola, line 13: Lets have a default roleValue of `User` here is the role is not given.

seeding/user.seeding.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const hashedPassword = await hashPassword('passworded');
```
andela-hpopoola, line 15: Lets create a DEFAULT_PASSWORD constant in the constant file so that it is easier to update (incase anyone wants to update the DEFAULT PASSWORD for their own use case).

seeding/user.seeding.js
```diff
@@ -0,0 +1,76 @@
...
+    }),
+    vendor: parseInt(roleValue, 10) === USER_ROLE.VENDOR ? vendor : {},
+  });
+
+  try {
+    if (!roleValue) {
+      logError(`Invalid role input. Available roles are: ${Object.values(ROLE)}`);
+    } else {
+      await User.insertMany(users);
+      logLoading(`${limit} ${parseInt(limit, 10) === 1 ? role : `${role}s`} created`);
```
andela-hpopoola, line 63: We would need to out a list of all the user's email and password created in the seed. You can use the package below:

https://www.npmjs.com/package/console.table

andela-hpopoola (changes requested):
Thank you for the updates on this. Here are some more things to fix

- Disconnect from the database after seeding
- Ensure that seeding is only done on the Development Environment (we don't want to mistakenly seed on the Production Database)
- Improve logs (maybe I will work on this on a later PR)

Thank you.

seeding/user.seeding.js
tobieeniafe: The role is already `user` by default.

seeding/reset.seeding.js
andela-hpopoola: Can you look this and see if you can fix this without using `setTimeout`?

seeding/user.seeding.js
```diff
@@ -0,0 +1,83 @@
...
+import User from '../server/models/user.model';
+import { USER_ROLE } from '../server/helpers/constants';
+import { logLoading, logError, logTable } from './seed-helpers';
+import { ROLE, DEFAULT_PASSWORD } from './seed-constants';
+import { hashPassword } from '../server/services/user.service';
+
+export const seedUsers = async (limit, role, customValues = {}) => {
+  const roleValue = Object.keys(ROLE).find((key) => ROLE[key] === role);
+
+  const hashedPassword = await hashPassword(customValues.password || DEFAULT_PASSWORD);
```
andela-hpopoola, line 15: We can get the password first before hashing the password so that it can be reused in multiple places instead of using `customValues.password || DEFAULT_PASSWORD`.

```js
const password = customValues.password || DEFAULT_PASSWORD;
const hashedPassword = await hashPassword(password);
```

seeding/user.seeding.js
```diff
@@ -0,0 +1,83 @@
...
+  try {
+    if (!roleValue) {
+      logError(`Invalid role input. Available roles are: ${Object.values(ROLE)}`);
+    } else {
+      await User.insertMany(users);
+      logLoading(`${limit} ${parseInt(limit, 10) === 1 ? role : `${role}s`} created`);
+
+      const table = [];
+      users.forEach((user) => {
+        table.push({ email: user.email, password: customValues.password || DEFAULT_PASSWORD });
```
andela-hpopoola, line 68: Take note of the capital `E` in email and `P` in password.

```suggestion
        table.push({ Email: user.email, Password: password });
```

seeding/index.js
```diff
@@ -0,0 +1,42 @@
...
+    logError(error);
+  }
+
+  closeDBConnection();
+};
+
+if (NODE_ENV === 'development') {
+  seedDB();
+} else {
+  logError('Seeding is only available in development');
```
andela-hpopoola, line 41:
```suggestion
  logError('Seeding is only available in `development environment`');
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments but the most important one is finding a way to remove the setTimeout used when clearing the database. Once that is resolved, this should be safe to merge. Thank you.

andela-hpopoola (approved):
Looks good to me. Thank you for working on this 🙌🏽

PR #170, opened 2021-05-12
#178088126 make admin approve new properties added by vendors

tobieeniafe:
#### What does this PR do?
Make admin approve new properties added by vendors

Make admin approve new properties added by vendors

`PUT /api/v1/property/approve/:id`

#178088126

### DIFF
--- a/server/services/property.service.js
+++ b/server/services/property.service.js
@@ -271,12 +271,6 @@ export const searchThroughProperties = async (user, { page = 1, limit = 10, ...q
     { $match: { $and: filterQuery } },
     { $sort: sortQuery },
     {
-      $match: {
-        'flagged.status': false,
-        'approved.status': true,
-      },
-    },
-    {
       $lookup: {
         from: 'users',
         localField: 'assignedTo',
@@ -319,7 +313,13 @@ export const searchThroughProperties = async (user, { page = 1, limit = 10, ...q
   }
 
   if (user.role === USER_ROLE.USER) {
-    propertyOptions.unshift({ $match: { assignedTo: { $nin: [ObjectId(user._id)] } } });
+    propertyOptions.unshift({
+      $match: {
+        assignedTo: { $nin: [ObjectId(user._id)] },
+        'flagged.status': false,
+        'approved.status': true,
+      },
+    });
   }
 
   const properties = await Property.aggregate(propertyOptions);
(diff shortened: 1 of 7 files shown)


andela-hpopoola (changes requested):
Thank you for working on this. Even though I am sure that these have been covered in the code, can we write more robust tests to confirm the following for future purposes:

1. When a user gets all properties, they see only approved properties
2. Vendors get all of their properties (approved and unapproved)
3. Admins get all properties (approved and unapproved)
4. User cannot view unapproved property by Id (returns not found)
5. Vendors and admins can view an unapproved property.

Thank you.

server/services/property.service.js
```diff
@@ -319,7 +313,13 @@ export const searchThroughProperties = async (user, { page = 1, limit = 10, ...q
   }
 
   if (user.role === USER_ROLE.USER) {
-    propertyOptions.unshift({ $match: { assignedTo: { $nin: [ObjectId(user._id)] } } });
+    propertyOptions.unshift({
```
andela-hpopoola, line 316: 🙌🏽 . This is one of the beauties of comprehensive testing. I am sure this bug was caught via the updated tests. Thank you.

test/controllers/property.controller.test.js
```diff
@@ -1329,6 +1329,48 @@ describe('Property Controller', () => {
       });
     });
 
+    context('when property is not approved', () => {
+      beforeEach(async () => {
+        await Property.findByIdAndUpdate(property._id, { 'approved.status': false });
+      });
+
+      context('with user id', () => {
```
andela-hpopoola, line 1337: I feel using token here makes more sense
```suggestion
    context('with user token', () => {
```

test/controllers/property.controller.test.js
```diff
@@ -2002,17 +2055,40 @@ describe('Property Controller', () => {
...
-          }),
-        );
+        itReturnsTheRightPaginationValue({
+          endpoint,
+          method,
+          user: regularUser,
+          useExistingUser: true,
+        });
+
+        context('with a admin and vendor token', () => {
```
andela-hpopoola, line 2074: Can we write equivalent test for user here too to be sure that the right results are returned for users.

andela-hpopoola (approved):
Thank you for working on this. I left a few comments but this should be ready to merge when you fix them. Thank you and Great work.

PR #171, opened 2021-05-19
#178016449 return all models in database

tobieeniafe:
#### What does this PR do?
Return all models in database

Return all models in database

`GET /api/v1/all`

#178016449

### DIFF
--- a/server/controllers/user.controllers.js
+++ b/server/controllers/user.controllers.js
@@ -5,7 +5,6 @@ import {
   forgotPasswordToken,
   resetPasswordViaToken,
   updateUser,
-  assignPropertyToUser,
   getAllUsers,
   getUserInfo,
   addPropertyToFavorites,
@@ -110,16 +109,6 @@ const UserController = {
           message: 'Your information has been successfully retrieved',
           user,
         });
-      })
-      .catch((error) => next(error));
-  },
-
-  assignProperty(req, res, next) {
-    const toBeAssigned = req.locals;
-    const { user } = req;
-    assignPropertyToUser({ ...toBeAssigned, assignedBy: user._id, vendor: user })
-      .then(() => {
-        res.status(httpStatus.OK).json({ success: true, message: 'Property assigned' });
       })
       .catch((error) => next(error));
   },
--- a/server/models/property.model.js
+++ b/server/models/property.model.js
@@ -184,10 +184,6 @@ const PropertySchema = new mongoose.Schema(
       type: ObjectId,
       required: true,
     },
-    updatedBy: {
-      type: ObjectId,
-      required: true,
-    },
     flagged: {
       status: {
         type: Boolean,
--- a/server/routes/model.routes.js
+++ b/server/routes/model.routes.js
@@ -1,0 +1,20 @@
+import express from 'express';
+import { authenticate, isAdminOrUserOrVendor } from '../helpers/middleware';
+import ModelController from '../controllers/model.controllers';
+
+const router = express.Router();
+
+/**
+ * @swagger
+ * path:
+ *  /:
+ *    get:
+ *      summary: Returns all models in database
+ *      tags: [Models]
+ *      responses:
+ *        "200":
+ *          description: All models
+ */
+router.get('/', authenticate, isAdminOrUserOrVendor, ModelController.getAllModels);
+
+module.exports = router;
--- a/server/services/model.service.js
+++ b/server/services/model.service.js
@@ -18,19 +18,23 @@ const getAllModels = async (user) => {
   let matchObject;
   let propertyMatchObject;
   let referralMatchObject;
+  let reportedPropertyMatchObject;
 
   if (user.role === USER_ROLE.USER) {
     matchObject = { userId: ObjectId(user._id) };
     propertyMatchObject = { assignedTo: { $in: [ObjectId(user._id)] } };
-    referralMatchObject = { userId: ObjectId(user._id) };
+    referralMatchObject = { referrerId: ObjectId(user._id) };
+    reportedPropertyMatchObject = { reportedBy: ObjectId(user._id) };
   } else if (user.role === USER_ROLE.VENDOR) {
     matchObject = { vendorId: ObjectId(user._id) };
     propertyMatchObject = { addedBy: ObjectId(user._id) };
-    referralMatchObject = { userId: ObjectId(user._id) };
+    reportedPropertyMatchObject = {};
+    referralMatchObject = { referrerId: ObjectId(user._id) };
   } else {
     propertyMatchObject = {};
     referralMatchObject = {};
     matchObject = {};
+    reportedPropertyMatchObject = {};
   }
   try {
     const enquiries = await Enquiry.countDocuments(matchObject);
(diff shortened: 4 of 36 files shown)


server/routes/model.routes.js
```diff
@@ -0,0 +1,20 @@
...
+import ModelController from '../controllers/model.controllers';
+
+const router = express.Router();
+
+/**
+ * @swagger
+ * path:
+ *  /:
+ *    get:
+ *      summary: Returns all models in database
```
andela-hpopoola, line 12: Please fix for description too below:

```suggestion
 *      summary: Count the total number of each model in the database
```

server/services/model.service.js
```diff
@@ -0,0 +1,79 @@
...
+import Referral from '../models/referral.model';
+import ReportedProperty from '../models/reportedProperty.model';
+import Transaction from '../models/transaction.model';
+import User from '../models/user.model';
+import Visitation from '../models/visitation.model';
+
+const { ObjectId } = mongoose.Types.ObjectId;
+
+const getAllModels = async (user) => {
+  let matchObject;
```
andela-hpopoola, line 18: Instead of defining the empty objects in the `else` block, I think it is better you define them here so that we can remove the else block.

```suggestion
  let matchObject = {};
```

server/routes/model.routes.js
```diff
@@ -0,0 +1,20 @@
...
+ *      tags: [Models]
+ *      responses:
+ *        "200":
+ *          description: All models
+ */
+router.get('/', authenticate, isAdminOrUserOrVendor, ModelController.getAllModels);
```
andela-hpopoola, line 18: Using `Model` gives me a strange meaning (not sure if it is just me or it would make understanding the purpose of the model unclear. Lets change this to something totally different like TotalCount. 

Kindly fix the filenames and other related stuff too 🙏🏽 

```suggestion
router.get('/', authenticate, isAdminOrUserOrVendor, TotalCountController.getAllModels);
```

server/services/model.service.js
```diff
@@ -0,0 +1,79 @@
...
+    referralMatchObject = { referrerId: ObjectId(user._id) };
+  } else {
+    propertyMatchObject = {};
+    referralMatchObject = {};
+    matchObject = {};
+    reportedPropertyMatchObject = {};
+  }
+  try {
+    const enquiries = await Enquiry.countDocuments(matchObject);
+    const offers = await Offer.countDocuments(matchObject);
```
andela-hpopoola, line 41: I think we would also need to retrieve the count of portfolios too (offer letters that have been signed)

server/services/model.service.js
```diff
@@ -0,0 +1,79 @@
...
+      properties,
+      referrals,
+      reportedProperties,
+      scheduledVisitations,
+      transactions,
+      users,
+    };
+
+    if (user.role === USER_ROLE.USER) {
+      delete models.users;
```
andela-hpopoola, line 66: Instead of deleting the `users` object property, I think it is best not to make the request on line 51 for user roles. We might remove it totally or make it have a default of 0 as the total count.

server/services/model.service.js
```diff
@@ -0,0 +1,79 @@
...
+    } else if (user.role === USER_ROLE.VENDOR) {
+      delete models.users;
```
andela-hpopoola, line 68: The same comment as above. It is expensive to make these 3 calls to the database for vendor role when they are not needed.

server/controllers/user.controllers.js
```diff
@@ -114,16 +113,6 @@ const UserController = {
       .catch((error) => next(error));
   },
 
-  assignProperty(req, res, next) {
```
andela-hpopoola, line 117: This doesn't seem related to this PR in anyway, not sure why you removed it. Even if it is necessary to remove it, It would have been better in a separate PR to enable you to state the actual reason why it is necessary to remove it.  

Putting everything in a single PR made reviewing this PR more difficult than expected.

server/models/property.model.js
```diff
@@ -184,10 +184,6 @@ const PropertySchema = new mongoose.Schema(
       type: ObjectId,
       required: true,
     },
-    updatedBy: {
```
andela-hpopoola, line 187: Same comment as above; it would have been better if this is removed in a separate PR -- the number of file changes based on this is even more than the number of file changes based on the context of this PR. 

Let's try to stick to the convention please and make life easier for each and every one of us 🙏🏽 . If we need to make any unrelated changes, let it be just a line or 2 line changes so that it doesn't disrupt us from the purpose of the task being reviewed in the PR itself.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on renaming `models` to `total-counts` for more context and ensuring that we only make the requests that are needed for each role. 

I also pointed out that we should try to minimize the implementation of tasks not related to a PR to make the tasks implemented in the pull request more understandable.

Welcome back, thank you and great work 👍🏽

server/services/totalCount.service.js
```diff
@@ -14,11 +14,20 @@ import Visitation from '../models/visitation.model';
...
-  let referralMatchObject;
-  let reportedPropertyMatchObject;
+const getTotalCount = async (user) => {
+  let matchObject = {};
+  let propertyMatchObject = {};
+  let referralMatchObject = {};
+  let reportedPropertyMatchObject = {};
+
+  const portfolioMatchObject = {
+    $or: [
```
andela-hpopoola, line 24: Can we have all of the portfolio statuses in a constant and replace all the instances we have used them so that we can have a single source of truth for them.

```js
const VALID_PORTFOLIO_OFFER = [
      { status: OFFER_STATUS.ASSIGNED },
      { status: OFFER_STATUS.ALLOCATED },
      { status: OFFER_STATUS.INTERESTED },
      { status: OFFER_STATUS.RESOLVED },
];

```

server/services/totalCount.service.js
```diff
@@ -48,32 +52,32 @@ const getAllModels = async (user) => {
...
       transactions,
-      users,
     };
 
-    if (user.role === USER_ROLE.USER) {
-      delete models.users;
-    } else if (user.role === USER_ROLE.VENDOR) {
-      delete models.users;
+    if (user.role === USER_ROLE.VENDOR) {
       delete models.reportedProperties;
```
andela-hpopoola, line 69: We can still get rid of this by only making the requests when the role is an admin or a user as you have done on line 73 (for user model). Doing this will save us from making 2 unneeded requests for vendors.

server/services/totalCount.service.js
andela-hpopoola, line 24: Can we have all of the portfolio statuses in a constant and replace all the instances we have used them so that we can have a single source of truth for them.

```js
const VALID_PORTFOLIO_OFFER = [
      { status: OFFER_STATUS.ASSIGNED },
      { status: OFFER_STATUS.ALLOCATED },
      { status: OFFER_STATUS.INTERESTED },
      { status: OFFER_STATUS.RESOLVED },
];

```

andela-hpopoola (approved):
Thank you for helping out with this. This looks good to me. I left a few comments but they should be easy to fix. Nice work :100:

andela-hpopoola (approved):
Thank you for helping out with this. This looks good to me. I left a few comments but they should be easy to fix. Nice work :100:

PR #172, opened 2021-05-23
#177810760 admin pay referrals

tobieeniafe:
#### What does this PR do?
Enable admin to pay out referrals

Enable admin to pay out referrals
`PUT /api/v1/referral/pay-referral/:id`

#177810760

### DIFF
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -730,3 +730,8 @@ export const reactivateOffer = async (offerInfo) => {
     throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error reactivating offer', error);
   }
 };
+
+export const getUserFirstOfferByUSerId = async (userId) =>
+  Offer.find({ userId: ObjectId(userId) })
+    .sort({ _id: 1 })
+    .limit(1);
--- a/server/controllers/referral.controllers.js
+++ b/server/controllers/referral.controllers.js
@@ -5,6 +5,7 @@ import {
   getUserByRefCode,
   getReferralById,
   getAllReferrals,
+  payReferral,
 } from '../services/referral.service';
 import httpStatus from '../helpers/httpStatus';
 import EMAIL_CONTENT from '../../mailer';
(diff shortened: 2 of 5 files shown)


andela-hpopoola (changes requested):
Thank you for working on this. However, I believe there are lots of flaws with your current implementation as you didn't consider so other workflows related to this. 

Here are my thought process (not properly fleshed out at the moment though):

1. When a user is buying his first property, we can check if he was referred. If yes, then we can add the referral Id to the offer. That way we would be able to get the referral details from the offer. We can add the offerId to the referral to make us know the offer related with the referral.
2. When a user makes his first payment, we should update the `REWARD_STATUS` of that offer to `Progress`
3. When the user completes payment (or make the last payment) for the property with referralId, we should update the `REWARD_STATUS` to `Completed`. The completed status makes the admin aware of who to pay and why they are paying them.
4. When the user has been paid, the `REWARD_STATUS` to `Paid`.

**Important things to Note**

- The `REWARD_STATUS` makes it easier to know who and when to pay so we will also need to add filter and sort to get all referrals.
- When a user makes his first payment, we want to inform the referral that he should be expecting his money and how much he is expecting
- When we are retrieving a single referral from the endpoint, we should retrieve it with the referral information, offer Info and maybe property info.
- The `getAllReferral` endpoint should be upgraded to support filters and sorting to enable the admin to know when to pay


The process is still a little bit fuzzy to me too, so please feel free to ask questions if you don't understand any process here, or you have a better suggestion or my thought process is flawed or you sensed an edge case I didn't see 🙏🏽 .

Thank you and Nice work.

server/models/offer.model.js
```diff
@@ -94,6 +94,9 @@ const OfferSchema = new mongoose.Schema(
       type: ObjectId,
       required: true,
     },
+    referralId: {
```
andela-hpopoola, line 97: I think it is safe to remove this.

server/services/offer.service.js
```diff
@@ -51,6 +55,19 @@ export const generateReferenceCode = async (propertyId) => {
   return referenceCode;
 };
 
+export const isUserFirstProperty = async ({ userId, offerId }) => {
+  const offer = await Offer.find({ userId: ObjectId(userId) });
```
andela-hpopoola, line 59: We could count the number of offers here to verify it is 1. Also, we might need to check that the offer is a valid portfolio offer (that is the offer has been accepted by the user) so that we don't link to a rejected offer.

andela-hpopoola (changes requested):
Thank you for working on this. This looks more complicated than I thought - Lets discuss and review the workflow during standup.

PR #175, opened 2021-05-25
#178279062 link offer to referral model

tobieeniafe:
#### What does this PR do?
Link offer to referral model when offer is accepted

Link offer to referral model when offer is accepted

`PUT /api/v1/offer/accept`

#178279062

### DIFF
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -412,7 +412,7 @@ export const acceptOffer = async (offerToAccept) => {
 
   const vendor = await getUserById(offer.vendorId);
 
-  await activatePendingUserReferral({ userId: offer.userId, offerId: offer._id });
+  await activatePendingUserReferral({ userId: offer.userId, offer });
 
   try {
     await assignPropertyToUser({
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -26,7 +26,7 @@ import { createNotification } from './notification.service';
 import NOTIFICATIONS from '../helpers/notifications';
 import { getFormattedName } from '../helpers/funtions';
 // eslint-disable-next-line import/no-cycle
-import { getPendingUserReferral } from './referral.service';
+import { activatePendingUserReferral } from './referral.service';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
@@ -412,7 +412,7 @@ export const acceptOffer = async (offerToAccept) => {
 
   const vendor = await getUserById(offer.vendorId);
 
-  await getPendingUserReferral({ userId: offer.userId, offerId: offer._id });
+  await activatePendingUserReferral({ userId: offer.userId, offerId: offer._id });
 
   try {
     await assignPropertyToUser({
--- a/server/services/offer.service.js
+++ b/server/services/offer.service.js
@@ -25,6 +25,8 @@ import { addNextPayment } from './nextPayment.service';
 import { createNotification } from './notification.service';
 import NOTIFICATIONS from '../helpers/notifications';
 import { getFormattedName } from '../helpers/funtions';
+// eslint-disable-next-line import/no-cycle
+import { userIsReferredWithoutActiveReferral, startReferral } from './referral.service';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
@@ -358,15 +360,15 @@ export const createOffer = async (offer) => {
 };
 
 export const acceptOffer = async (offerToAccept) => {
-  const offer = await getOffer(offerToAccept.offerId, offerToAccept.user).catch((error) => {
+  const [offer] = await getOffer(offerToAccept.offerId, offerToAccept.user).catch((error) => {
     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
   });
 
-  if (offer.length < 1) {
+  if (!offer) {
     throw new ErrorHandler(httpStatus.PRECONDITION_FAILED, 'Invalid offer');
   }
 
-  if (offer[0].userId.toString() !== offerToAccept.user._id.toString()) {
+  if (offer.userId.toString() !== offerToAccept.user._id.toString()) {
     throw new ErrorHandler(
       httpStatus.PRECONDITION_FAILED,
       'You cannot accept offer of another user',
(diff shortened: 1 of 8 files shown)


server/services/offer.service.js
```diff
@@ -358,67 +360,73 @@ export const createOffer = async (offer) => {
...
+    offerId: offer._id,
+    propertyId: offer.propertyId,
+    userId: offer.userId,
+    vendorId: offer.vendorId,
   };
 
-  const vendor = await getUserById(offer[0].vendorId);
+  const vendor = await getUserById(offer.vendorId);
+
+  const { isReferred, referralId } = await userIsReferredWithoutActiveReferral(offer.userId);
```
andela-hpopoola, line 415: I think we don't need the `isReferred` variable here as we can make use of `referralId` to confirm if a user has a pending referral. If `referralId` is null, then `isReferred` will be false, hence making it obsolete.

Also, we can update the function to `getPendingUserReferral` and return the `referralId` only.

```suggestion
  const referralId = await getPendingUserReferral(offer.userId);
```

server/services/referral.service.js
```diff
@@ -212,3 +219,24 @@ export const calculateReferralRewards = async (referrerId) =>
...
     },
   ]);
+
+export const userIsReferredWithoutActiveReferral = async (userId) => {
+  const referral = await Referral.find({
+    userId: ObjectId(userId),
+    'reward.status': REWARD_STATUS.PENDING,
+  });
+
+  return { isReferred: referral.length !== 0, referralId: referral[0]?._id || null };
```
andela-hpopoola, line 229: I think this should work 🤷🏽‍♂️ 

```suggestion
  return referral?.[0]?._id
```

server/services/offer.service.js
```diff
@@ -358,67 +360,73 @@ export const createOffer = async (offer) => {
...
+
+  if (isReferred) {
+    await startReferral({ referralId, offerId: offer._id });
```
andela-hpopoola, line 418: Not sure if this will make the code better, but I think we can combine the `userIsReferredWithoutActiveReferral`  and the `startReferral` function together into a single function like `activateReferral` so that the checking if the referral exist and the actual updates is handled by one function.

server/services/referral.service.js
```diff
@@ -111,8 +111,8 @@ export const sendReferralInvite = async (invite) => {
   }
 };
 
-export const getReferralById = async (referralId) =>
-  Referral.aggregate([
+export const getReferralById = async (referralId) => {
+  const referral = await Referral.aggregate([
```
andela-hpopoola, line 115: Can we destructure this to make it easier to use:
```suggestion
  const [referral] = await Referral.aggregate([
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments but I think we are in the right direction. Can you write tests for this so that we can proceed to the next story 🙏🏽 . Thank you.

server/services/referral.service.js
```diff
@@ -212,3 +219,22 @@ export const calculateReferralRewards = async (referrerId) =>
       },
     },
   ]);
+
+export const getPendingUserReferral = async ({ userId, offerId }) => {
```
andela-hpopoola, line 223: This should be `activatePendingUserReferral` instead of `getPendingUserReferral` as the main purpose of this function is to update the reward status to `Started` and not to get the Referral information.

server/services/referral.service.js
```diff
@@ -212,3 +219,22 @@ export const calculateReferralRewards = async (referrerId) =>
...
+  const referral = await Referral.find({
+    userId: ObjectId(userId),
+    'reward.status': REWARD_STATUS.PENDING,
+  });
+
+  const referralId = referral?.[0]?._id;
+
+  try {
```
andela-hpopoola, line 231: I would be careful to only run this try block if and only if `referral` is not empty. We don't want to update when `referralId` is null 🤷🏽‍♂️

server/services/referral.service.js
```diff
@@ -212,3 +219,22 @@ export const calculateReferralRewards = async (referrerId) =>
       },
     },
   ]);
+
+export const getPendingUserReferral = async ({ userId, offerId }) => {
+  const referral = await Referral.find({
```
andela-hpopoola, line 224: I think I would want to use findOne here since we are expecting the returned referral to be 1.

server/services/referral.service.js
```diff
@@ -212,3 +219,22 @@ export const calculateReferralRewards = async (referrerId) =>
...
+    userId: ObjectId(userId),
+    'reward.status': REWARD_STATUS.PENDING,
+  });
+
+  const referralId = referral?.[0]?._id;
```
andela-hpopoola, line 229: If we use `findOne` above, this will no longer be necessary as we can check if referral is empty and get referralId with `referral._id` if referral is not empty.

andela-hpopoola (changes requested):
Thank you for making the updates, I added a few nits to improve the code quality. Thank you.

server/services/referral.service.js
andela-hpopoola: This has not been resolved. We should not update the Referral model if referral is null. Apart from throwing an error on line 231 for `referral._id`,  we are making a request that is not needed.

test/controllers/referral.controller.test.js
```diff
@@ -601,9 +601,10 @@ describe('Referral Controller', () => {
             .set('authorization', adminToken)
             .send(invalidData)
             .end((err, res) => {
-              expect(res).to.have.status(400);
+              expect(res).to.have.status(500);
```
andela-hpopoola, line 604: This looks strange. I would expect `res.status` and `error.statusCode` on line 606 to have the same value.

test/services/offer.service.test.js
```diff
@@ -492,6 +495,31 @@ describe('Offer Service', () => {
...
+
+      context('when all is valid', () => {
+        it('returns a valid accepted offer', async () => {
+          const acceptedOffer = await acceptOffer(toAcceptValid);
+          const updatedReferral = await Referral.findById(referral._id).select();
+          expect(acceptedOffer.status).to.eql('Interested');
+          expect(acceptedOffer.contributionReward).to.eql(2000000);
+          expect(acceptedOffer.signature).to.eql(toAcceptValid.signature);
+          expect(updatedReferral.userId).to.eql(user1._id);
+          expect(updatedReferral.referrerId).to.eql(user2._id);
```
andela-hpopoola, line 517: I think we should also update the referral amount (1.5%) when linking the offer to the referral.

test/services/offer.service.test.js
```diff
@@ -492,6 +495,31 @@ describe('Offer Service', () => {
...
+      const referral = ReferralFactory.build(
+        { referrerId: user2._id, userId: user1._id, 'reward.status': REWARD_STATUS.PENDING },
+        { generateId: true },
+      );
+
+      beforeEach(async () => {
+        await addReferral(referral);
+      });
+
+      context('when all is valid', () => {
```
andela-hpopoola, line 509:
```suggestion
      context('when the referral is pending', () => {
```

We need to confirm that the reward status is only updated when it is `PENDING`. You might need to use a loop to test for other reward status.

test/services/offer.service.test.js
```diff
@@ -492,6 +495,31 @@ describe('Offer Service', () => {
         expect(acceptedOffer.signature).to.eql(toAcceptInvalid.signature);
       });
     });
+
+    describe('when user was referred', () => {
```
andela-hpopoola, line 499: Also, we need to add tests for situations where the user is not referred.

andela-hpopoola (changes requested):
Thank you for making the changes. I left a few comments but the most important on is adding the calculated amount to the referral model, adding more tests and fixing the broken tests. Thank you.

server/services/referral.service.js
```diff
@@ -212,3 +219,20 @@ export const calculateReferralRewards = async (referrerId) =>
       },
     },
   ]);
+
+export const activatePendingUserReferral = async ({ userId, offerId }) => {
+  const referral = await Referral.findOne({
+    userId: ObjectId(userId),
+    'reward.status': REWARD_STATUS.PENDING,
```
andela-hpopoola, line 226: I think we also need to check that `referral.status` is `REFERRAL_STATUS.REGISTERED` to cater for situations where more than one person referral a user and use the actual referralId that they registered with.

You would also need to add tests to verify this (i.e when a user is referred multiple times, we only get the one they registered with and add one more test to confirm that we don't get a referral with `REFERRAL_STATUS` of `REWARDED`)

test/controllers/referral.controller.test.js
tobieeniafe: The error is thrown from line `146` `referral.service.js` when the `getReferralById` function is called by the `updateReferralToRewarded`.

server/services/offer.service.js
```diff
@@ -412,7 +412,7 @@ export const acceptOffer = async (offerToAccept) => {
 
   const vendor = await getUserById(offer.vendorId);
 
-  await activatePendingUserReferral({ userId: offer.userId, offerId: offer._id });
+  await activatePendingUserReferral({ userId: offer.userId, offer });
```
andela-hpopoola, line 415: I think it makes more sense to just pass `offer` only here since we can get the `userId` from offer.

```suggestion
  await activatePendingUserReferral({ offer });
```

test/controllers/referral.controller.test.js
andela-hpopoola, line 604: Lets remove this, please. It is so confusing.

test/services/offer.service.test.js
```diff
@@ -454,42 +462,105 @@ describe('Offer Service', () => {
...
+          expect(updatedReferral.referrerId).to.eql(user2._id);
+          expect(updatedReferral.offerId).to.eql(offer1._id);
+          expect(updatedReferral.reward.status).to.eql(REWARD_STATUS.STARTED);
+          expect(updatedReferral.reward.amount).to.eql(270_000);
+        });
+      });
+
+      context('when the referral is started', () => {
+        const startedReferralStatuses = Object.keys(REWARD_STATUS);
+        delete startedReferralStatuses[0];
```
andela-hpopoola, line 538: You will need to be more explicit here, that is to provide the key of the object you are deleting. This will be problematic if a new status is added as index 0 to REWARD_STATUS

test/services/offer.service.test.js
```diff
@@ -454,42 +462,105 @@ describe('Offer Service', () => {
...
-        const acceptedOffer = await acceptOffer(toAcceptInvalid);
-        expect(acceptedOffer.status).to.eql('Interested');
-        expect(acceptedOffer.contributionReward).to.eql(0);
-        expect(acceptedOffer.signature).to.eql(toAcceptInvalid.signature);
+    describe('when user was referred', () => {
+      const referral = ReferralFactory.build(
+        {
+          referrerId: user2._id,
+          userId: user1._id,
+          status: REFERRAL_STATUS.REGISTERED,
```
andela-hpopoola, line 511: I think we would need another set of tests to confirm that the updates are made when status is `REFERRAL_STATUS.REGISTERED`

andela-hpopoola (changes requested):
Thank you for your hardwork on this. I left a few comments to conclude this. Nice work.

test/services/offer.service.test.js
```diff
@@ -533,15 +533,45 @@ describe('Offer Service', () => {
         });
       });
 
-      context('when the referral is started', () => {
-        const startedReferralStatuses = Object.keys(REWARD_STATUS);
-        delete startedReferralStatuses[0];
+      context('when the reward status has started', () => {
+        const startedReferralStatuses = Object.keys(REWARD_STATUS).filter(
```
andela-hpopoola, line 537:
```suggestion
        const nonPendingReferralStatuses = Object.keys(REWARD_STATUS).filter(
```

andela-hpopoola (approved):

PR #177, opened 2021-05-30
#178279081 update referral status when payment is made

tobieeniafe:
#### What does this PR do?
Update referral status when payment is made

Update referral status when payment is made

`/api/v1/offline-payment/resolve/:id`

#178279081

### DIFF
--- a/server/helpers/constants.js
+++ b/server/helpers/constants.js
@@ -26,9 +26,9 @@ export const REWARD_STATUS = {
   PENDING: 'Pending',
   STARTED: 'Started',
   PAYMENT_STARTED: 'Payment Started',
-  PROGRESS: 'Progress',
-  AWAITING_PAYMENT: 'Awaiting Payment',
-  PAID: 'Paid',
+  PAYMENT_IN_PROGRESS: 'Payment In Progress',
+  PAYMENT_COMPLETED: 'Payment Completed',
+  REFERRAL_PAID: 'Referral Paid',
 };
 
 export const CONCERN_STATUS = {
--- a/server/helpers/constants.js
+++ b/server/helpers/constants.js
@@ -26,6 +26,7 @@ export const REWARD_STATUS = {
   PENDING: 'Pending',
   STARTED: 'Started',
   PROGRESS: 'Progress',
+  AWAITING_PAYMENT: 'Awaiting Payment',
   PAID: 'Paid',
 };
 
--- a/server/helpers/constants.js
+++ b/server/helpers/constants.js
@@ -25,6 +25,7 @@ export const REFERRAL_STATUS = {
 export const REWARD_STATUS = {
   PENDING: 'Pending',
   STARTED: 'Started',
+  PAYMENT_STARTED: 'Payment Started',
   PROGRESS: 'Progress',
   AWAITING_PAYMENT: 'Awaiting Payment',
   PAID: 'Paid',
--- a/server/services/referral.service.js
+++ b/server/services/referral.service.js
@@ -7,7 +7,7 @@ import { REFERRAL_STATUS, REWARD_STATUS, REFERRAL_PERCENTAGE } from '../helpers/
 // eslint-disable-next-line import/no-cycle
 import { getUserById, getUserByEmail } from './user.service';
 // eslint-disable-next-line import/no-cycle
-import { isFirstOrLastPayment } from './transaction.service';
+import { isFirstOrLastOrOtherPayment } from './transaction.service';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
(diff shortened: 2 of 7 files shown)


server/services/referral.service.js
```diff
@@ -246,3 +248,23 @@ export const activatePendingUserReferral = async (offer) => {
...
+      referralId,
+      { $set: { offerId, 'reward.status': REWARD_STATUS.STARTED } },
+      { new: true },
+    );
+  }
+
+  if (ifLastPayment) {
+    await Referral.findByIdAndUpdate(
+      referralId,
+      { $set: { offerId, 'reward.status': REWARD_STATUS.PROGRESS } },
```
andela-hpopoola, line 266: `PROGRESS` here doesn't seem ideal. Let's create a new REWARD_STATUS called `AWAITING PAYMENT` to indicate that payment is the next step.

Lets use `PROGRESS` for other payments (payments not first or last payment)

server/services/referral.service.js
```diff
@@ -246,3 +248,23 @@ export const activatePendingUserReferral = async (offer) => {
     }
   }
 };
+
+export const updateReferralRewardStatus = async ({ referralId, offerId }) => {
+  const { ifLastPayment, ifFirstPayment } = await isFirstOrLastPayment(offerId);
+
+  if (ifFirstPayment) {
+    await Referral.findByIdAndUpdate(
```
andela-hpopoola, line 256: I think it is better to have just one Referral updates (instead of multiple entries).

You can use Javascript `switch` statement to get the current status as explained here:
 - First payment - STARTED
 - Last Payment - AWATING_PAYMENT
 - Other Payments - PROGRESS

Once you have gotten the status, you can use a using Referral update to update the status.

server/services/transaction.service.js
```diff
@@ -327,3 +335,16 @@ export const addRemittance = async (remittanceInfo) => {
...
+
+export const isFirstOrLastPayment = async (offerId) => {
+  const offer = await getOfferById(offerId).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  const totalPaid = await getTotalTransactionByOfferId(offer._id);
+
+  const ifLastPayment = totalPaid >= offer.totalAmountPayable;
+  const ifFirstPayment = totalPaid <= offer.initialPayment;
```
andela-hpopoola, line 347: This looks wrong because if a user makes a first payment greater their initial payment, this will fail 😞. 

Can we get the list of transactions made  and manually sum them together instead so that if the length is 1 and it is less than `totalAmountPayable`, then it is firstPayment; 

if `totalSum >= offer.totalAmountPayable` than last payment

Else it can be classified as `isOtherPayment` - Not first or last payment.

server/services/transaction.service.js
```diff
@@ -327,3 +335,16 @@ export const addRemittance = async (remittanceInfo) => {
...
+
+  return { ifLastPayment, ifFirstPayment };
```
andela-hpopoola, line 349: We need to add another payment step here for `isOtherPayment`

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments. Thank you.

server/services/transaction.service.js
```diff
@@ -67,6 +69,14 @@ export const addTransaction = async (transaction) => {
       throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
     });
 
+    const referral = await activatePendingUserReferral(offer);
```
andela-hpopoola, line 72: Hmmm! I think this is also wrong. We only need to `activatePendingUserReferral(offer)` when a user accepts an offer (you would need this for your test or simply add the offerId).

The right thing to do here is to get Referral by OfferId

```suggestion
    const referral = await getReferralByOfferId(offer);
```

test/services/transaction.service.test.js
```diff
@@ -176,6 +180,62 @@ describe('Transaction Service', () => {
         }
       });
     });
+
+    context.only('when transaction is made by referred user', () => {
+      const referral = ReferralFactory.build(
+        {
+          referrerId: admin._id,
```
andela-hpopoola, line 187: As advised above, you would need to add `offerId: offer._id` here to link the referral to the offer or you would need to use `activatePendingUserReferral(offer)` to link it via the service.

andela-hpopoola (changes requested):
Thank you for making the updates. I left a few comments on fixing your failing tests. Thank you.

server/helpers/constants.js
```diff
@@ -25,7 +25,9 @@ export const REFERRAL_STATUS = {
 export const REWARD_STATUS = {
   PENDING: 'Pending',
   STARTED: 'Started',
+  PAYMENT_STARTED: 'Payment Started',
   PROGRESS: 'Progress',
```
andela-hpopoola, line 29: On a better thought, lets update the REWARD_STATUS to the following to make them more readable imho 🙏🏽:

```js
export const REWARD_STATUS = {
  PENDING: 'Pending',
  STARTED: 'Started',
  PAYMENT_STARTED: 'Payment Started',
  PAYMENT_IN_PROGRESS: 'Payment In Progress',
  PAYMENT_COMPLETED: 'Payment Completed',
  REFERRAL_PAID: 'Referral Paid'
}
```

andela-hpopoola (approved):
Thank you for working on this. This looks good to me. I left a comment about properly renaming the `REWARD_STATUS`. Feel free to merge after fixing this. 

Thank you and Nice work.

PR #179, opened 2021-06-03
#178404047 allow all users view all referrals

tobieeniafe:
#### What does this PR do?
Allow all users to view all referrals

Deprecate `GET /api/v1/referral`

`GET /api/v1/referral/all`

#178404047

### DIFF
--- a/server/services/referral.service.js
+++ b/server/services/referral.service.js
@@ -15,7 +15,7 @@ import { getUserById, getUserByEmail } from './user.service';
 import { getPaymentDuration } from './transaction.service';
 import { REFERRAL_FILTERS, buildFilterAndSortQuery } from '../helpers/filters';
 import { generatePagination, generateFacetData, getPaginationTotal } from '../helpers/pagination';
-import { projectRefereeIfAdmin } from '../helpers/projectedSchemaInfo';
+import { projectedReferralInfoForAdmin } from '../helpers/projectedSchemaInfo';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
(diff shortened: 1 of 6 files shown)


test/controllers/referral.controller.test.js
```diff
@@ -331,14 +331,38 @@ describe('Referral Controller', () => {
...
+                  ...defaultPaginationResult,
+                  total: 17,
+                });
+                expect(res.body.result[0]._id).to.be.eql(referrals[0]._id.toString());
+                done();
+              });
+          });
+        });
+
+        context('with vendor token', () => {
```
andela-hpopoola, line 350: Lets add some vendor referral so that we are sure that this is working for vendors too. Also, can we add tests for `editors` too.

server/services/referral.service.js
```diff
@@ -231,6 +207,10 @@ export const getAllReferrals = async ({ page = 1, limit = 10, ...query } = {}) =
     referralOptions.shift();
   }
 
+  if (user.role !== USER_ROLE.ADMIN) {
```
andela-hpopoola, line 210: For admins, can we ensure that we are getting the referrer and referee's details from the user table. It looks like we are only getting the referee's details.

andela-hpopoola (changes requested):
Thank you for working on this. I left 2 comments which I believe are necessary to conclude this PR. Thank you and Nice work.

PS: The PR header looks strange, you could have created a story for this task on Pivotal Tracker or asked me to help create a story. Can you do that and update the PR title to match our naming convention 🙏🏽 . Thank you.

server/services/referral.service.js
```diff
@@ -199,6 +201,22 @@ export const getAllReferrals = async (user, { page = 1, limit = 10, ...query } =
     },
   ];
 
+  if (user.role === USER_ROLE.ADMIN) {
```
andela-hpopoola, line 204: Since we are adding this for admin, we also need to include it in the test to confirm that when an admin is viewing the route, they see all the expected properties and when others are viewing the route, they don't get unnecessary information.

andela-hpopoola (approved):
Thank you for helping out with this. Changes look good to me. Feel free to merge after adding the additional tests.

PR #181, opened 2021-06-07
#178426407 allow admin view offers without users

tobieeniafe:
#### What does this PR do?
Fix bug that prevents admin from viewing referrals without users

Fix bug that prevents admin from viewing referrals without users

`GET /api/v1/referral/all`

#178426407


andela-hpopoola (approved):
Awesome. Looks good. Nice work.

PR #184, opened 2021-06-09
#178426404 allow vendor request for property to be unflagged

tobieeniafe:
#### What does this PR do?
Allow vendor request for property to be unflagged

Allow vendor request for property to be unflagged

`POST /api/v1/property/request-unflag`

#178426404

### DIFF
--- a/server/services/property.service.js
+++ b/server/services/property.service.js
@@ -758,19 +758,19 @@ export const flagProperty = async (propertyInfo) => {
     });
   }
 
+  const updatedFlaggedcase = [
+    {
+      flaggedBy: propertyInfo.adminId,
+      flaggedDate: Date.now(),
+      flaggedReason: propertyInfo.reason,
+    },
+    ...property.flagged.case,
+  ];
+
   try {
     const flaggedProperty = await Property.findByIdAndUpdate(
       property._id,
-      {
-        $set: { 'flagged.status': true },
-        $push: {
-          'flagged.case': {
-            flaggedBy: propertyInfo.adminId,
-            flaggedDate: Date.now(),
-            flaggedReason: propertyInfo.reason,
-          },
-        },
-      },
+      { $set: { 'flagged.status': true, 'flagged.case': updatedFlaggedcase } },
       { new: true },
     );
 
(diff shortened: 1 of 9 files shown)


server/services/property.service.js
```diff
@@ -924,3 +924,35 @@ export const approveProperty = async ({ propertyId, adminId }) => {
...
+  if (!property.flagged.status) {
+    throw new ErrorHandler(httpStatus.PRECONDITION_FAILED, 'Property is not flagged');
+  }
+
+  try {
+    await createNotification(NOTIFICATIONS.REQUEST_UNFLAG, property.flagged.case[0].flaggedBy, {
+      actionId: property._id,
+    });
+
+    return property;
```
andela-hpopoola, line 950: I think you missed the part where we need to update the property with the vendor's comment and changing the status to inform the admin that the property issue has been resolved (that is more productive than sending only the notification).

Apart from notifications, we want it to be in the database that the user has requested that his property should be unflagged. We also want the user to write comments in case he/she has been misunderstood or give him/her the opportunity to explain the situation. (I thought I discussed all of this with you)

andela-hpopoola (changes requested):
Thank you for working on this. However, I think you implemented the secondary focus of this story (which is notification) rather than the primary focus. The most important thing in this story is to update the database with the vendor's comment and update the status of the property to `UNFLAG_REQUESTED`. This serves the following purposes
- The admin would be able to filter properties that need to be unflagged
- The admin knows why and when he should unflag a property via the vendor's comment

Can you implement this 🙏🏽 ? Thank you.

server/services/property.service.js
```diff
@@ -924,3 +924,46 @@ export const approveProperty = async ({ propertyId, adminId }) => {
...
+  if (!property.flagged.status) {
+    throw new ErrorHandler(httpStatus.PRECONDITION_FAILED, 'Property is not flagged');
+  }
+
+  try {
+    const flaggedProperty = await Property.findByIdAndUpdate(
+      property._id,
+      {
+        $set: {
+          'flagged.requestUnflag': true,
```
andela-hpopoola, line 950: As long as we are setting this to true here, then we should set it back to false when an admin unflags the property.

test/controllers/property.controller.test.js
```diff
@@ -4816,4 +4836,112 @@ describe('Property Controller', () => {
...
+        request()
+          [method](endpoint)
+          .set('authorization', vendorToken)
+          .send(propertyInfo)
+          .end((err, res) => {
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body.message).to.be.eql('Property unflag request sent');
+            expect(res.body.property._id).to.be.eql(property._id.toString());
+            expect(res.body.property.flagged.requestUnflag).to.be.eql(true);
```
andela-hpopoola, line 4882: Let's add an extra expect statement here to confirm that `flagged.status` is still false.

test/controllers/property.controller.test.js
```diff
@@ -4816,4 +4836,112 @@ describe('Property Controller', () => {
...
+              propertyInfo.comment,
+            );
+            done();
+          });
+      });
+    });
+
+    context('property is not flagged', () => {
+      beforeEach(async () => {
+        await Property.findByIdAndUpdate(property._id, { 'flagged.status': false });
```
andela-hpopoola, line 4893: Can you try to avoid doing things like this for your tests. I believe it would increase the time the test would run since we are adding a record and updating the same record. 

In this case, it would be advisable to update the property object before adding it to the database. That way, we are making just 1 `inserts` with no `updates`.

andela-hpopoola (approved):
Looks good to me. Thank you for working on this. I left a few comments of ensuring that the introduced fields are property synced with previous fields and improving the performance of your tests. Feel free to merge after fixing.

PR #185, opened 2021-06-10
#178491462 add badges

tobieeniafe:
#### What does this PR do?
Add related schema, routes and services for badges

Add related schema, routes, and services for badges

This can be tested by...

#178491462

### DIFF
--- a/server/helpers/constants.js
+++ b/server/helpers/constants.js
@@ -122,7 +122,7 @@ export const ACTIVE_PORTFOLIO_OFFER = [
 export const REFERRAL_PERCENTAGE = 1.5;
 
 export const BADGE_ACCESS_LEVEL = {
-  ALL: 0,
-  USER: 1,
-  VENDOR: 2,
+  ALL: -1,
+  USER: USER_ROLE.USER,
+  VENDOR: USER_ROLE.VENDOR,
 };
--- a/server/helpers/constants.js
+++ b/server/helpers/constants.js
@@ -120,3 +120,9 @@ export const ACTIVE_PORTFOLIO_OFFER = [
 ];
 
 export const REFERRAL_PERCENTAGE = 1.5;
+
+export const BADGE_ACCESS_LEVEL = {
+  ALL: 0,
+  USER: 1,
+  VENDOR: 2,
+};
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -362,8 +362,8 @@ export const REFERRAL_FILTERS = {
 };
 
 export const BADGE_FILTERS = {
-  createdAt: { type: FILTER_TYPE.DATE },
   addedBy: { type: FILTER_TYPE.OBJECT_ID },
   assignedRole: { type: FILTER_TYPE.INTEGER },
+  createdAt: { type: FILTER_TYPE.DATE },
   name: { type: FILTER_TYPE.STRING },
 };
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -365,4 +365,5 @@ export const BADGE_FILTERS = {
   createdAt: { type: FILTER_TYPE.DATE },
   addedBy: { type: FILTER_TYPE.OBJECT_ID },
   assignedRole: { type: FILTER_TYPE.INTEGER },
+  name: { type: FILTER_TYPE.STRING },
 };
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -361,4 +361,8 @@ export const REFERRAL_FILTERS = {
   paidOn: { key: 'reward.paidOn', type: FILTER_TYPE.DATE },
 };
 
-export const BADGE_FILTERS = {};
+export const BADGE_FILTERS = {
+  createdAt: { type: FILTER_TYPE.DATE },
+  addedBy: { type: FILTER_TYPE.OBJECT_ID },
+  assignedRole: { type: FILTER_TYPE.INTEGER },
+};
--- a/server/helpers/filters.js
+++ b/server/helpers/filters.js
@@ -360,3 +360,5 @@ export const REFERRAL_FILTERS = {
   paidBy: { key: 'reward.paidBy', type: FILTER_TYPE.OBJECT_ID },
   paidOn: { key: 'reward.paidOn', type: FILTER_TYPE.DATE },
 };
+
+export const BADGE_FILTERS = {};
--- a/server/models/assignedBadge.model.js
+++ b/server/models/assignedBadge.model.js
@@ -1,0 +1,54 @@
+import mongoose from 'mongoose';
+
+/**
+ * @swagger
+ *  components:
+ *    schemas:
+ *      AssignedBadge:
+ *        type: object
+ *        required:
+ *          - badgeId
+ *          - userId
+ *          - assignedBy
+ *          - dateAssigned
+ *        properties:
+ *          badgeId:
+ *            type: string
+ *          userId:
+ *            type: string
+ *          assignedBy:
+ *            type: string
+ *          dateAssigned:
+ *            type: date
+ *        example:
+ *           badgeId: 5f2b39035a086cfc4b7fa7f6
+ *           userId: 5f5a71e1e26485102a780795
+ *           assignedBy: 5f2b39035a086cfc4b7fa7f6
+ *           dateAssigned: 2020-11-20
+ */
+
+const { ObjectId } = mongoose.Schema.Types;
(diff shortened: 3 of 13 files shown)


server/helpers/constants.js
```diff
@@ -120,3 +120,9 @@ export const ACTIVE_PORTFOLIO_OFFER = [
 ];
 
 export const REFERRAL_PERCENTAGE = 1.5;
+
+export const BADGE_ACCESS_LEVEL = {
+  ALL: 0,
+  USER: 1,
```
andela-hpopoola, line 126: I think it is better to link this directly with the user and vendor role.
```suggestion
  USER: USER_ROLE.USER,
```

server/helpers/constants.js
```diff
@@ -120,3 +120,9 @@ export const ACTIVE_PORTFOLIO_OFFER = [
 ];
 
 export const REFERRAL_PERCENTAGE = 1.5;
+
+export const BADGE_ACCESS_LEVEL = {
+  ALL: 0,
```
andela-hpopoola, line 125: Since we can't have a role of `-1`, I think it would be a better alternative to avoid using the same role value as an admin.
```suggestion
  ALL: -1,
```

server/helpers/filters.js
```diff
@@ -360,3 +360,5 @@ export const REFERRAL_FILTERS = {
   paidBy: { key: 'reward.paidBy', type: FILTER_TYPE.OBJECT_ID },
   paidOn: { key: 'reward.paidOn', type: FILTER_TYPE.DATE },
 };
+
+export const BADGE_FILTERS = {};
```
andela-hpopoola, line 364: Can we add the existing filters it the model. We can always add more when we update the model.

server/models/assignedBadge.model.js
```diff
@@ -0,0 +1,54 @@
...
+  {
+    badgeId: {
+      type: ObjectId,
+      required: true,
+    },
+    userId: {
+      type: ObjectId,
+      required: true,
+    },
+    addedBy: {
```
andela-hpopoola, line 41: I think we should make this `optional`. Remember, we would want to assign some badges automatically in the future without an admin intervention. Something like assigning badges when a vendor sells over `10` properties e.t.c

server/models/badge.model.js
```diff
@@ -0,0 +1,55 @@
...
+    },
+    userLevel: {
+      type: Number,
+      default: BADGE_ACCESS_LEVEL.ALL,
+    },
+    addedBy: {
+      type: ObjectId,
+      required: true,
+    },
+    dateAdded: {
```
andela-hpopoola, line 46: I think the `createdAt` in the timestamp is the same as `dateAdded`. I believe we can remove this.

server/routes/badge.routes.js
```diff
@@ -0,0 +1,110 @@
...
+  '/',
+  authenticate,
+  isAdmin,
+  schemaValidation(updateBadgeSchema),
+  BadgeController.updateBadge,
+);
+
+/**
+ * @swagger
+ * /badge/delete/:id:
```
andela-hpopoola, line 66: We should also add endpoints to `getOne` and `edit` badges

server/routes/badge.routes.js
```diff
@@ -0,0 +1,110 @@
...
+ *          schema:
+ *            $ref: '#/components/schemas/Badge'
+ *      description: Deletes specific badge
+ *     responses:
+ *      '200':
+ *        description: Badge deleted
+ *      '500':
+ *       description: Internal server error
+ */
+router.delete('/delete/:id', authenticate, hasValidObjectId, isAdmin, BadgeController.deleteBadge);
```
andela-hpopoola, line 85: We can remove the `delete` word here since this endpoint would only be called if we use the `delete` method.
```suggestion
router.delete('/:id', authenticate, hasValidObjectId, isAdmin, BadgeController.deleteBadge);
```

server/services/badge.service.js
```diff
@@ -0,0 +1,99 @@
...
+
+export const updateBadge = async (updatedBadge) => {
+  const badge = await getBadgeById(updatedBadge.id).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+  if (!badge) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Badge not found');
+  }
+
+  const badgePayload = {
```
andela-hpopoola, line 30: I am not sure I understand why you did this, however, it could have been better to update as:

```js
const badgePayload = {
   ...badge,
   ...updatedBadge
};
```

server/models/badge.model.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+const { ObjectId } = mongoose.Schema.Types;
+const BadgeSchema = new mongoose.Schema(
+  {
+    name: {
+      type: String,
+      required: true,
+    },
+    userLevel: {
```
andela-hpopoola, line 38:
```suggestion
    assignedRole: {
```

server/schemas/badge.schema.js
```diff
@@ -0,0 +1,19 @@
+import Joi from '@hapi/joi';
+import { requiredObjectId, requiredString, optionalString } from './helper.schema';
+import { BADGE_ACCESS_LEVEL } from '../helpers/constants';
+
+const userLevel = Joi.number()
+  .label('Step')
```
andela-hpopoola, line 6: Using `Step` here looks awkward and difficult to understand. Let's update the `userLevel` to `assignedRole` to make it more relatable.

```suggestion
  .label('Role')
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments but I think it is better to defer all `assignedBadges` features to another PR so that we can focus solely on Badges only for this PR. This will make it easier for it to review and understand how both models relate with each other. 

Nice work.

server/helpers/filters.js
```diff
@@ -361,4 +361,8 @@ export const REFERRAL_FILTERS = {
   paidOn: { key: 'reward.paidOn', type: FILTER_TYPE.DATE },
 };
 
-export const BADGE_FILTERS = {};
+export const BADGE_FILTERS = {
+  createdAt: { type: FILTER_TYPE.DATE },
+  addedBy: { type: FILTER_TYPE.OBJECT_ID },
+  assignedRole: { type: FILTER_TYPE.INTEGER },
```
andela-hpopoola, line 367: Not sure why you omitted the Badge name which is the most important filter here.

andela-hpopoola (changes requested):
Thank you for helping out with this. Once you add the tests, this should be good to go.

PS: Sorry for the late review, I just remembered that I have a PR to review.

server/services/badge.service.js
```diff
@@ -28,14 +28,15 @@ export const updateBadge = async (updatedBadge) => {
   }
 
   const badgePayload = {
-    ...badge,
+    // eslint-disable-next-line no-underscore-dangle
+    ...badge._doc,
```
andela-hpopoola, line 32: Is it possible to use another thing about from `_doc`. Maybe `.toJson()` or any other method that works. Apart from _id, it is not advisable to use variable names with `_`

server/helpers/filters.js
```diff
@@ -360,3 +360,10 @@ export const REFERRAL_FILTERS = {
   paidBy: { key: 'reward.paidBy', type: FILTER_TYPE.OBJECT_ID },
   paidOn: { key: 'reward.paidOn', type: FILTER_TYPE.DATE },
 };
+
+export const BADGE_FILTERS = {
+  createdAt: { type: FILTER_TYPE.DATE },
```
andela-hpopoola, line 365: Lets arrange this in alphabetical order 🙏🏽

server/services/badge.service.js
```diff
@@ -0,0 +1,109 @@
...
+
+  if (!badge) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Badge not found');
+  }
+
+  const attachedUsers = await getAssignedBadgesByBadgeId().catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (attachedUsers.length > 0) {
```
andela-hpopoola, line 56: Can we add a test for this 🙏🏽

andela-hpopoola (approved):
Thank you for working on this. Looks good to me. Feel free to merge after fixing the comments I left. Thank you.

PR #188, opened 2021-06-17
#178564163 restructure seeding process

tobieeniafe:
#### What does this PR do?
Restructure seeding process

Restructure seeding process

`npm run seed`

#178564163

### DIFF
--- a/seeding/defaults.seeding.js
+++ b/seeding/defaults.seeding.js
@@ -1,9 +1,9 @@
 export const CONFIG = {
   LIMIT: 1,
   MODEL: '', // reset, user
-  ROLE: 'vendor', // user, admin, vendor, editor
+  ROLE: 'user', // user, admin, vendor, editor
 };
 
 export const USER_DEFAULTS = {
-  password: 'passworded',
+  email: 'test@hello.com',
 };
--- a/seeding/defaults.seeding.js
+++ b/seeding/defaults.seeding.js
@@ -1,0 +1,9 @@
+export const CONFIG = {
+  LIMIT: 1,
+  MODEL: 'reset', // reset, user
+  ROLE: 'user', // user, admin, vendor, editor
+};
+
+export const USER_DEFAULTS = {
+  PASSWORD: 'passworded',
+};
--- a/seeding/defaults.seeding.js
+++ b/seeding/defaults.seeding.js
@@ -4,6 +4,4 @@ export const CONFIG = {
   ROLE: 'user', // user, admin, vendor, editor
 };
 
-export const USER_DEFAULTS = {
-  email: 'test@hello.com',
-};
+export const USER_DEFAULTS = {};
--- a/seeding/defaults.seeding.js
+++ b/seeding/defaults.seeding.js
@@ -1,9 +1,9 @@
 export const CONFIG = {
   LIMIT: 1,
-  MODEL: 'reset', // reset, user
-  ROLE: 'user', // user, admin, vendor, editor
+  MODEL: '', // reset, user
+  ROLE: 'vendor', // user, admin, vendor, editor
 };
 
 export const USER_DEFAULTS = {
-  PASSWORD: 'passworded',
+  password: 'passworded',
 };
--- a/seeding/index.js
+++ b/seeding/index.js
@@ -3,7 +3,7 @@ import { logSuccess, logError } from './seed-helpers';
 import { MODEL } from './seed-constants';
 import { seedUsers } from './user.seeding';
 import resetDB from './reset.seeding';
-import { CONFIG } from './defaults.seeding';
+import { CONFIG, USER_DEFAULTS } from './defaults.seeding';
 
 const NODE_ENV = process.env.NODE_ENV || 'development';
 
@@ -17,7 +17,7 @@ const seedDB = async () => {
   try {
     switch (model) {
       case MODEL.USER:
-        await seedUsers(limit, role);
+        await seedUsers(limit, role, USER_DEFAULTS);
         break;
       case MODEL.RESET:
         await resetDB();
--- a/seeding/index.js
+++ b/seeding/index.js
@@ -3,9 +3,13 @@ import { logSuccess, logError } from './seed-helpers';
 import { MODEL } from './seed-constants';
 import { seedUsers } from './user.seeding';
 import resetDB from './reset.seeding';
-import { model, limit, role } from './seed-config';
+import { CONFIG } from './defaults.seeding';
 
 const NODE_ENV = process.env.NODE_ENV || 'development';
+
+const model = (process.env.MODEL || CONFIG.MODEL).toUpperCase();
+const limit = process.env.LIMIT || CONFIG.LIMIT;
+const role = process.env.ROLE || CONFIG.ROLE;
 
 const seedDB = async () => {
   startDBConnection();
--- a/seeding/index.js
+++ b/seeding/index.js
@@ -3,7 +3,7 @@ import { logSuccess, logError } from './seed-helpers';
 import { MODEL } from './seed-constants';
 import { seedUsers } from './user.seeding';
 import resetDB from './reset.seeding';
-import { CONFIG, USER_DEFAULTS } from './defaults.seeding';
+import { CONFIG } from './defaults.seeding';
 
 const NODE_ENV = process.env.NODE_ENV || 'development';
 
@@ -17,7 +17,7 @@ const seedDB = async () => {
   try {
     switch (model) {
       case MODEL.USER:
-        await seedUsers(limit, role, USER_DEFAULTS);
+        await seedUsers(limit, role);
         break;
       case MODEL.RESET:
         await resetDB();
--- a/seeding/index.js
+++ b/seeding/index.js
@@ -1,16 +1,13 @@
 import { startDBConnection, closeDBConnection } from '../server/config';
 import { logSuccess, logError } from './seed-helpers';
-import { MODEL, ROLE } from './seed-constants';
+import { MODEL } from './seed-constants';
 import { seedUsers } from './user.seeding';
 import resetDB from './reset.seeding';
+import { model, limit, role } from './seed-config';
 
 const NODE_ENV = process.env.NODE_ENV || 'development';
 
 const seedDB = async () => {
-  const model = process.env.MODEL?.toUpperCase();
-  const limit = process.env.LIMIT || 1;
-  const role = process.env.ROLE || ROLE.USER;
-
   startDBConnection();
 
   try {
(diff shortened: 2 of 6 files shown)


seeding/user.seeding.js
```diff
@@ -15,25 +15,26 @@ export const seedUsers = async (limit, role, customValues = {}) => {
...
-      accountNumber: faker.finance.routingNumber(),
-      bankName: faker.company.companyName(),
-    },
-    directors: [
-      {
-        name: faker.name.findName(),
-        isSignatory: true,
-        signature: faker.image.abstract(),
-        phone: faker.phone.phoneNumber(),
+  const vendorInfo = () =>
```
andela-hpopoola, line 18:
```suggestion
  const getVendorInfo = () =>
```

seeding/index.js
```diff
@@ -1,16 +1,13 @@
...
-import { MODEL, ROLE } from './seed-constants';
+import { MODEL } from './seed-constants';
 import { seedUsers } from './user.seeding';
 import resetDB from './reset.seeding';
+import { model, limit, role } from './seed-config';
 
 const NODE_ENV = process.env.NODE_ENV || 'development';
 
 const seedDB = async () => {
-  const model = process.env.MODEL?.toUpperCase();
```
andela-hpopoola, line 10: Lets return this as the config file would be used for a setting defaults only.

seeding/seed-config.js
```diff
@@ -0,0 +1,3 @@
+export const model = (process.env.MODEL || 'user').toUpperCase(); // reset, user
```
andela-hpopoola, line 1: Nice implementation here, however, this can sure be improved 😉 

The first suggestion is to rename this `defaults.seeding.js`.

Also, it is better to use a default config like this:

```js
export const config = {
   limit: 1,
   model: 'user', // reset, user
   role: 'user', // user, admin, vendor, editor
}

 // this would contain all defaults for users
export const user = {
       password: 'passworded'
}

```

For the config, we can then replace their usage in the code. Something like this for `index.js`:

```js
const model = (process.env.MODEL || config.user).toUpperCase();
export const limit = process.env.LIMIT || config.limit;
export const role = process.env.ROLE || config.role;
```

This way, we have 2 ways to set configs -- command line and via `defaults.seeding.js` file

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments but it looks like you omitted the fix for the `clearDatabase()` functionality not working as expected. Can you fix that too 🙏🏽 ?

seeding/defaults.seeding.js
```diff
@@ -0,0 +1,9 @@
+export const CONFIG = {
+  LIMIT: 1,
+  MODEL: 'reset', // reset, user
```
andela-hpopoola, line 3: Let's make this empty since you don't want people to mistakenly reset their precious database just because they forgot to update the configuration.

seeding/user.seeding.js
```diff
@@ -44,7 +46,7 @@ export const seedUsers = async (limit, role, customValues = {}) => {
       lastName: faker.name.lastName(),
       email: faker.internet.email(),
       phone: faker.phone.phoneNumber(),
-      referralCode: `${faker.lorem.word()}${index}`,
+      referralCode: `${faker.lorem.word()}${Math.random() + index}`,
```
andela-hpopoola, line 49: Math.random generates a random decimal value between 0 and 1. So this will generate a referral code like  `word1.5822424812684586` which is clearly not what we wanted. 

Can we generate the referral code the same way we generated it in the code: 
firstName + lastName + 4 random numbers or random thousand number

I know this is also not perfect but it is better than the implementation above. I don't mind if you have a better implementation than this too.

seeding/user.seeding.js
```diff
@@ -68,7 +70,10 @@ export const seedUsers = async (limit, role, customValues = {}) => {
 
       const table = [];
       users.forEach((user) => {
-        table.push({ Email: user.email, Password: customValues.password || DEFAULT_PASSWORD });
+        table.push({
+          Email: user.email,
+          Password: customValues.password || USER_DEFAULTS.PASSWORD,
```
andela-hpopoola, line 75: Looking at this again, I think it is better to have a fallback password incase the user removes the password when updating the user defaults.

```suggestion
          Password: customValues.password || USER_DEFAULTS?.PASSWORD || DEFAULT_PASSWORD,
```

andela-hpopoola (changes requested):
Thank you for working on this. The last thing is for us to do here is to ensure that users can set random defaults and they would be saved in the database.

Here are the catches:
- We want to make it easier in a case where we want to add a user with an email and password, we could just update the user defaults as 

```
{
   email: 'testing@email.com',
    password: '123456',
}
```
- We should not add field names that doesn't exist (this would occur mainly due to typo as done below)

```
{
 emaill: 'stilltesting@yahoo.com',
 pasword: '123456',
}
```

3. We should have the luxury to update a single field for vendors, so that other parts are auto generated. Something like this:

```
{
 vendor: {
    companyName: 'Test'
 }
}
```

seeding/defaults.seeding.js
```diff
@@ -1,9 +1,9 @@
 export const CONFIG = {
   LIMIT: 1,
-  MODEL: 'reset', // reset, user
-  ROLE: 'user', // user, admin, vendor, editor
+  MODEL: '', // reset, user
+  ROLE: 'vendor', // user, admin, vendor, editor
```
andela-hpopoola, line 4: This should be 'user' by default.

seeding/defaults.seeding.js
```diff
@@ -1,9 +1,9 @@
...
 };
 
 export const USER_DEFAULTS = {
-  PASSWORD: 'passworded',
+  password: 'passworded',
```
andela-hpopoola, line 8: Lets remove this as discussed since we are setting a default password.

seeding/index.js
```diff
@@ -17,7 +17,7 @@ const seedDB = async () => {
   try {
     switch (model) {
       case MODEL.USER:
-        await seedUsers(limit, role);
+        await seedUsers(limit, role, USER_DEFAULTS);
```
andela-hpopoola, line 20: It looks like adding the USER_DEAFAULTS constant here. 

I think we can assign USER_DEFAULTS to customValues by default in the `user.seeding.js` file

seeding/user.seeding.js
```diff
@@ -6,15 +6,17 @@ import AddressFactory from '../test/factories/address.factory';
...
 import { USER_ROLE } from '../server/helpers/constants';
 import { logLoading, logError, logTable } from './seed-helpers';
-import { ROLE } from './seed-constants';
-import { hashPassword } from '../server/services/user.service';
+import { ROLE, DEFAULT_PASSWORD } from './seed-constants';
+import { hashPassword, generateReferralCode } from '../server/services/user.service';
 import { USER_DEFAULTS } from './defaults.seeding';
 
 export const seedUsers = async (limit, role, customValues = {}) => {
   const roleValue = Object.keys(ROLE).find((key) => ROLE[key] === role);
```
andela-hpopoola, line 15: I think it is better to extend customValues here instead of overriding the total values. Something like this.

```js
const userDefaults = { ...customValues, ...USER_DEFAULTS };
```

`userDefaults` should then be used to replace all instances of `customValues`

seeding/user.seeding.js
```diff
@@ -6,15 +6,17 @@ import AddressFactory from '../test/factories/address.factory';
...
 
-  const password = customValues.password || USER_DEFAULTS.PASSWORD;
+  const password = customValues.password || USER_DEFAULTS?.password || DEFAULT_PASSWORD;
```
andela-hpopoola, line 16: If you implement the `userDefaults` suggestion above, then this can be rewritten as:

```suggestion
const password = userDefaults?.password || DEFAULT_PASSWORD;
```

seeding/user.seeding.js
```diff
@@ -6,15 +6,17 @@ import AddressFactory from '../test/factories/address.factory';
...
   const hashedPassword = await hashPassword(password);
+  const firstName = faker.name.firstName();
```
andela-hpopoola, line 18: This is wrong when we are generating for multiple users - it would give the same error you fixed for vendor here as it would keep using the same firstName irrespective of the number of users we wish to generate.

seeding/user.seeding.js
```diff
@@ -55,8 +58,8 @@ export const seedUsers = async (limit, role, customValues = {}) => {
         street1: faker.address.streetName(),
         street2: faker.address.streetName(),
       }),
-      vendor: parseInt(roleValue, 10) === USER_ROLE.VENDOR ? getVendorInfo() : {},
       ...customValues,
+      vendor: parseInt(roleValue, 10) === USER_ROLE.VENDOR ? getVendorInfo() : {},
```
andela-hpopoola, line 62: Since we are setting DEFAULT values for users, you might want to reconsider your usage of `roleValue` to also cater for the time when a user would input a default role.

andela-hpopoola (changes requested):
Thank you for working on this. I believe some of the comments I dropped would have been caught if you tested this more. Can you try to do more manual testing to ensure that we are implementing this right?

seeding/defaults.seeding.js
```diff
@@ -0,0 +1,9 @@
+export const CONFIG = {
+  LIMIT: 1,
+  MODEL: '', // reset, user
+  ROLE: 'user', // user, admin, vendor, editor
+};
+
+export const USER_DEFAULTS = {
+  email: 'test@hello.com',
```
andela-hpopoola, line 8: I think you might need to remove this

seeding/user.seeding.js
```diff
@@ -7,44 +7,50 @@ import User from '../server/models/user.model';
...
 export const seedUsers = async (limit, role, customValues = {}) => {
-  const roleValue = Object.keys(ROLE).find((key) => ROLE[key] === role);
+  const userDefaults = { ...customValues, ...USER_DEFAULTS };
+  const roleValue = Object.keys(ROLE).find((key) => ROLE[key] === (userDefaults?.role || role));
 
-  const password = customValues.password || DEFAULT_PASSWORD;
+  const password = userDefaults?.password || DEFAULT_PASSWORD;
   const hashedPassword = await hashPassword(password);
+  const getFirstName = () => faker.name.firstName();
+  const referralCode = await generateReferralCode(getFirstName());
```
andela-hpopoola, line 20: We might want to generate the referralCode from the actual first name gotten from `getFirstName` instead of calling the function again.

seeding/user.seeding.js
```diff
@@ -68,7 +74,10 @@ export const seedUsers = async (limit, role, customValues = {}) => {
 
       const table = [];
       users.forEach((user) => {
-        table.push({ Email: user.email, Password: customValues.password || DEFAULT_PASSWORD });
+        table.push({
+          Email: user.email,
+          Password: userDefaults.password || DEFAULT_PASSWORD,
```
andela-hpopoola, line 79: I think this can be replaced with `password` which is defined above.

```suggestion
          Password: password,
```

andela-hpopoola (approved):
Thank you for working on this. I think this is good to go for now, so feel free to merge it after fixing the minor nitpicks mentioned below. If any other thing comes up again, we can surely fix them in another Pull Request. Thank you.

PR #189, opened 2021-06-18
#178088003 create seeding for properties

tobieeniafe:
#### What does this PR do?
Create seeding for properties

Create seeding for properties

`npm run test`

#178088003

### DIFF
--- a/seeding/property.seeding.js
+++ b/seeding/property.seeding.js
@@ -1,4 +1,5 @@
-import PropertyFactory from '../test/factories/property.factory';
+/* eslint-disable import/no-extraneous-dependencies */
+import faker from 'faker';
 import Property from '../server/models/property.model';
 import User from '../server/models/user.model';
 import { USER_ROLE } from '../server/helpers/constants';
@@ -10,7 +11,7 @@ import {
   getMultipleRandomItems,
   getNumberWithinRange,
 } from './seed-helpers';
-import { HOUSE_TYPES, DEFAULT_PROPERTY_FEATURES, ADDRESSES } from './seed-constants';
+import { HOUSE_TYPES, DEFAULT_PROPERTY_FEATURES, ADDRESSES, NEIGHBORHOODS } from './seed-constants';
 import { seedUsers } from './user.seeding';
 import { USER_DEFAULTS } from './defaults.seeding';
 
@@ -39,10 +40,27 @@ const seedProperties = async (limit, customValues = {}) => {
 
     const features = getMultipleRandomItems(DEFAULT_PROPERTY_FEATURES, numberOfFeatures);
 
-    return PropertyFactory.build({
+    const address = getOneRandomItem(ADDRESSES);
+
+    const gallerySize = getNumberWithinRange(0, 5);
+    const gallery = new Array(gallerySize).fill().map((_, index) => {
+      return { title: `gallery image ${index}`, url: faker.image.business() };
+    });
+
+    const floorPlanSize = getNumberWithinRange(0, 2);
+    const floorPlans = new Array(floorPlanSize).fill().map((_, index) => {
+      return { name: `floor plan ${index}`, plan: faker.image.city() };
+    });
+
+    const neighborhoodSize = getNumberWithinRange(0, 6);
+    const neighborhood = Object.fromEntries(
+      getMultipleRandomItems(Object.entries(NEIGHBORHOODS), neighborhoodSize),
+    );
+
+    return {
       address: {
         country: 'Nigeria',
-        ...getOneRandomItem(ADDRESSES),
+        ...address,
       },
       flagged: { status: false, requestUnflag: false },
       approved: { status: true },
@@ -56,8 +74,13 @@ const seedProperties = async (limit, customValues = {}) => {
       price: getNumberWithinRange(10_000_000, 50_000_000, 1_000_000),
       toilets: rooms + 1,
       units: getNumberWithinRange(1, 10),
+      mainImage: faker.image.business(),
+      gallery,
+      neighborhood: address.state === 'Lagos' ? neighborhood : {},
+      mapLocation: address.mapLocation,
+      floorPlans,
       ...customValues,
-    });
+    };
   });
 
   try {
--- a/seeding/property.seeding.js
+++ b/seeding/property.seeding.js
@@ -6,17 +6,22 @@ import {
   logLoading,
   logError,
   logTable,
-  returnOneItemFromArray,
-  returnMultipleItemsFromArray,
-  generateNumberWithinRange,
+  getOneRandomItem,
+  getMultipleRandomItems,
+  getNumberWithinRange,
 } from './seed-helpers';
 import { HOUSE_TYPES, DEFAULT_PROPERTY_FEATURES, ADDRESSES } from './seed-constants';
 import { seedUsers } from './user.seeding';
+import { USER_DEFAULTS } from './defaults.seeding';
 
 const seedProperties = async (limit, customValues = {}) => {
   let vendor;
 
-  [vendor] = await User.find({ role: USER_ROLE.VENDOR }).sort({ _id: -1 }).limit(1);
+  const parsedLimit = parseInt(limit, 10);
+
+  [vendor] = await User.find({ ...USER_DEFAULTS, role: USER_ROLE.VENDOR })
+    .sort({ _id: -1 })
+    .limit(1);
 
   if (!vendor) {
     try {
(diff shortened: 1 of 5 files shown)


seeding/property.seeding.js
```diff
@@ -0,0 +1,64 @@
...
+    try {
+      await User.create(vendor);
+    } catch (error) {
+      logError(error);
+    }
+  }
+
+  const properties = [...new Array(parseInt(limit, 10))].map(() =>
+    PropertyFactory.build({
+      address: AddressFactory.build({
```
andela-hpopoola, line 31: Can we try to make this a valid Nigerian address (because of the property map). Also, I don't think we need to use a Factory here.

To make this more accurate, can you get a list of 15 different Nigerian addresses so that we can randomize getting one of them for this. The properties on Ballers would be based in Nigeria, so it makes sense to get Nigerian addresses for this to make it more realistic.

seeding/property.seeding.js
```diff
@@ -0,0 +1,64 @@
...
+        street2: faker.address.streetName(),
+      }),
+      flagged: { status: false, requestUnflag: false },
+      approved: { status: true },
+      addedBy: vendor._id,
+      bathrooms: Math.floor(Math.random() * 5),
+      bedrooms: Math.floor(Math.random() * 5),
+      features: ['swimming pool', 'tiled roads', 'electricity'],
+      houseType: 'penthouse apartment',
+      name: 'penthouse apartment',
```
andela-hpopoola, line 45: We need a better way to generate the property name.

seeding/property.seeding.js
```diff
@@ -0,0 +1,64 @@
...
+        street1: faker.address.streetName(),
+        street2: faker.address.streetName(),
+      }),
+      flagged: { status: false, requestUnflag: false },
+      approved: { status: true },
+      addedBy: vendor._id,
+      bathrooms: Math.floor(Math.random() * 5),
+      bedrooms: Math.floor(Math.random() * 5),
+      features: ['swimming pool', 'tiled roads', 'electricity'],
+      houseType: 'penthouse apartment',
```
andela-hpopoola, line 44: These are the house types used in the frontend, can we randomly get one.

```js
export const HOUSE_TYPES = [
  'Bungalow',
  'Detached Duplex',
  'Flat',
  'Maisonette',
  'Penthouse',
  'Semi-detached Bungalow',
  'Semi-detached Duplex',
  'Studio Apartment',
  'Terraced Bungalow',
  'Terraced Duplex',
];
```

seeding/property.seeding.js
```diff
@@ -0,0 +1,64 @@
...
+        state: faker.address.state(),
+        street1: faker.address.streetName(),
+        street2: faker.address.streetName(),
+      }),
+      flagged: { status: false, requestUnflag: false },
+      approved: { status: true },
+      addedBy: vendor._id,
+      bathrooms: Math.floor(Math.random() * 5),
+      bedrooms: Math.floor(Math.random() * 5),
+      features: ['swimming pool', 'tiled roads', 'electricity'],
```
andela-hpopoola, line 43: Here are the features used in the frontend, can we randomly get 3:

```js
export const DEFAULT_PROPERTY_FEATURES = [
  'Car Parking',
  'Guest Toilet',
  'Guest Room',
  "Governor's Consent",
  'Electricity',
  'Paved Roads',
  'Perimeter Fence',
  'Portable Water',
  'Fully Tiled',
  'Ensuite Rooms',
  'Easy Access to Roads',
  'Cable TV Distibution',
  'Core Fiber Internet',
  'Inverter System',
  'Security Fence',
  'Parking Lot',
  'Dedicated Car Port',
  'Maid Room',
  'Surveillance System',
  'Smart Solar System',
  'Panic Alarm',
  'Intercom System',
  'Spacious Kitchen',
  'Video door phone',
  'Fire detection',
  'Swimming Pool',
  'Rooftop Gym',
  'Garage',
  'Free WiFi',
  'Peaceful Environment',
  'Fitted Kitchen',
];
```

seeding/property.seeding.js
```diff
@@ -0,0 +1,64 @@
...
+import { USER_ROLE } from '../server/helpers/constants';
+import { logLoading, logError, logTable } from './seed-helpers';
+
+const seedProperties = async (limit, customValues = {}) => {
+  let vendor;
+
+  [vendor] = await User.find({ role: USER_ROLE.VENDOR }).sort({ _id: -1 }).limit(1);
+
+  if (!vendor) {
+    vendor = UserFactory.build(
```
andela-hpopoola, line 17: I think it might be better to use the user seeding service here. Also, you didn't create an avenue for us to input a default vendor email or company name. This should be done from the command line or via USER_DEFAULTS.

seeding/property.seeding.js
```diff
@@ -0,0 +1,64 @@
...
+      }),
+      flagged: { status: false, requestUnflag: false },
+      approved: { status: true },
+      addedBy: vendor._id,
+      bathrooms: Math.floor(Math.random() * 5),
+      bedrooms: Math.floor(Math.random() * 5),
+      features: ['swimming pool', 'tiled roads', 'electricity'],
+      houseType: 'penthouse apartment',
+      name: 'penthouse apartment',
+      price: 10_000_000,
```
andela-hpopoola, line 46: Lets make this a random price from 10 million to 50 million (approximated to the nearest million)

seeding/property.seeding.js
```diff
@@ -0,0 +1,64 @@
...
+      toilets: Math.floor(Math.random() * 5),
+      units: Math.floor(Math.random() * 10),
```
andela-hpopoola, line 48: I think `Math.cei`l is better here to avoid having `0`
```suggestion
      units: Math.ceil(Math.random() * 10),
```

seeding/property.seeding.js
```diff
@@ -0,0 +1,64 @@
...
+        city: faker.address.city(),
+        country: faker.address.country(),
+        state: faker.address.state(),
+        street1: faker.address.streetName(),
+        street2: faker.address.streetName(),
+      }),
+      flagged: { status: false, requestUnflag: false },
+      approved: { status: true },
+      addedBy: vendor._id,
+      bathrooms: Math.floor(Math.random() * 5),
```
andela-hpopoola, line 41: Lets make the number of `bathrooms` and `bedrooms` be the same. Lets add 1 to this number and use it for toilets (most recent house has visitor toilets). Let this number range from 2 to 4.

andela-hpopoola (changes requested):
Thank you for working on this. I dropped a few comments on ensuring that we have more accurate information. Please don't forget that we need to set DEFAULTS for properties too. Once you get it right, you should make the same implementation here too.

seeding/property.seeding.js
```diff
@@ -1,54 +1,56 @@
...
+    const houseType = HOUSE_TYPES[Math.floor(Math.random() * HOUSE_TYPES.length)];
+
+    return PropertyFactory.build({
       address: AddressFactory.build({
-        city: faker.address.city(),
-        country: faker.address.country(),
-        state: faker.address.state(),
-        street1: faker.address.streetName(),
-        street2: faker.address.streetName(),
+        city: CITIES[Math.floor(Math.random() * CITIES.length)],
```
andela-hpopoola, line 30: This can be converted into a function and reused for other fields.

seeding/property.seeding.js
```diff
@@ -1,54 +1,56 @@
...
-      features: ['swimming pool', 'tiled roads', 'electricity'],
-      houseType: 'penthouse apartment',
-      name: 'penthouse apartment',
-      price: 10_000_000,
-      toilets: Math.floor(Math.random() * 5),
-      units: Math.floor(Math.random() * 10),
+      bathrooms: rooms,
+      bedrooms: rooms,
+      description: `Newly built ${houseType}`,
+      features: DEFAULT_PROPERTY_FEATURES.sort(() => Math.random() - Math.random()).slice(0, 3),
```
andela-hpopoola, line 42: Not sure why you used sort here (it is completely not necessary). It would have been better to create a function to get a random list of items from an array.

seeding/property.seeding.js
```diff
@@ -1,54 +1,56 @@
...
+      [vendor] = await User.find({ role: USER_ROLE.VENDOR }).sort({ _id: -1 }).limit(1);
     } catch (error) {
       logError(error);
     }
   }
 
-  const properties = [...new Array(parseInt(limit, 10))].map(() =>
-    PropertyFactory.build({
+  const properties = [...new Array(parseInt(limit, 10))].map(() => {
+    const rooms = Math.floor(Math.random() * (4 - 2 + 1)) + 2;
```
andela-hpopoola, line 25: This looks gibberish and difficult to understand what is going on as it contains too much magic numbers. Can this be simplified and created as a function. 

Writting code should be fun and easy to understand 😏

seeding/property.seeding.js
```diff
@@ -1,54 +1,56 @@
...
+    const houseType = HOUSE_TYPES[Math.floor(Math.random() * HOUSE_TYPES.length)];
+
+    return PropertyFactory.build({
       address: AddressFactory.build({
```
andela-hpopoola, line 29: Your implementation of the address is still wrong and not different from using faker's implementation. With the current implementation, a street in Lekki will be mapped to Epe and then mapped to Oyo. The best implementation is to create the whole address as an object and just pick a random one.

Something like this: 

```js
// we would randomly get an address from the list here
const ADDRESS_LIST = [
  {
   city: 'Lekki', // city should be related to state and street
   state: 'Lagos', // other states should be considered too, but we can have more addresses in Lagos state
   street1: 'Ozumba Mbadiwe Avenue',
   street2: '', // some addresses should have street 2, while some would not have
  }, 
...
]
```

Also, I still maintain that we don't need to use a Factory here.

seeding/property.seeding.js
```diff
@@ -1,54 +1,56 @@
...
-      price: 10_000_000,
-      toilets: Math.floor(Math.random() * 5),
-      units: Math.floor(Math.random() * 10),
+      bathrooms: rooms,
+      bedrooms: rooms,
+      description: `Newly built ${houseType}`,
+      features: DEFAULT_PROPERTY_FEATURES.sort(() => Math.random() - Math.random()).slice(0, 3),
+      houseType,
+      name: `Newly built ${houseType}`,
+      price:
```
andela-hpopoola, line 45: This should be implemented with 2 functions:

1. The first function to get the random value from 10 Million to 50 Million. This same function should be used above for `room`.
2. The second function should be used to convert the returned value to the nearest million.

seeding/seed-constants.js
```diff
@@ -13,3 +14,76 @@ export const ROLE = {
...
+  'Epe',
+  'Ibeju Lekki',
+  'Ikate',
+  'Ikoyi',
+  'Lekki Phase 1',
+  'VGC',
+  'Victoria Island',
+];
+
+export const STATES = ['Lagos', 'Ogun', 'Oyo'];
```
andela-hpopoola, line 76: Lets Abuja, Kano, and PortHarcourt to the list. We can remove `Ogun` (no disrespect intended)

andela-hpopoola (changes requested):
Thank you for working on this, however, I think your implementation can be greatly improved. I left a few comments though to help make it easier.

seeding/property.seeding.js
```diff
@@ -0,0 +1,66 @@
...
+  returnMultipleItemsFromArray,
+  generateNumberWithinRange,
+} from './seed-helpers';
+import { HOUSE_TYPES, DEFAULT_PROPERTY_FEATURES, ADDRESSES } from './seed-constants';
+import { seedUsers } from './user.seeding';
+
+const seedProperties = async (limit, customValues = {}) => {
+  let vendor;
+
+  [vendor] = await User.find({ role: USER_ROLE.VENDOR }).sort({ _id: -1 }).limit(1);
```
andela-hpopoola, line 19: The support for a company name or vendor email address is still missing here 😏

andela-hpopoola (changes requested):
Nice work with this. I left more comments to ensure that the generated data is close to perfection. Thank you.

andela-hpopoola:
I noticed you updated the label to `Review Requested`. I have taken a look and your changes make sense. Thank you for making those updates. 

https://github.com/Highrachy/ballers-api/pull/189#discussion_r657643433

Please note that this PR is not a priority, however, I think you would need to implement the above comment to finalize this PR.

andela-hpopoola (approved):
Thank you for working on this. I noticed that you have implemented the most recent fixes, looks good 💯. Let's merge this asap and continue to test it out.

andela-hpopoola (commented):
I just noticed that you have not fixed selecting the vendor the property would be attributed to via company name or email. Let's do that via another Pull Request so that it is easier to keep track of.

PR #191, opened 2021-06-21
#178601149 add bank details to user model

tobieeniafe:
#### What does this PR do?
Add bank details to user model

Add bank details to user model

`PUT /api/v1/user/update`

#178601149

### DIFF
--- a/test/controllers/user.controller.test.js
+++ b/test/controllers/user.controller.test.js
@@ -876,20 +876,6 @@ describe('User Controller', () => {
         });
       });
 
-      context('with invalid updated user', () => {
-        it('returns a updated user', (done) => {
-          request()
-            .put('/api/v1/user/update')
-            .set('authorization', userToken)
-            .end((err, res) => {
-              expect(res).to.have.status(412);
-              expect(res.body.success).to.be.eql(false);
-              expect(res.body.message).to.be.eql('Validation Error');
-              done();
-            });
-        });
-      });
-
       context('when update service returns an error', () => {
         it('returns the error', (done) => {
           sinon.stub(User, 'findOneAndUpdate').throws(new Error('Type Error'));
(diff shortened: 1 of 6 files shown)


andela-hpopoola (changes requested):
Thank you for working on this. I think we need 2 more things to wrap up this PR:

1. Let's return the additionalInfo object only when the user or an admin views the user information
2. Add tests to ensure that a user can correctly add his bank account details via the update endpoint for users.

Thank you.

server/controllers/user.controllers.js
```diff
@@ -116,7 +116,7 @@ const UserController = {
 
   update(req, res, next) {
     const updatedUser = req.locals;
-    updateUser({ ...req.user, ...updatedUser, id: req.user._id })
+    updateUser({ ...updatedUser, id: req.user._id })
```
andela-hpopoola, line 119: Not sure why you are removing this 🤷🏽‍♂️

server/services/user.service.js
```diff
@@ -294,8 +294,15 @@ export const resetPasswordViaToken = async (password, token) => {
 };
 
 export const updateUser = async (updatedUser) => {
+  const user = await getUserById(updatedUser.id);
```
andela-hpopoola, line 297: Getting the user information just to update their information seems wrong. If this is necessary, then the previous implementation seems better than your new implementation. Not sure why you made the changes though.

test/controllers/user.controller.test.js
```diff
@@ -876,20 +876,6 @@ describe('User Controller', () => {
         });
       });
 
-      context('with invalid updated user', () => {
```
andela-hpopoola, line 879: Not sure why you removed this.

andela-hpopoola (changes requested):
Thank you for working on this. This looks good enough but your implementation seems off. Also, I think you missed this:

> Let's return the additionalInfo object only when the user or an admin views the user information

We would also need to add tests for this too.

test/controllers/user.controller.test.js
tobieeniafe: The test was passing because no body was being sent. Now that the fields are all optional, it'd fail.

andela-hpopoola (changes requested):
Thank you for making the changes. I think the PR is still missing this:

> Let's return the additionalInfo object only when the user or an admin views the user information

tobieeniafe:
> Thank you for making the changes. I think the PR is still missing this:
> 
> > Let's return the additionalInfo object only when the user or an admin views the user information

The endpoints `/user/all` and `/user/:id` only list what data not to project. Unless you have another endpoint in mind.

andela-hpopoola:
> > Thank you for making the changes. I think the PR is still missing this:
> > > Let's return the additionalInfo object only when the user or an admin views the user information
> 
> The endpoints `/user/all` and `/user/:id` only list what data not to project. Unless you have another endpoint in mind.

Yes, however we need to consider all the places where the following are used to be sure that the bankDetails are returned only where necessary and for admins and or users:

- NON_PROJECTED_USER_INFO
- PROJECTED_ASSIGNED_USER_INFO
- NON_PROJECTED_USER_INFO
-  projectedReferralInfoForAdmin - For this, we would want to include the bank details of the referral.

Please note that I didn't check the usage of the 3 constants above -- I am leaving it to you to confirm that the bankDetails is safe where they are used.

andela-hpopoola:
Also, we might also write tests (one-liner tests ) to ensure that `additional info` is returned or not returned where it is needed. Something like `expect additionalInfo to (not) be present/ exist/ not empty` should work.

andela-hpopoola (approved):
Thank you for making the changes. Looks good to me.

PR #198, opened 2021-06-30
#178687463 return assigned users when viewing badge

tobieeniafe:
#### What does this PR do?
Return assigned users when viewing badge

Return assigned users when viewing badge

`GET /api/v1/badge/:id`
`GET /api/v1/badge/all`

#178687463

### DIFF
--- a/test/controllers/badge.controller.test.js
+++ b/test/controllers/badge.controller.test.js
@@ -351,31 +351,42 @@ describe('Badge Controller', () => {
   });
 
   describe('Get single badge route', () => {
-    const badge = BadgeFactory.build({}, { generateId: true });
-    const endpoint = `/api/v1/badge/${badge._id}`;
+    const badge1 = BadgeFactory.build({}, { generateId: true });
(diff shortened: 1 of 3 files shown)


test/controllers/badge.controller.test.js
```diff
@@ -354,8 +355,27 @@ describe('Badge Controller', () => {
...
+      {
+        badgeId: badge._id,
+        userId: regularUser._id,
+      },
+      { generateId: true },
+    );
+
     beforeEach(async () => {
       await addBadge(badge);
+      await addUser(vendorUser);
```
andela-hpopoola, line 375: Let's add one additional test for another badge with 1 assigned user to confirm that this works as expected. 

PS: For tests loading stuffs, it is always better to have 2 different tests to confirm that it works as expected.

andela-hpopoola (approved):
Looks good to me. Thank you for working on this. I left a comment on adding an additional test to confirm that this works as expected for different badges.

PR #199, opened 2021-06-30
#178687444 return badges and assigned badges for total count

tobieeniafe:
#### What does this PR do?
Return badges and assigned badges for total count

Return badges and assigned badges for total count

`GET /api/v1/total-count`

#178687444

### DIFF
--- a/server/services/totalCount.service.js
+++ b/server/services/totalCount.service.js
@@ -11,6 +11,8 @@ import ReportedProperty from '../models/reportedProperty.model';
 import Transaction from '../models/transaction.model';
 import User from '../models/user.model';
 import Visitation from '../models/visitation.model';
+import Badge from '../models/badge.model';
+import AssignedBadge from '../models/assignedBadge.model';
 
 const { ObjectId } = mongoose.Types.ObjectId;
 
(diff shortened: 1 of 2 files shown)


server/services/totalCount.service.js
```diff
@@ -42,6 +44,7 @@ const getTotalCount = async (user) => {
       ...matchObject,
     });
     const transactions = await Transaction.countDocuments(matchObject);
+    const assignedbadges = await AssignedBadge.countDocuments({ userId: ObjectId(user._id) });
```
andela-hpopoola, line 47: Typo:
```suggestion
    const assignedBadges = await AssignedBadge.countDocuments({ userId: ObjectId(user._id) });
```

Since this would also be returned for the admin, I think we should also add a test for it.

test/controllers/totalCount.controller.test.js
```diff
@@ -269,6 +303,8 @@ describe('TotalCount Controller', () => {
             expect(res.body.models).to.not.have.property('reportedProperties');
             expect(res.body.models).to.not.have.property('offlinePayments');
             expect(res.body.models.portfolios).to.be.eql(2);
+            expect(res.body.models).to.not.have.property('badges');
+            expect(res.body.models.assignedbadges).to.be.eql(5);
```
andela-hpopoola, line 307: I think we should make this 4 so that it is different from the total badges from the admin to ensure that we catch unintended bugs

andela-hpopoola (approved):
Thank you for working on this. I left a few comments but this looks good to me generally.

PR #201, opened 2021-07-01
#178687478 add accumulated referral to referral model

tobieeniafe:
#### What does this PR do?
Add accumulated referral to referral model

Add accumulated referral to referral model

NA

#178687478

### DIFF
--- a/server/models/referral.model.js
+++ b/server/models/referral.model.js
@@ -62,6 +62,25 @@ const ReferralSchema = new mongoose.Schema(
         type: Date,
       },
     },
+    assignedReferral: {
+      total: {
+        type: Number,
+        default: 0,
+      },
+      transactions: [
+        {
+          transactionId: {
+            type: ObjectId,
+          },
+          percentage: {
+            type: Number,
+          },
+          amountPayable: {
+            type: Number,
+          },
+        },
+      ],
+    },
   },
   { timestamps: true },
 );
--- a/server/models/referral.model.js
+++ b/server/models/referral.model.js
@@ -62,7 +62,7 @@ const ReferralSchema = new mongoose.Schema(
         type: Date,
       },
     },
-    assignedReferral: {
+    accumulatedReward: {
       total: {
         type: Number,
         default: 0,
@@ -75,7 +75,7 @@ const ReferralSchema = new mongoose.Schema(
           percentage: {
             type: Number,
           },
-          amountPayable: {
+          amount: {
             type: Number,
           },
         },
--- a/server/services/referral.service.js
+++ b/server/services/referral.service.js
@@ -343,7 +343,15 @@ export const activatePendingUserReferral = async (offer) => {
   }
 };
 
-export const updateReferralRewardStatus = async ({ referralId, offerId }) => {
+export const getReferralByOfferId = async (offerId) =>
+  Referral.findOne({ offerId: ObjectId(offerId) });
+
+export const updateReferralAccumulatedRewardAndRewardStatus = async ({
+  referralId,
+  transactionId,
+  offerId,
+  amountPaid,
+}) => {
   const paymentType = await getPaymentDuration(offerId);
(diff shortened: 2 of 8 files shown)


server/models/referral.model.js
```diff
@@ -62,6 +62,25 @@ const ReferralSchema = new mongoose.Schema(
         type: Date,
       },
     },
+    assignedReferral: {
```
andela-hpopoola, line 65:
```suggestion
    accumulatedReward: {
```

server/models/referral.model.js
```diff
@@ -62,6 +62,25 @@ const ReferralSchema = new mongoose.Schema(
...
+      },
+      transactions: [
+        {
+          transactionId: {
+            type: ObjectId,
+          },
+          percentage: {
+            type: Number,
+          },
+          amountPayable: {
```
andela-hpopoola, line 78: Changing this for consistency with the amount in reward
```suggestion
          amount: {
```

server/services/referral.service.js
```diff
@@ -364,3 +364,38 @@ export const updateReferralRewardStatus = async ({ referralId, offerId }) => {
...
+  });
+
+  const user = await getUserById(referral.referrerId);
+
+  const percentage = user.additionalInfo.referralPercentage;
+  const amountPayable = (percentage / 100) * amountPaid;
+  const total =
+    amountPayable +
+    referral.assignedReferral.transactions.reduce(
+      (accum, transaction) => accum + transaction.amountPayable,
```
andela-hpopoola, line 380: Not sure why you used `accum`,  `acc` is a better alternative in this case.

```suggestion
      (acc, transaction) => acc + transaction.amountPayable,
```

Also, can we assign the reduce function to another variable like `previousTotal` to make it more understandable

server/services/transaction.service.js
```diff
@@ -73,6 +77,12 @@ export const addTransaction = async (transaction) => {
 
     if (referral) {
       await updateReferralRewardStatus({ referralId: referral._id, offerId: offer._id });
+
+      await updateAssignedReferral({
```
andela-hpopoola, line 81: I think we should update the referral with only one query instead of 2

andela-hpopoola: We need to write tests for this in the transaction model too.

andela-hpopoola (changes requested):
Thank you for working on this. I did a quick review on it now but will review it properly tomorrow.

server/services/referral.service.js
```diff
@@ -355,44 +363,32 @@ export const updateReferralRewardStatus = async ({ referralId, offerId }) => {
...
-    );
+  const amount = (percentage / 100) * amountPaid;
+  const previousTotal = referral.accumulatedReward.transactions.reduce(
+    (acc, transaction) => acc + transaction.amount,
+    0,
+  );
+  const total = amount + previousTotal;
 
-  const newAssignedReferral = {
+  const newAccumulatedReward = {
```
andela-hpopoola, line 380: I think we can leave this as `accumulatedReward`.

```suggestion
  const accumulatedReward = {
```

andela-hpopoola (approved):
Thank you for working on this. I wasn't able to critically analyze this but it looks good to me. If anything comes up when I am implementing this in the front end, I will let you know. Please feel free to merge when ready. 👍🏽

server/models/referral.model.js
```diff
@@ -62,6 +62,25 @@ const ReferralSchema = new mongoose.Schema(
         type: Date,
       },
     },
+    accumulatedReward: {
```
andela-hpopoola, line 65: This is not loaded when getting All Referrals. Can you fix this before merging, please? Can you also confirm that it is loaded in other referral services like when getting a single referral by Id.

andela-hpopoola (approved):
I left a quick comment, please fix before merging. Thank you.

PR #202, opened 2021-07-05
#178769329 fix returned info bug for badges

tobieeniafe:
#### What does this PR do?
Fix returned info bug for badges

- Badges info is missing for a single user (badges name and other necessary info are not returned)
- Icons not returned for, getAll badges endpoint

NA

#178769329

### DIFF
--- a/server/services/user.service.js
+++ b/server/services/user.service.js
@@ -687,6 +687,14 @@ export const getOneUser = async (userId) =>
       },
     },
     {
+      $lookup: {
+        from: 'badges',
+        localField: 'assignedBadges.badgeId',
+        foreignField: '_id',
+        as: 'badges',
+      },
+    },
+    {
       $project: {
         password: 0,
         notifications: 0,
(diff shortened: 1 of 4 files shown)


server/services/user.service.js
```diff
@@ -686,6 +686,14 @@ export const getOneUser = async (userId) =>
         as: 'assignedBadges',
       },
     },
+    {
+      $lookup: {
```
andela-hpopoola, line 690: We need to add some test expectations to confirm that the badge name, icon, and image are properly returned. Testing for the Id only is one of the reasons why we have this kind of bug.

andela-hpopoola (changes requested):

From the PR description, you were supposed to explicitly list out all the things you were fixing. Here is what is in the description box in the Pivotal Tracker:

```
- Badges info is missing for a single user (badges name and other necessary info are not returned)
- User's profile image is missing for assigned image in badges
- Icons not returned for get all badges endpoint
```

The bug for the user profile image has not been fixed too.

andela-hpopoola (approved):
Thank you for the explanation. I have confirmed that the profile image is indeed returned as expected. Approving this Pull Request asap.

PR #204, opened 2021-07-07
#178703269 add automated badges

tobieeniafe:
#### What does this PR do?
Add automated badges

Add automated badges

NA

#178703269

### DIFF
--- a/server/helpers/automatedBadges.js
+++ b/server/helpers/automatedBadges.js
@@ -1,9 +1,45 @@
+import { slugify } from './funtions';
+import { BADGE_ACCESS_LEVEL } from './constants';
+
+const badgeNames = {
+  ACTIVATED_USER: 'Activated User Badge',
+  USER_FIRST_PAYMENT: 'First Payment User Badge',
+  VENDOR_CERTIFIED: 'Certified Vendor Badge',
+  VENDOR_FIRST_SALE: 'First Sale Vendor Badge',
+  VENDOR_VERIFIED: 'Verified Vendor Badge',
+};
+
 const AUTOMATED_BADGES = {
-  ACTIVATED_USER: 'activated-user-badge',
-  VENDOR_VERIFIED: 'verified-vendor-badge',
-  VENDOR_CERTIFIED: 'certified-vendor-badge',
-  VENDOR_FIRST_SALE: 'first-sale-vendor-badge',
-  USER_FIRST_PAYMENT: 'first-payment-user-badge',
+  ACTIVATED_USER: {
+    name: badgeNames.ACTIVATED_USER,
+    slug: slugify(badgeNames.ACTIVATED_USER),
+    assignedRole: BADGE_ACCESS_LEVEL.ALL,
+    icon: { name: 'default', color: '#000000' },
+  },
+  USER_FIRST_PAYMENT: {
+    name: badgeNames.USER_FIRST_PAYMENT,
+    slug: slugify(badgeNames.USER_FIRST_PAYMENT),
+    assignedRole: BADGE_ACCESS_LEVEL.USER,
+    icon: { name: 'default', color: '#000000' },
+  },
+  VENDOR_CERTIFIED: {
+    name: badgeNames.VENDOR_CERTIFIED,
+    slug: slugify(badgeNames.VENDOR_CERTIFIED),
+    assignedRole: BADGE_ACCESS_LEVEL.VENDOR,
+    icon: { name: 'default', color: '#000000' },
+  },
+  VENDOR_FIRST_SALE: {
+    name: badgeNames.VENDOR_FIRST_SALE,
+    slug: slugify(badgeNames.VENDOR_FIRST_SALE),
+    assignedRole: BADGE_ACCESS_LEVEL.VENDOR,
+    icon: { name: 'default', color: '#000000' },
+  },
+  VENDOR_VERIFIED: {
+    name: badgeNames.VENDOR_VERIFIED,
+    slug: slugify(badgeNames.VENDOR_VERIFIED),
+    assignedRole: BADGE_ACCESS_LEVEL.VENDOR,
+    icon: { name: 'default', color: '#000000' },
+  },
 };
 
 export default AUTOMATED_BADGES;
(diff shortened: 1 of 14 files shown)


server/helpers/automatedBadges.js
```diff
@@ -0,0 +1,9 @@
+const AUTOMATED_BADGES = {
```
andela-hpopoola, line 1: We need to be more explicit with the defaults defined here as we have done for `mailer/index.js`. We should define the name, slug, assignedRole, icon name (can be set to 'default'), and colour.

server/services/assignedBadge.service.js
```diff
@@ -138,3 +139,28 @@ export const getAllAssignedBadges = async (user, { page = 1, limit = 10, ...quer
...
+
+export const assignBadgeAutomatically = async (slug, userId) => {
+  let badge;
+
+  const badgeFromSlug = await getBadgeBySlug(slug);
+
+  [badge] = badgeFromSlug;
+
+  if (!badge) {
+    badge = await addBadge({ name: slug });
```
andela-hpopoola, line 151: We should add the whole badge object with a valid name, role and icon information -- not using a generated name from the slug.

server/services/assignedBadge.service.js
```diff
@@ -138,3 +139,28 @@ export const getAllAssignedBadges = async (user, { page = 1, limit = 10, ...quer
   const pagination = generatePagination(page, limit, total);
   return { pagination, result: assignedBadges[0].data };
 };
+
+export const assignBadgeAutomatically = async (slug, userId) => {
```
andela-hpopoola, line 143: I think it is better we pass the user object here to confirm that the user has the right role before we assign the badge to them.

server/services/transaction.service.js
```diff
@@ -26,6 +26,9 @@ import {
   getReferralByOfferId,
   updateReferralAccumulatedRewardAndRewardStatus,
 } from './referral.service';
+// eslint-disable-next-line import/no-cycle
```
andela-hpopoola, line 29: I don't think this is necessary as the transaction service is not used in assignedBadge, can you re-verify?

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to ensure that the implementation is done in the right way. Also, you would need to create a custom function test to make testing automatic badges easier.

server/services/assignedBadge.service.js
```diff
@@ -138,3 +139,28 @@ export const getAllAssignedBadges = async (user, { page = 1, limit = 10, ...quer
...
+  let badge;
```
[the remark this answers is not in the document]
tobieeniafe: `badge` can't be a constant here, because it's going to be reassigned to the generated badge if the slug does not exist.

server/services/assignedBadge.service.js
tobieeniafe: This check is already carried out in the `assignBadge` service

test/services/transaction.service.test.js
```diff
@@ -120,6 +121,10 @@ describe('Transaction Service', () => {
           expect(matchedNextPayments[0].expectedAmount).to.eql(500_000);
           expect(matchedNextPayments[0].expiresOn).to.eql(expiresOnForSecondCycle);
         });
+
+        context('when badge is assigned', () => {
```
andela-hpopoola, line 125:
```suggestion
        context('when user makes their first payment', () => {
```

test/services/transaction.service.test.js
```diff
@@ -165,6 +170,19 @@ describe('Transaction Service', () => {
...
 
+    context('when its the last payment', () => {
+      beforeEach(async () => {
+        fakeDate = sinon.useFakeTimers({
+          now: new Date('2020-06-8'),
+        });
+        await addTransaction({ ...transaction, amount: 4_000_000 });
+      });
+
+      context('when badge is assigned', () => {
```
andela-hpopoola, line 181:
```suggestion
      context("when its the vendor's first sale", () => {
```

test/services/user.service.test.js
```diff
@@ -397,12 +398,14 @@ describe('User Service', () => {
         expect(activatedUser).to.have.property('activationDate');
       });
 
-      context('when new notification is added', () => {
+      context('when new notification is added & badge is assigned', () => {
```
andela-hpopoola, line 401: Lets separate the tests please 🙏🏽 ? It is better for the badge test to be in a new test case.

test/services/user.service.test.js
```diff
@@ -959,6 +962,8 @@ describe('User Service', () => {
         });
 
         expectNewNotificationToBeAdded(NOTIFICATIONS.VERIFY_VENDOR, vendor._id);
+
+        expectBadgeToBeAssignedAutomatically(AUTOMATED_BADGES.VENDOR_VERIFIED, vendor._id);
```
andela-hpopoola, line 966: This is not supposed to be under the notifications context as it would make the tests difficult to understand in the future.

server/services/assignedBadge.service.js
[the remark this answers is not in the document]
andela-hpopoola: Your comment is well-noted and very correct. 

However, the point is not about the `const`, it is about not using the `badgeFromSlug` variable as it is not relevant in the code. 

If you need to use `let` instead of `const` , then this can be written as:

```suggestion
  let [badge] = await getBadgeBySlug(slug)
```

server/services/assignedBadge.service.js
```diff
@@ -138,3 +139,28 @@ export const getAllAssignedBadges = async (user, { page = 1, limit = 10, ...quer
...
+
+  const badgeFromSlug = await getBadgeBySlug(badgeDetails.slug);
+
+  [badge] = badgeFromSlug;
+
+  if (!badge) {
+    badge = await addBadge({ ...badgeDetails, automated: true });
+  }
+
+  if (!badge.automated) {
```
andela-hpopoola, line 154: Can we write tests to trigger this error 🙏🏽

andela-hpopoola (approved):
Thank you for making the updates. I left a few more comments on this, however, this should be good to go once you fix the comments. Thank you and nice work.

PR #207, opened 2021-07-14
#178794426 update offer model to include new terms and conditions

tobieeniafe:
#### What does this PR do?
Update offer model to include new terms and conditions

[Source document](https://docs.google.com/spreadsheets/d/1fcpFHb7_BLfRxkx5_szAq_LsvvvRTGVxJssYIv8VteI/edit?usp=sharing)

NA

#178794426


server/models/offer.model.js
```diff
@@ -175,6 +175,45 @@ const OfferSchema = new mongoose.Schema(
         },
       },
     ],
+    additionalClause: {
```
andela-hpopoola, line 178: I think this is a wrong implementation of the additional Clause stated in the offer letter template. I thought you suggested that we handle this in a later PR. Moreover, this should be an array of Strings 🤷🏽‍♂️

server/models/offer.model.js
```diff
@@ -175,6 +175,45 @@ const OfferSchema = new mongoose.Schema(
...
+      {
+        name: {
+          type: String,
+        },
+        amount: {
+          type: Number,
+        },
+      },
+    ],
+    paymentAccount: {
```
andela-hpopoola, line 191: I think creating the Bank Model should be done in a later PR. However, I think this should be an array of IDs. We are not displaying just one Bank Account, we are displaying all the bank accounts in the database at the time the offer was created.

Also lets update this to `bankAccounts` for easy reconcilation.

server/models/offer.model.js
```diff
@@ -175,6 +175,45 @@ const OfferSchema = new mongoose.Schema(
...
+      type: ObjectId,
+    },
+    otherTerms: {
```
andela-hpopoola, line 194: All of the properties here have a default, can we use them please? Also, can you follow up with Nnamdi on the section b of the Other Terms and Conditions. We can't proceed without him finalizing this as this determines `terminationInterest` and `terminationPeriod`. You would also need to get the

server/models/offer.model.js
```diff
@@ -175,6 +175,45 @@ const OfferSchema = new mongoose.Schema(
...
+      administrativeCharge: {
+        type: Number,
+      },
+      bankDraftDue: {
+        type: Date,
+      },
+      dateDue: {
+        type: Date,
```
andela-hpopoola, line 202: Not sure the rationale behind using the datatype of Date for values we wish to indicate in days. Since the output would be `2 days` or `5 working days`, then it is not different from `2 bedrooms`, `5 bedrooms`. 

I think the best datatype to indicate days (in the sheet) should be Number.

server/models/offer.model.js
```diff
@@ -175,6 +175,45 @@ const OfferSchema = new mongoose.Schema(
         },
       },
     ],
+    additionalClause: {
+      type: String,
+    },
```
andela-hpopoola, line 180: You didn't add `paymentPlans` as indicated in the sheets (on line 23). When we add `paymentPlans`, we need to review all the places where the `paymentPlans` are regenerate to use the one we have saved in the database.

server/models/offer.model.js
```diff
@@ -175,6 +175,45 @@ const OfferSchema = new mongoose.Schema(
...
+    otherPayments: [
```
andela-hpopoola, line 181: For OtherPayments, Nnamdi had already concluded on the six types of Other payments that we might have. Legal Fee, Agency Fee e.t.c. 

Lets use this values directly instead of making the values here open. This will make it easier to validate each fields and use different values for them based on this. Please note that some are in percentage and some are normal number.

Adding this would force us to add a new value to the model called `overallTotal` (or anyname similar), that contains the sum of the price of the property + the sum of everything we have placed in the OtherPayments.

andela-hpopoola (changes requested):
Thank you for creating the initial offer model update. I have a couple of comments:

1. For your PR description, you need to include the updated offer letter template and the sheets we used for this review. Those are our source of truth, including them in this PR will make it easier to locate them in the future if we need to confirm anything. Please take note of this when you are creating any similar PR in the future.
2. We need to create the bank model before proceeding with this PR
3. You need to reach out to Nnamdi to conclude the section b part of the terms and condition.


Thank you.

andela-hpopoola (changes requested):
Thank you for working on this, this looks good to me. 

However, we need to update or ensure that we can add and update offers with the new properties in the model. We should also add tests to ensure that they work seamlessly. I believe that could be done in this PR, so I am requesting for more changes to test the updated model.

andela-hpopoola (approved):

PR #209, opened 2021-07-16
#178877471 include role in projected user values

tobieeniafe:
#### What does this PR do?
Include role in projected user values

Make role projected when projecting user values

NA

#178877471


andela-hpopoola (approved):
Thank you for working on this. I would have suggested we write at least one test for this but would ignore it since it is not a code-breaking change even if it doesn't work as expected.

PR #210, opened 2021-07-19
#178926229 fix bugs for automated badges

tobieeniafe:
#### What does this PR do?
Fix bugs for automated badges

Fix bugs for automated badges

NA

#178926229


server/helpers/automatedBadges.js
```diff
@@ -2,18 +2,18 @@ import { slugify } from './funtions';
...
+  VENDOR_FIRST_SALE: 'First Sale Vendor',
+  VENDOR_VERIFIED: 'Registered Vendor',
 };
 
 const AUTOMATED_BADGES = {
   ACTIVATED_USER: {
     name: badgeNames.ACTIVATED_USER,
     slug: slugify(badgeNames.ACTIVATED_USER),
-    assignedRole: BADGE_ACCESS_LEVEL.ALL,
+    assignedRole: BADGE_ACCESS_LEVEL.USER,
```
andela-hpopoola, line 16: Not sure this is the expected way to update this. I think this would cause an error when adding other user roles like vendor and editor. I think we need to add a conditional statement to make this available for only normal users in the `activateUser` where this is used.

andela-hpopoola (changes requested):
Thank you for working on this. I left an important comment on ensuring that this doesn't throw an error when users are activating their account.

test/services/user.service.test.js
```diff
@@ -417,6 +417,31 @@ describe('User Service', () => {
         });
       });
 
+      context('when activated user is not a user', () => {
```
andela-hpopoola, line 420:
```suggestion
      context('when activated user is not a user role', () => {
```

test/services/user.service.test.js
```diff
@@ -417,6 +417,31 @@ describe('User Service', () => {
...
+        const editor = UserFactory.build(
+          { activated: false, role: USER_ROLE.EDITOR },
+          { generateId: true },
+        );
+
+        beforeEach(async () => {
+          editorToken = await addUser(editor);
+        });
+
+        context('expects user to be assigned', () => {
```
andela-hpopoola, line 432:
```suggestion
        context('when user is an editor', () => {
```

test/services/user.service.test.js
```diff
@@ -417,6 +417,31 @@ describe('User Service', () => {
...
+          beforeEach(async () => {
+            await activateUser(editorToken);
+          });
+
+          it('activates the user account', async () => {
+            const activatedUser = await activateUser(editorToken);
+            expect(activatedUser.activated).to.eql(true);
```
andela-hpopoola, line 439: We need to add a test here to confirm that the badge is not assigned to this user.

andela-hpopoola (approved):
Thank you for working on this. I left a few comment, feel free to merge after fixing.

PR #211, opened 2021-07-22
#178926255 add video to property

tobieeniafe:
#### What does this PR do?
Add video to property

Add video to property

`POST /api/v1/property/:id/videos`
`PUT /api/v1/property/:id/videos`
`DELETE /api/v1/property/:id/videos`

#178926255


server/models/property.model.js
```diff
@@ -174,6 +174,16 @@ const PropertySchema = new mongoose.Schema(
...
       },
     ],
+    videos: [
+      {
+        title: {
+          type: String,
+        },
+        url: {
+          type: String,
+        },
```
andela-hpopoola, line 184: I think we need to add `approved` (with other necessary information) here to know the videos that have been approved by the admin.

andela-hpopoola (changes requested):
Thank you for working on this. This looks good but I noticed a flaw we need to avoid. We need the admin to verify the videos that would be linked before displaying them to the users. Apologies for just realizing that 🤦🏽. 

The problem is that the current implementation will make it difficult to load newly uploaded videos for the admin to approve. If you think this can be achieved in a very easy way, let's continue with this implementation, however, if you think it would be difficult to achieve this, I would suggest we move the property video implementationinto its own model and link it to the property model. The current implementation can be easily converted to support a custom video model, so I don't really see it as a major waste of time.

Thank you.

server/schemas/video.schema.js
```diff
@@ -0,0 +1,20 @@
...
+} from './helper.schema';
+
+export const addVideoSchema = Joi.object().keys({
+  title: requiredString('Video Name'),
+  url: requiredString('Video URL'),
+  propertyId: requiredObjectId('Property id'),
+});
+
+export const updateVideoSchema = Joi.object({
+  videoId: requiredObjectId('Video id'),
```
andela-hpopoola, line 16: Let's make this one `id`

```suggestion
  id: requiredObjectId('Video id'),
```

server/services/video.service.js
```diff
@@ -0,0 +1,178 @@
...
+export const getAllVideos = async (user, propertyId, { page = 1, limit = 10, ...query } = {}) => {
+  const property = await getPropertyById(propertyId).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+
+  if (!property) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Invalid property');
+  }
+
+  if (user.role === USER_ROLE.VENDOR) {
```
andela-hpopoola, line 129: This can be merged into a single statement:

```suggestion
  if (user.role === USER_ROLE.VENDOR && user._id.toString() !== property.addedBy.toString()) {
```

server/services/video.service.js
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  try {
+    return Video.findByIdAndDelete(video._id);
+  } catch (error) {
+    throw new ErrorHandler(httpStatus.BAD_REQUEST, 'Error deleting video', error);
+  }
+};
+
+export const getAllVideos = async (user, propertyId, { page = 1, limit = 10, ...query } = {}) => {
```
andela-hpopoola, line 120: This is only needed for an admin. For users and vendors, the video should be loaded with the property (like Gallery).

For users, only approved videos should be loaded for a single property. For admin and vendors, all videos should be loaded.

server/services/video.service.js
```diff
@@ -0,0 +1,178 @@
...
+
+  const { filterQuery, sortQuery } = buildFilterAndSortQuery(VIDEO_FILTERS, query);
+
+  const videoOptions = [
+    { $match: { $and: filterQuery } },
+    { $sort: sortQuery },
+    { $match: { _id: ObjectId(propertyId) } },
+    {
+      $lookup: {
+        from: 'assignedbadges',
```
andela-hpopoola, line 143: I believe this is not caught in the tests because you have not written the tests yet. This is video not badges. Please fix in other places too.

server/models/property.model.js
```diff
@@ -174,16 +174,9 @@ const PropertySchema = new mongoose.Schema(
...
-      {
-        title: {
-          type: String,
-        },
-        url: {
-          type: String,
-        },
-      },
-    ],
+    videos: {
```
andela-hpopoola, line 177: Lets be careful about this and only use it if there are no alternative. It might make video maintenance difficult.

andela-hpopoola (changes requested):
Thank you for working on this. A couple of suggestions to improve

- The model should be called property-videos so that it doesn't disturb us if we want to add another video model in the future
- We should enforce that the videos are loaded when a single property is loaded (like gallery). 
- We need to write tests to ensure the code works as expected.

server/models/propertyVideo.model.js
```diff
@@ -0,0 +1,57 @@
...
+    title: {
+      type: String,
+      required: true,
+    },
+    url: {
+      type: String,
+      required: true,
+    },
+    propertyId: {
+      type: ObjectId,
```
andela-hpopoola, line 41: `required: true` is needed here.

server/helpers/filters.js
```diff
@@ -387,3 +387,9 @@ export const BANK_ACCOUNT_FILTERS = {
   bankName: { type: FILTER_TYPE.STRING },
   createdAt: { type: FILTER_TYPE.DATE },
 };
+
+export const PROPERTY_VIDEO_FILTERS = {
```
andela-hpopoola, line 391: Lets add `url` here to ensure that people can search videos via URL too.

server/models/propertyVideo.model.js
```diff
@@ -0,0 +1,57 @@
...
+ *           title: building overview
+ *           url: intro.mp4
+ *           propertyId: 5f31e1234a36150920130fda
+ *           approved: false
+ *           approvedBy: 5f31e1a5580ea36150920130
+ */
+
+const { ObjectId } = mongoose.Schema.Types;
+
+const PropertyVideoSchema = new mongoose.Schema(
```
andela-hpopoola, line 30: Lets arrange the schema in alphabetical order. As a standard now, we would ensure that our schemas are arranged in alphabetical order. I had issues searching through offer letters for a field since the schema wasn't arranged in alphabetical order.

server/routes/propertyVideo.routes.js
```diff
@@ -0,0 +1,165 @@
...
+ *            $ref: '#/components/schemas/PropertyVideo'
+ *      description: Get all videos for a property
+ *     responses:
+ *      '200':
+ *        description: returns videos
+ *      '500':
+ *       description: Internal server error
+ */
+router.get(
+  '/:id/all',
```
andela-hpopoola, line 158:
```suggestion
  '/:propertyId/all',
```

In this scenario, we need to use the `propertyId` instead of `id` to differentiate that the id in question is not the video id but the property Id

server/services/propertyVideo.service.js
```diff
@@ -0,0 +1,178 @@
...
+};
+
+export const updateVideo = async (updatedVideo) => {
+  const video = await getVideoById(updatedVideo.id).catch((error) => {
+    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
+  });
+  if (!video) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Video not found');
+  }
+
```
andela-hpopoola, line 54: Since we are trying to make the video title to be unique, then we should also check that the updated title doesn't exist in the database (as verified when adding a new video). 

You might need to do this check if the video title is changed i.e  current video title is not equal to the old video title.

andela-hpopoola (approved):
Thank you for working on this. As discussed, I have taken a look through it and let a few comments to fix before merging. Thank you for the great work.

server/routes/propertyVideo.routes.js
tobieeniafe: Using `:propertyId` will prevent Joi from validating the id. The description will be stated in the description.

PR #212, opened 2021-07-27
#179002965 fix offer model bugs

tobieeniafe:
#### What does this PR do?
Fix offer model bugs

Fix offer model bugs
[Source document](https://docs.google.com/spreadsheets/d/1fcpFHb7_BLfRxkx5_szAq_LsvvvRTGVxJssYIv8VteI/edit?ts=60ee8918#gid=0)

`POST /api/v1/offer/create`

#179002965


server/schemas/offer.schema.js
```diff
@@ -26,13 +27,13 @@ const otherPayments = Joi.object().keys({
 });
```
andela-hpopoola, line 27: AgencyFee and LegalFee are in percentage. Can we ensure that they range between 0 - 100 too? It might be better to create an `optionalPercentage` function (if it doesn't exist) to deal with cases like this in the future.

test/controllers/offer.controller.test.js
```diff
@@ -316,6 +316,137 @@ describe('Offer Controller', () => {
...
+                  `"Termination Interest" must be less than or equal to 10`,
+                );
+                done();
+              });
+          });
+        });
+
+        context('when termination period is outside range', () => {
+          it('returns an error', (done) => {
+            const offer = OfferFactory.build({ otherTerms: { terminationPeriod: 211 } });
```
andela-hpopoola, line 434:
```suggestion
            const offer = OfferFactory.build({ otherTerms: { terminationPeriod: 181 
```

Lets use 181 here. I am afraid that if we mistakenly update this to a larger number in the future, something like `200`, the test would still pass which is not the expected behaviour.

test/controllers/offer.controller.test.js
```diff
@@ -316,6 +316,137 @@ describe('Offer Controller', () => {
               });
           });
         });
+
+        context('when administrative charge is outside range', () => {
```
andela-hpopoola, line 320: Can we add tests for cases when the fields are below the minimum? In this case, we should test for '-1' or whatever value that is minimum.

andela-hpopoola (approved):
Thank you for working on this. I left a few comments on testing the minimum values too and not allowing the maximum value used in test to exceed +1. Also, we need to add validation for agencyFee and legalFee since they are in percentages. 

Feel free to merge after making this changes. Thank you.

PR #228, opened 2021-08-16
#179232133 create endpoint to disapprove videos 

tobieeniafe:
#### What does this PR do?
Create endpoint to disapprove videos 

-Create endpoint to disapprove videos 
-Add `status` to propertyVideo model

`PUT /api/v1/property-video/:id/disapprove`

#179232133 #179232174


server/helpers/constants.js
```diff
@@ -134,3 +134,9 @@ export const PAYMENT_BREAKDOWN_OPTIONS = {
   FINAL_DEPOSIT: 'Final Deposit',
   INITIAL_DEPOSIT: 'Initial Deposit',
 };
+
+export const PROPERTY_VIDEO_STATUS = {
+  APPROVED: 'Approved',
+  PENDING_ADMIN_REVIEW: 'Pending Admin Review',
+  PENDING_VENDOR_RESPONSE: 'Pending Vendor Response',
```
andela-hpopoola, line 141: Not sure why you used this when the vendor is not responding to the video status. Lets use `Disapproved`

server/helpers/filters.js
```diff
@@ -394,6 +394,7 @@ export const PROPERTY_VIDEO_FILTERS = {
   approved: { type: FILTER_TYPE.BOOLEAN },
   createdAt: { type: FILTER_TYPE.DATE },
   propertyId: { type: FILTER_TYPE.OBJECT_ID },
+  status: { type: FILTER_TYPE.STRING },
```
andela-hpopoola, line 397: Since we are using status, approved is now redundant (approved is now part of status so it makes no sense retaining it). Can we remove it to prevent redundant updating?

server/models/propertyVideo.model.js
```diff
@@ -40,6 +41,13 @@ const PropertyVideoSchema = new mongoose.Schema(
       type: ObjectId,
       required: true,
     },
+    reason: {
```
andela-hpopoola, line 44: Lets use comment here instead of reason. 

On a better thought, Lets make this an array of String to enable us to retain previous admin comments. This means that when disapproving a video, an admin can add more comments without losing the previous ones.

server/services/propertyVideo.service.js
```diff
@@ -60,6 +61,14 @@ export const updateVideo = async (updatedVideo) => {
     throw new ErrorHandler(httpStatus.FORBIDDEN, 'You are not permitted to perform this action');
   }
 
+  let { status } = video;
```
andela-hpopoola, line 64: Even though this will no longer be valid as we are removing approved, you could have written the lines together as one.

```suggestion
  let { approved, status } = video;
```

test/controllers/propertyVideo.controller.test.js
```diff
@@ -570,4 +606,92 @@ describe('Property Video Controller', () => {
...
+      method,
+      user: adminUser,
+      userId: adminUser._id,
+      data,
+      useExistingUser: true,
+    });
+
+    itReturnsForbiddenForNoToken({ endpoint, method, data });
+
+    context('when disapproval service returns an error', () => {
```
andela-hpopoola, line 681:
```suggestion
    context('when disapproveVideo service returns an error', () => {
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to ensure that this is properly implemented. Thank you.

test/controllers/propertyVideo.controller.test.js
```diff
@@ -552,7 +548,6 @@ describe('Property Video Controller', () => {
 
       context('when multiple filters are used', () => {
         const multipleDetails = {
-          approved: property2Video.approved,
```
andela-hpopoola, line 519: Lets add status here to confirm that we can filter by status too.

test/controllers/propertyVideo.controller.test.js
```diff
@@ -573,7 +568,6 @@ describe('Property Video Controller', () => {
                 totalPage: 1,
               });
               expect(res.body.result[0]._id).to.be.eql(property2Video._id.toString());
-              expect(res.body.result[0].approved).to.be.eql(multipleDetails.approved);
```
andela-hpopoola, line 540: Lets replace this with `status`

test/controllers/property.controller.test.js
```diff
@@ -1262,9 +1262,11 @@ describe('Property Controller', () => {
               expectResponseToExcludeSensitiveUserData(res.body.property.vendorInfo);
               expectResponseToContainNecessaryVendorData(res.body.property.vendorInfo);
               expect(res.body.property.videos.length).to.be.eql(3);
-              expect(res.body.property.videos[0].approved).to.be.eql(true);
-              expect(res.body.property.videos[1].approved).to.be.eql(true);
-              expect(res.body.property.videos[2].approved).to.be.eql(false);
+              expect(res.body.property.videos[0].status).to.be.eql(PROPERTY_VIDEO_STATUS.APPROVED);
+              expect(res.body.property.videos[1].status).to.be.eql(PROPERTY_VIDEO_STATUS.APPROVED);
```
andela-hpopoola, line 1266: It would be better to change this to PENDING_ADMIN_APPROVAL  to confirm that all the statuses are loaded.

Also, we need to add an assertion to confirm that the comments are loaded here. A single liner test should work in this case to confirm that video's comment is loaded here too.

andela-hpopoola (approved):
Thank you for working on this. I left a few comments but it should be ready to merge after fixing.

PR #230, opened 2021-08-16
#179084246 add ratings and review for property

tobieeniafe:
#### What does this PR do?
Add ratings and review for a property

Add ratings and review for a property

`POST /api/v1/property-rating/`

#179084246


server/services/propertyRating.service.js
```diff
@@ -0,0 +1,29 @@
...
+
+  if (offer.status !== OFFER_STATUS.RESOLVED) {
+    throw new ErrorHandler(
+      httpStatus.PRECONDITION_FAILED,
+      'You cannot rate a property you do not own',
+    );
+  }
+
+  try {
+    const addedArea = await new PropertyRating(ratingInfo).save();
```
andela-hpopoola, line 24: Not sure why you used addedArea here 🤷🏽‍♂️ 

```suggestion
    const addedArea = await new PropertyRating(ratingInfo).save();
```

server/models/propertyRating.model.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+const { ObjectId } = mongoose.Schema.Types;
+const PropertyRatingSchema = new mongoose.Schema(
+  {
+    offerId: {
+      type: ObjectId,
+      required: true,
+    },
+    propertyId: {
```
andela-hpopoola, line 33: Can you confirm if ratings should only be available to users that have an offer for a property? On most sites, ratings is available to everyone but something like 'Verified Purchase' is added near the entry of the people that have bought that property.

server/models/propertyRating.model.js
```diff
@@ -0,0 +1,65 @@
...
+      valueForPrice: {
+        type: Number,
+        default: 5,
+      },
+    },
+  },
+  { timestamps: true },
+);
+
+const PropertyRating = mongoose.model('PropertyRating', PropertyRatingSchema);
```
andela-hpopoola, line 63: Lets use PropertyReview instead of Rating to make it more explicit.

server/models/propertyRating.model.js
```diff
@@ -0,0 +1,65 @@
...
+      required: true,
+    },
+    review: {
+      type: String,
+      required: true,
+    },
+    rating: {
+      condition: {
+        type: Number,
+        default: 5,
```
andela-hpopoola, line 48: Since this would be used in calculating the average, we might need to confirm if we should omit this or use a default of 5.

server/models/propertyRating.model.js
```diff
@@ -0,0 +1,65 @@
...
+      },
+      location: {
+        type: Number,
+        default: 5,
+      },
+      valueForPrice: {
```
andela-hpopoola, line 54: Please note that there is an invisible 4th entry (Average Ratings) that will not be included in the database but should be returned when we are loading each property. That is the average of the 3 fields you have used here.

andela-hpopoola (changes requested):
Thank you for working on this. This is a good start; I left a few comments to improve the current implementation. Thank you.

PR #233, opened 2021-08-19
#179280518 fix next payment reminder

tobieeniafe:
#### What does this PR do?
Fix next payment reminder

Fix next payment reminder

`GET /api/v1/next-payment/reminder`

#179280518


server/services/nextPayment.service.js
```diff
@@ -193,15 +193,30 @@ export const getUnresolvedNextPayments = async (user, { page = 1, limit = 10 } =
 };
 
 export const sendReminder = async () => {
-  const today = new Date();
+  const today = endOfDay(new Date());
```
andela-hpopoola, line 196: Any reason why we are using `endOfDay` here? With the current implementation, if `endOfDay` should work, then using `new Date()` should work too.

server/services/nextPayment.service.js
```diff
@@ -193,15 +193,30 @@ export const getUnresolvedNextPayments = async (user, { page = 1, limit = 10 } =
...
   const reminders = await NextPayment.aggregate([
     { $match: { resolved: false } },
     {
       $match: {
         $or: [
-          { expiresOn: { $eq: new Date(format(add(today, { days: 1 }), 'yyyy-MM-dd')) } },
-          { expiresOn: { $eq: new Date(format(add(today, { days: 7 }), 'yyyy-MM-dd')) } },
-          { expiresOn: { $eq: new Date(format(add(today, { days: 30 }), 'yyyy-MM-dd')) } },
+          {
+            expiresOn: {
```
andela-hpopoola, line 203: Can we add the days to send reminder in an array and use it to generate the object here to make updating this in the future easier. 

Something like this makes more sense (not tested though)

```js
 ...
  const reminderDays = [1, 7, 30];
  const validReminderQuery = [];

  reminderDays.forEach((day) =>
    validReminderQuery.push({
      expiresOn: {
        $gt: add(today, { days: day - 1 }),
        $lte: add(today, { days: day }),
      },
    }),
  );
  const reminders = await NextPayment.aggregate([
    { $match: { resolved: false } },
    {
      $match: {
        $or: validReminderQuery,
      },
    },
...

andela-hpopoola (approved):
Thank you for working on this. I left a few suggestions otherwise, this looks good to me.

server/services/nextPayment.service.js
tobieeniafe: Using `new Date()` causes inconsistencies. For instance, if the endpoint is called by 11am on `2021-08-10`, Nextpayments for `2021-08-10T20:30:00.000+00:00` and `2021-08-11T00:00:00.000+00:00` will be be valid as they are both greater than the date `2021-08-10T11:00:00.000+00:00`. Hence the `endofDay` to avoid such inconsistency.

PR #234, opened 2021-08-22
#179311638 add property testimonials

tobieeniafe:
#### What does this PR do?
Add routes to add hide and reveal testimonials

Add routes to add hide and reveal testimonials

`PUT /api/v1/property/testimonial`
`PUT /api/v1/property/testimonial/hide`
`PUT /api/v1/property/testimonial/show`

#179311638


server/models/property.model.js
```diff
@@ -231,6 +231,26 @@ const PropertySchema = new mongoose.Schema(
...
+          type: Boolean,
+          default: false,
+        },
+        offerId: {
+          type: ObjectId,
+        },
+        testimonial: {
+          type: String,
+        },
+        userId: {
```
andela-hpopoola, line 249: When we are loading the testimonial, we would need to load the user information so that we can be sure of the person making the testimonial. This might be a good reason to create a different model for testimonials if it would be difficult to implement.

server/models/property.model.js
```diff
@@ -231,6 +231,26 @@ const PropertySchema = new mongoose.Schema(
         default: false,
       },
     },
+    testimonials: [
+      {
+        dateAdded: {
+          type: Date,
+        },
+        hidden: {
```
andela-hpopoola, line 239: This is no longer needed as discussed in today's spring planning.

server/services/property.service.js
```diff
@@ -1004,3 +1011,107 @@ export const requestToUnflagProperty = async (propertyInfo) => {
     );
   }
 };
+
+export const addTestimonial = async (testimonialInfo) => {
+  const offer = await getOfferById(testimonialInfo.offerId).catch((error) => {
```
andela-hpopoola, line 1016: I think we should just open up testimonial to everyone that has a valid offer to that property even if they have not paid.

andela-hpopoola (changes requested):
Thank you for working on this. Based on the conversation we heard this morning, I think it might be safer to move testimonials to their own model since we are introducing a vendor's comment. This will make it easier to load testimonials with vendors' comments for a single user. This will also make it easier for a vendor to load testimonials on a single page and add comments if possible.

Also, if we are using a new model, can we retain the `/property/testimonial` endpoint to cater for this 🙏🏽 .

Thank you.

tobieeniafe:
> Thank you for working on this. Based on the conversation we heard this morning, I think it might be safer to move testimonials to their own model since we are introducing a vendor's comment. This will make it easier to load testimonials with vendors' comments for a single user. This will also make it easier for a vendor to load testimonials on a single page and add comments if possible.
> 
> Also, if we are using a new model, can we retain the `/property/testimonial` endpoint to cater for this 🙏🏽 .
> 
> Thank you.
🙃

server/services/property.service.js
```diff
@@ -1004,3 +1011,88 @@ export const requestToUnflagProperty = async (propertyInfo) => {
...
+
+  if (!offer) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Invalid offer');
+  }
+
+  if (offer.userId.toString() !== testimonialInfo.userId.toString()) {
+    throw new ErrorHandler(httpStatus.FORBIDDEN, 'You are not permitted to perform this action');
+  }
+
+  if (offer.status !== OFFER_STATUS.RESOLVED) {
```
andela-hpopoola, line 1028: > I think we should just open up testimonial to everyone that has a valid offer to that property even if they have not paid.

Based on this comment, I don't think it is wise to wait till they have finished paying before we allow them to make a testimonial. At least, they should be able to make testimonial if they have made their initial payment. 

Can you follow up with Nnamdi on this to reduce back and forth?

server/models/property.model.js
andela-hpopoola: This is very important, can you please make it work?

server/services/property.service.js
```diff
@@ -1004,3 +1011,88 @@ export const requestToUnflagProperty = async (propertyInfo) => {
...
+  });
+
+  if (existingTestimonial.length > 0) {
+    throw new ErrorHandler(
+      httpStatus.PRECONDITION_FAILED,
+      'You already have a testimonial for this property',
+    );
+  }
+
+  const property = await getPropertyById(offer.propertyId).catch((error) => {
```
andela-hpopoola, line 1046: Instead of getting the property before adding testimonials, I am not sure why you are not using mongoose push directly.

https://docs.mongodb.com/manual/reference/operator/update/push/

andela-hpopoola (changes requested):
Thank you for the updates. As said in my previous response, we would need the following endpoints to simply responding to testimonials and following up testimonials:

`Get All` Testimonials Endpoint: This should be a paginated list of all testimonials in the database.
      - Admin gets all the testimonials in the database so that they can review
      - Vendors gets all the testimonials for all of their property so that they can review and respond
      - User gets all the testimonials they have made for review

Thank you.

server/routes/index.js
```diff
@@ -47,4 +48,5 @@ export default (app) => {
   app.use('/api/v1/assign-badge', assignedBadge);
   app.use('/api/v1/bank-account', bankAccount);
   app.use('/api/v1/property-video', propertyVideo);
+  app.use('/api/v1/property/testimonial', propertyTestimonial);
```
andela-hpopoola, line 51: I think this should come before property so that it is not colliding with the `property/:id` endpoint. 

Note: In the future, we would also need to fix the endpoints for other offending endpoints.

server/services/propertyTestimonial.service.js
```diff
@@ -0,0 +1,88 @@
...
+
+export const addTestimonial = async (testimonialInfo) => {
+  const offer = await getOfferById(testimonialInfo.offerId);
+
+  if (!offer) {
+    throw new ErrorHandler(httpStatus.NOT_FOUND, 'Invalid offer');
+  }
+
+  if (
+    offer.status !== OFFER_STATUS.INTERESTED &&
```
andela-hpopoola, line 23: Can we put this in a function to avoid repetition when we need to use valid offers?

server/services/propertyTestimonial.service.js
```diff
@@ -0,0 +1,88 @@
...
+    offer.status !== OFFER_STATUS.ALLOCATED &&
+    offer.status !== OFFER_STATUS.RESOLVED
+  ) {
+    throw new ErrorHandler(
+      httpStatus.PRECONDITION_FAILED,
+      'You cannot review a property you do not own',
+    );
+  }
+
+  if (offer.userId.toString() !== testimonialInfo.userId.toString()) {
```
andela-hpopoola, line 34: We can combine this with the if statement above since they have the same error message.

server/services/propertyTestimonial.service.js
```diff
@@ -0,0 +1,88 @@
...
+    );
+  }
+
+  const testimonial = {
+    ...testimonialInfo,
+    offerId: offer._id,
+    propertyId: offer.propertyId,
+    userId: testimonialInfo.userId,
+    vendorId: offer.vendorId,
+    testimonial: testimonialInfo.testimonial,
```
andela-hpopoola, line 56: I think this can be removed since it is already part of `...testimonialInfo` on line 51.

andela-hpopoola (changes requested):
Thank you for working on this. This looks much better. Can you fix the comments I left on this PR and add the `\all` testimonials endpoint for all roles. Thank you.

PR #235, opened 2021-08-24
#179316728 allow users to disable pagination limit

tobieeniafe:
#### What does this PR do?
Allow users to disable pagination limit

Allow users to disable pagination limit

All '/all' endpoints 

#179316728


server/helpers/pagination.js
```diff
@@ -1,11 +1,11 @@
 export const generatePagination = (page, limit, total) => {
   const currentPage = parseInt(page, 10) < 1 ? 1 : parseInt(page, 10);
   return {
-    currentPage,
+    currentPage: limit === 0 ? 1 : currentPage,
```
andela-hpopoola, line 4: I think it is better to use a variable for this instead of the expression.

```suggestion
    currentPage:  removeLimit ? 1 : currentPage,
```

and somewhere above, that variable is defined as

```js
   const removeLimit = limit === 0;
```

server/helpers/pagination.js
```diff
@@ -1,11 +1,11 @@
...
     limit: parseInt(limit, 10),
     offset: (currentPage - 1) * limit,
     total,
-    totalPage: Math.ceil(total / limit) || 0,
+    totalPage: limit === 0 ? 1 : Math.ceil(total / limit) || 0,
```
andela-hpopoola, line 8: Same here:

```suggestion
    totalPage: removeLimit ? 1 : Math.ceil(total / limit) || 0,
```

Also, we might want to consider the value of `totalPage` if limit is 0 and total is 0.

server/helpers/pagination.js
```diff
@@ -14,5 +14,9 @@ export const generateFacetData = (page, limit) => [
   { $limit: parseInt(limit, 10) },
 ];
 
-export const getPaginationTotal = (item) =>
-  item[0].metadata.length > 0 ? item[0].metadata[0].total : 0;
+export const getPaginationTotal = (item) => {
```
andela-hpopoola, line 17: I believe this can be rewritten as a single line in two ways:

1. Combining all the variables together

```js
export const getPaginationTotal = (item) => item[0]?.metadata && item[0].metadata.length > 0 ? item[0].metadata[0].total : item?.length || 0;
```

2. Using only the relevant variables

```js
export const getPaginationTotal = (item) => item?.[0]?.metadata?.[0]?.total || item?.length || 0;
```

Feel free to use whichever works for you.

server/services/area.service.js
```diff
@@ -154,11 +154,15 @@ export const getAllAreas = async ({ page = 1, limit = 10, ...query } = {}) => {
...
+  if (limit === 0) {
+    areaOptions.pop();
+  }
+
   const areas = await Area.aggregate(areaOptions);
 
   const total = getPaginationTotal(areas);
   const pagination = generatePagination(page, limit, total);
-  return { pagination, result: areas[0].data };
+  return { pagination, result: areas[0].data || areas };
```
andela-hpopoola, line 165: Lets define result as a variable here as done for pagination:

```js
const result = areas?.[0].data || areas
```

```suggestion
  return { pagination, result };
```

andela-hpopoola (changes requested):
Thank you for working on this. This looks good to me. I am not sure if all of the tests would pass, so if the tests are failing, let's fix for other models too since the fix is relatively easy.

andela-hpopoola (approved):
Thank you for working on this. Feel free to merge once you have fixed broken tests.

PR #238, opened 2021-08-29
#179311500 add sms service to next payment

tobieeniafe:
#### What does this PR do?
Add SMS service to next payment

Add SMS service to next payment

`GET /api/v1/next-payment/reminder`

#179311500


server/helpers/sms.js
```diff
@@ -0,0 +1,16 @@
...
+
+const KUDI_API_URL = `https://account.kudisms.net/api/?username=${KUDI_SMS_USERNAME}&password=${KUDI_SMS_PASSWORD}&`;
+
+export const sendSms = async ({ number, message }) => {
+  const messageEndpoint = `${KUDI_API_URL}message=${message}&sender=${KUDI_SMS_SENDERNAME}&mobiles=${number}`;
+
+  await axios.post(messageEndpoint);
+};
+
+export const MESSAGES = {
```
andela-hpopoola, line 13:
```suggestion
export const SMS_MESSAGES = {
```

server/helpers/sms.js
```diff
@@ -0,0 +1,16 @@
+import axios from 'axios';
+
+const { KUDI_SMS_USERNAME, KUDI_SMS_PASSWORD, KUDI_SMS_SENDERNAME } = process.env;
+
+const KUDI_API_URL = `https://account.kudisms.net/api/?username=${KUDI_SMS_USERNAME}&password=${KUDI_SMS_PASSWORD}&`;
```
andela-hpopoola, line 5:
```suggestion
const KUDI_API_URL = 'https://account.kudisms.net/api/';
```

server/helpers/sms.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const sendSms = async ({ number, message }) => {
+  const messageEndpoint = `${KUDI_API_URL}message=${message}&sender=${KUDI_SMS_SENDERNAME}&mobiles=${number}`;
+
+  await axios.post(messageEndpoint);
```
andela-hpopoola, line 10: I think for axios we should be able to specify everything as query params like this 


```js
axios.post(
        KUDI_API_URL,
        {},
        {
          params: {
            message,
            password: KUDI_SMS_PASSWORD,
            username: KUDI_SMS_USERNAME,
            ... // add the remaining parameters here
          }
        }
      )

```

You can check axios documentation for more info:

https://github.com/axios/axios

server/helpers/sms.js
```diff
@@ -0,0 +1,16 @@
+import axios from 'axios';
+
+const { KUDI_SMS_USERNAME, KUDI_SMS_PASSWORD, KUDI_SMS_SENDERNAME } = process.env;
```
andela-hpopoola, line 3: I think we should pursue using the same sendername everywhere. If it is different in the future, we can manually add it.
```suggestion
const { KUDI_SMS_USERNAME, KUDI_SMS_PASSWORD, SMS_SENDERNAME } = process.env;
```

.env.sample
```diff
@@ -12,3 +12,6 @@ SENDGRID_API_KEY=
 USER_SECRET=
 ENVIRONMENT=
 SLACK_TOKEN=
+KUDI_SMS_USERNAME=
+KUDI_SMS_PASSWORD=
+KUDI_SMS_SENDERNAME
```
andela-hpopoola, line 17: Same as above.

server/helpers/sms.js
```diff
@@ -0,0 +1,16 @@
...
+
+const KUDI_API_URL = `https://account.kudisms.net/api/?username=${KUDI_SMS_USERNAME}&password=${KUDI_SMS_PASSWORD}&`;
+
+export const sendSms = async ({ number, message }) => {
+  const messageEndpoint = `${KUDI_API_URL}message=${message}&sender=${KUDI_SMS_SENDERNAME}&mobiles=${number}`;
```
andela-hpopoola, line 8: For phone numbers, can we check if it is a valid phone number and convert '07' and '08' prefix  to 234 when the length of the phone is 13. SMS will only be sent if it starts with 234 and not with '07' and '08'. I think I noticed that the last time I used their platform.

This should also be used to confirm that the number is a valid Nigerian number. If the number is not a valid Nigerian number, then there is no point in sending the SMS.

server/helpers/sms.js
```diff
@@ -0,0 +1,16 @@
+import axios from 'axios';
+
+const { KUDI_SMS_USERNAME, KUDI_SMS_PASSWORD, KUDI_SMS_SENDERNAME } = process.env;
+
+const KUDI_API_URL = `https://account.kudisms.net/api/?username=${KUDI_SMS_USERNAME}&password=${KUDI_SMS_PASSWORD}&`;
+
+export const sendSms = async ({ number, message }) => {
```
andela-hpopoola, line 7: Lets use phoneNumber to make it easily understandable.

```suggestion
export const sendSms = async ({ phoneNumber, message }) => {
```

server/services/nextPayment.service.js
```diff
@@ -241,6 +242,14 @@ export const sendReminder = async () => {
       await createNotification(NOTIFICATIONS.PAYMENT_REMINDER, reminder.userId, {
         actionId: reminder.offerId,
       });
+
+      const message = `This is a quick reminder that the periodic payment on your property ${
+        reminder.propertyInfo.name
+      }, is due on ${convertDateToShortHumanFormat(reminder.expiresOn)}.`;
```
andela-hpopoola, line 248: Not sure why you made this error again. I thought we agreed that expiresOn will only be used for internal purposes (for Cron Jobs). To make this consistent with what Nnamdi told us and the Figma file, in cases where the user has not paid, then the due date should be used.

We might want to update the SMS message to reflect this too for past dates.

```suggestion
      }, is due on ${convertDateToShortHumanFormat(reminder.dueDate)}.`;
```

server/services/nextPayment.service.js
```diff
@@ -241,6 +242,14 @@ export const sendReminder = async () => {
...
+
+      const phoneNumber = reminder.userInfo.phone.replace('+', '');
```
andela-hpopoola, line 250: The phoneNumber should be properly sanitized as explained above with a different function.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on ensuring that this is properly implemented with fewer bugs. Thank you.

server/helpers/funtions.js
```diff
@@ -16,3 +16,20 @@ export const slugify = (text) =>
     .replace(/\s+/g, '-')
     .replace(/[^\w-]+/g, '')
     .replace(/--+/g, '-');
+
+export const verifyNigerianPhoneNumber = (number) =>
```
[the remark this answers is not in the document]
andela-hpopoola: To ensure that this works as expected, we could add some services test to confirm that the function works as expected by testing for the following numbers.

1. 2348012345678 - valid number
2. 2347012345678 - valid number
3. 2348112345678 - valid number
4. 2349012345678 - valid number
5.  08012345678 - valid number
6. 07012345678 - valid number
7. 08112345678 - valid number
8. 09012345678 - valid number

Then add a couple of invalid numbers.
1. 23480123456780 - invalid number
2. 2346012345678 - invalid number
3. ... add more invalid Nigeria number variant

This is necessary as it will improve our confidence if we need to update this method in the future. Tests like this are very very easy to write since you just need to call the function in the test and confirm that it is returning true or false.

The best way to go about this is to put all the valid numbers in an array and map it through a valid test and the invalid values in another array and map it too.

server/helpers/funtions.js
```diff
@@ -16,3 +16,20 @@ export const slugify = (text) =>
...
+  /^([0]{1}|\+?234)([7-9]{1})([0|1]{1})([\d]{1})([\d]{7})$/g.test(number);
+
+export const sanitizePhoneNumber = (number) => {
```
andela-hpopoola, line 23: Same comment as above but we would just add various variants of valid Nigerian phone numbers and test that they return the right sanitized number. We should test for space, dash, and plus as specified in the function.

server/helpers/sms.js
```diff
@@ -0,0 +1,35 @@
...
+
+const NODE_ENV = process.env.NODE_ENV || 'development';
+
+const KUDI_API_URL = 'https://account.kudisms.net/api/';
+
+export const sendSms = async ({ phoneNumber, message }) => {
+  if (NODE_ENV === 'production') {
+    const sanitizedNumber = sanitizePhoneNumber(phoneNumber);
+
+    if (verifyNigerianPhoneNumber(sanitizedNumber)) {
```
andela-hpopoola, line 14: If a user has a valid second phone number in our database and the first phone number is invalid for some reasons, then the best thing is to send to the valid second phone number. Can you update the implementation to cater for this 🙏🏽

server/helpers/funtions.js
```diff
@@ -16,3 +16,20 @@ export const slugify = (text) =>
...
+  const countryCode = 234;
+  let cleanedNumber = number.trim().replace('+', '').replace(/\s/g, '').replace(/-/g, '');
+
+  if (cleanedNumber.startsWith('0')) {
+    cleanedNumber = cleanedNumber.substring(1);
+  }
+  if (cleanedNumber.startsWith(countryCode)) {
```
andela-hpopoola, line 30:
```js
   return cleanedNumber.startsWith(countryCode) ? cleanedNumber : `${countryCode}${cleanedNumber}`;
```

server/helpers/funtions.js
andela-hpopoola: With this implementation, I think it is better to add the `countryCode` to the function and make it default to Nigeria Country code. That way, the function would easily be reusable for all countries.

```suggestion
export const sanitizePhoneNumber = (number, countryCode = 234) => {
```

andela-hpopoola (changes requested):
Thank you for working on this. I am sorry, I left more comments to ensure that we get this totally right. 

Also, those tests are extremely important to reduce bugs when we are updating those functions in the future (and of course, they are very easy to write, else we would have avoided them). 

This should be ready to merge after you fixed the comments. Thank you.

test/functions.test.js
```diff
@@ -0,0 +1,66 @@
...
+      '2346012345678',
+      '02012345667',
+      '080123456780',
+      '08312345678',
+      '09442345678',
+    ];
+
+    context('when phone number is valid', () => {
+      validNumbers.map((number) =>
+        it('returns true', async () => {
```
andela-hpopoola, line 28: Async and await is not required here since we are not making an API call. Please fix for below test too.

```js
    context('when phone number is valid', () => {
      validNumbers.map((number) =>
        it('returns true', () => {
          expect(isValidNigerianPhoneNumber(number)).to.eql(true);
        }),
      );
    });

```

test/functions.test.js
```diff
@@ -0,0 +1,66 @@
...
+      '+23408012345678',
+      '2348012345678',
+      '23408012345678',
+      '08012345678',
+      '0801-234-5678',
+      '0801 234 5678',
+      '+0801  2345678',
+    ];
+
+    context('when phone number is valid', () => {
```
andela-hpopoola, line 57:
```suggestion
    context('when a phone number is given', () => {
```

server/helpers/sms.js
```diff
@@ -0,0 +1,42 @@
+import axios from 'axios';
+import { sanitizePhoneNumber, isValidNigerianPhoneNumber } from './funtions';
+
+const { KUDI_SMS_USERNAME, KUDI_SMS_PASSWORD, SMS_SENDERNAME } = process.env;
+
+const NODE_ENV = process.env.NODE_ENV || 'development';
+
+const KUDI_API_URL = 'https://account.kudisms.net/api/';
+
+export const sendSms = async ({ phoneNumbers, message }) => {
```
andela-hpopoola, line 10: Since we are not saving phoneNumbers as an array in the database and we are only dealing with a maximum of 2 phone numbers, I think it will be advisable to pass it as directly instead of converting it to an array.

Converting to an array will mean that we would have to convert phoneNumbers to an array everywhere we want to use them which is not a great idea.

Please note that this function will be updated in the future to accommodate sending SMS to foreign numbers and we need to make it easy to understand while maintaining best practices. 

```suggestion
export const sendSms = async ({ message, phoneNumber1, phoneNumber2 }) => {
```

server/services/nextPayment.service.js
```diff
@@ -241,6 +242,18 @@ export const sendReminder = async () => {
...
+        reminder.propertyInfo.name
+      }, is due on ${convertDateToShortHumanFormat(reminder.dueDate)}.`;
+
+      const phoneNumbers = [reminder.userInfo.phone];
+
+      if (reminder.userInfo.phone2) {
+        phoneNumbers.push(reminder.userInfo.phone2);
+      }
+
+      await sendSms({ phoneNumbers, message });
```
andela-hpopoola, line 256:
```suggestion
      await sendSms({ message, phoneNumber1: reminder.userInfo.phone, phoneNumber2: reminder.userInfo.phone2  });
```

server/helpers/sms.js
```diff
@@ -0,0 +1,42 @@
+import axios from 'axios';
+import { sanitizePhoneNumber, isValidNigerianPhoneNumber } from './funtions';
```
andela-hpopoola, line 2: Can we rename the `functions.js` file. It is a little confusing...

```suggestion
import { sanitizePhoneNumber, isValidNigerianPhoneNumber } from './functions';
```

andela-hpopoola (approved):
Thank you for working on this. For now, this looks good to me. We might need to rewrite some of the functions / implementation when we introduce SMS for foreign numbers. Feel free to merge after fixing the comments. Thank you.

PR #240, opened 2021-09-01
#179316664 add value added service

tobieeniafe:
#### What does this PR do?
The purpose of this PR is to...

Have the following working...

`/api/v1/value-added-service`

#179316664


andela-hpopoola (commented):
Thank you for working on this. As discussed, Lets create a VAS Model and a requestVAS model to make this even more granular.

server/services/valueAddedService.service.js
```diff
@@ -0,0 +1,29 @@
...
+import { ErrorHandler } from '../helpers/errorHandler';
+import httpStatus from '../helpers/httpStatus';
+import { VALUE_ADDED_SERVICE_ACCESS_LEVEL } from '../helpers/constants';
+import VALUE_ADDED_SERVICES from '../helpers/valueAddedServices';
+
+export const getValueAddedServiceById = async (id) => ValueAddedService.findById(id).select();
+
+export const addValueAddedService = async (valueAddedService) => {
+  if (
+    VALUE_ADDED_SERVICES[valueAddedService.name].type !== valueAddedService.USER_ROLE.role &&
```
andela-hpopoola, line 11: I am not sure this is still needed as VAS will be added by admins only

server/services/valueAddedService.service.js
```diff
@@ -0,0 +1,29 @@
...
+    VALUE_ADDED_SERVICES[valueAddedService.name].type !== VALUE_ADDED_SERVICE_ACCESS_LEVEL.ALL
+  ) {
+    throw new ErrorHandler(
+      httpStatus.PRECONDITION_FAILED,
+      'You are not permitted to perform this action',
+    );
+  }
+
+  try {
+    return await new ValueAddedService({
```
andela-hpopoola, line 21: I think we should just allow the admin to add a new VAS without being dependent on the VAS file. Your current implementation will force us to update the VAS file anytime the admin wants to add a new VAS.

server/helpers/valueAddedServices.js
```diff
@@ -0,0 +1,26 @@
+import { VALUE_ADDED_SERVICE_ACCESS_LEVEL } from './constants';
+
+const VALUE_ADDED_SERVICES = {
```
andela-hpopoola, line 3: If the VAS is empty, lets save the values here in the database so that the admin can add more without requiring human intervention.

```suggestion
const INITIAL_VAS = {
```

server/helpers/valueAddedServices.js
```diff
@@ -0,0 +1,26 @@
...
+  SURVEY_PLAN_INVESTIGATION: {
+    name: 'Survey Plan Investigation',
+    price: 30_000,
+    type: VALUE_ADDED_SERVICE_ACCESS_LEVEL.USER,
```
andela-hpopoola, line 7: As discussed, this is more related to Property than User. Please fix for others.
```suggestion
    type: VALUE_ADDED_SERVICE_ACCESS_LEVEL.PROPERTY,
```

server/routes/index.js
```diff
@@ -49,4 +50,5 @@ export default (app) => {
   app.use('/api/v1/assign-badge', assignedBadge);
   app.use('/api/v1/bank-account', bankAccount);
   app.use('/api/v1/property-video', propertyVideo);
+  app.use('/api/v1/value-added-service', valueAddedService);
```
andela-hpopoola, line 53: As discussed above, I think we can use `vas` and it would still make complete sense.
```suggestion
  app.use('/api/v1/vas', valueAddedService);
```

andela-hpopoola (changes requested):
Thank you for working on this. This is still in progress but can we change the model name and other related services from `valueAddedService` to just `vas` to make it shorter. Also, don't forget to work on the implementation of `requestVas` which has a many-to-many relationship with the current VAS model. Thank you.

andela-hpopoola (changes requested):

server/routes/vas.routes.js
```diff
@@ -0,0 +1,126 @@
...
+ *      description: Updates existing Value added service
+ *     responses:
+ *      '200':
+ *        description: VAS updated
+ *      '400':
+ *        description: Error updating VAS
+ *      '500':
+ *       description: Internal server error
+ */
+router.put('/', authenticate, isAdmin, schemaValidation(updateVASSchema), VASController.updateVas);
```
andela-hpopoola, line 56: For javascript variables, methods, models, and services can we write VAS in normal letter conventions to ensure consistency. I see lots of `vas`, `Vas`, and `VAS`  in different places for different things in the code. 

To maintain coding consistency and prevent unnecessary typos in the future, can we just try VAS like normal text like User, or Vendor and ensure we follow the proper javascript naming convention for names that involve VAS.

```suggestion
router.put('/', authenticate, isAdmin, schemaValidation(updateVasSchema), VasController.updateVas);
```

andela-hpopoola (changes requested):
Thank you for the updates. Can we wrap this up for VAS so that we can focus on VASRequest? Let's write the necessary tests so that we can move to the next thing. Thank you.

andela-hpopoola (approved):

PR #241, opened 2021-09-01
#179401061 ensure single portfolio and single property are synced

tobieeniafe:
#### What does this PR do?
Ensure single portfolio and single property are synced

Ensure single portfolio and single property are synced
Make testimonial for property per user

`POST /api/v1/property/testimonial`

#179401061


andela-hpopoola (changes requested):
Thank you for working on this. It looks like you fixed the second part of the story, I will approve once you fix the first part. Thank you.

> Testimonial is currently missing in single portfolio endpoint
> Testimonial should be added per user (not per offer) to avoid confusion

PR #242, opened 2021-09-05
#179316675 user should pay for VAS - Part 1

tobieeniafe:
#### What does this PR do?
User should pay for VAS

User should pay for VAS

`/vas/request/`

#179316675


andela-hpopoola (commented):
Thank you for working on this. I took a quick look at this, lets update the code to focus only on VasRequest and updating VasRequest to the right status.

1. I think we should be able to use a single service to update all of the statuses. When a status is sent, we can check a list of all status and update it if the status is valid.
2. Lets remove all payments references in this PR and save them for the next implementation. For consistency, add ` - Part 1 of 3` to the PR's title.
```
#179316675 user should pay for VAS - Part 1 of 3
```
3. The next PR should be the same title but with `Part 2 of 3` and should focus on  updating the offline payments only (I separated this with transaction model) to reduce the number of files and tests we would be working with.
4. The last PR should be with `Part 3 of 3` and should focus on online payments and transactions.

Thank you.

server/models/vasRequest.model.js
```diff
@@ -45,10 +45,10 @@ const VasRequestSchema = new mongoose.Schema(
     propertyId: {
       type: ObjectId,
     },
-    resolvedBy: {
+    lastUpdatedBy: {
```
andela-hpopoola, line 48: Not sure of your thought process for this, but the `lastUpdated` (Date) looks similar to the `updatedAt` that is included in the timestamp with `createdAt`. Can we revert this back to `resolvedBy` and `resolvedDate`. If we need to track other dates, lets add them explicitly.

server/services/vasRequest.service.js
```diff
@@ -83,14 +88,106 @@ export const setVasRequestStatusToInProgress = async (id) => {
   try {
     return await VasRequest.findByIdAndUpdate(
       vasRequest._id,
-      { $set: { status: VAS_REQUEST_STATUS.IN_PROGRESS } },
+      { $set: { ...updateStatusInfo, lastUpdated: Date.now() } },
```
andela-hpopoola, line 91: Before updating the status, can we check that we can update the old status to the new one before updating. For example, we should be unable to update the status from `IN_PROGRESS` to `RESOLVED`.

andela-hpopoola: Also, for status lets add `CANCELLED` and `REFUNDED` in case a user wants to cancel or refund. We might need to confirm this from Namo if we support refund for this.

server/routes/index.js
```diff
@@ -50,5 +51,6 @@ export default (app) => {
   app.use('/api/v1/assign-badge', assignedBadge);
   app.use('/api/v1/bank-account', bankAccount);
   app.use('/api/v1/property-video', propertyVideo);
-  app.use('/api/v1/vas', valueAddedService);
+  app.use('/api/v1/vas/request', valueAddedService);
```
andela-hpopoola, line 54:
```suggestion
  app.use('/api/v1/vas/request', valueAddedServiceRequest);
```

server/routes/index.js
```diff
@@ -50,5 +51,6 @@ export default (app) => {
...
+  app.use('/api/v1/vas', valueAddedServiceRequest);
```
andela-hpopoola, line 55:
```suggestion
  app.use('/api/v1/vas', valueAddedService);
```

andela-hpopoola (changes requested):
Thank you for making the updates. I add a few more comments to ensure we get this right.

server/models/vasRequest.model.js
tobieeniafe: The `lastUpdatedBy` felt better since every status change wouldn't always be resolving it. So to know which admin made what change. Just for reference purposes. 
PS: Nice catch on the `lastUpdated `

andela-hpopoola: Noted. Thank you for clearing that. 

However, my point is that if need to know the admin that updates each of the statuses that we shouldn't override the previous ones.

For Example,
Admin A approves Payment - `lastUpdatedBy` contains Admin A id
Admin B updates to Pending - `lastUpdatedBy` is overridden with Admin B id
Admin C resolves status - `lastUpdatedBy` now contains Admin C id

From your analogy, if there is a  problem with the admin that approves Payment, we won't know since `lastUpdatedBy`  would currently contain Admin C id.

If Namo confirms that there are no refunds, then we would need only the `resolvedBy` field as explained below:

`REQUESTED` - We know this was done by the userId
`PAYMENT_VERIFIED` - Admin Id is already saved in transaction id (so I see no reason why we need to add it here)
`IN_PROGRESS` - I am not really sure we need to save the Id of the admin who updated the status to In_progress , if the VAS is not in progress, then the admin that notices this can start the procedure to make it be in progress.
`RESOLVED`: 'Resolved' - This is the only thing I feel we should save for now.

PS: These are just suggestions and `lastUpdatedBy` is a good idea. However, if you want to implement it that way then we should not override the Ids.

I hope this clear things up.

server/schemas/vasRequest.schema.js
```diff
@@ -0,0 +1,26 @@
...
+import { VAS_REQUEST_STATUS } from '../helpers/constants';
+
+const status = Joi.string()
+  .label('Status')
+  .valid(...Object.values(VAS_REQUEST_STATUS).map(String))
+  .required();
+
+export const addVasRequestSchema = Joi.object({
+  vasId: requiredObjectId('VAS Id'),
+  offerId: optionalObjectId('Offer Id'),
```
andela-hpopoola, line 12: In the next PR, when we find a solution to handling multiple models for transactions, we would use the same fix for this too.

server/services/vasRequest.service.js
tobieeniafe: Just spoke to Namo about this. And his take was; once payment is made, there can't be cancellation or refunds as payments are made to 3rd parties.

server/routes/vasRequest.routes.js
```diff
@@ -0,0 +1,163 @@
...
+ *      '404':
+ *        description: Request not found
+ *      '500':
+ *       description: Internal server error
+ */
+router.put(
+  '/status',
+  authenticate,
+  isAdmin,
+  schemaValidation(updateVasRequestStatusSchema),
```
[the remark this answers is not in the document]
tobieeniafe: `PUT /vas/request/` is used by the User or Vendor to update the basic details for the VAS request (maybe the request was made for the wrong property or offer) as long as the status is still in the default. `PUT /vas/request/status` is used by the admin solely to update the status.

test/controllers/vasRequest.controller.test.js
```diff
@@ -0,0 +1,806 @@
...
+      { generateId: true },
+    );
+
+    beforeEach(async () => {
+      vendorToken = await addUser(vendorUser);
+      userToken = await addUser(regularUser);
+      adminToken = await addUser(adminUser);
+    });
+
+    describe('Pagination Tests', () => {
```
andela-hpopoola, line 643: When viewing all VAS Requests, we should also add the value test to confirm that the right information is returned for vasInfo, userInfo, propertyInfo and offerInfo.

test/controllers/vasRequest.controller.test.js
```diff
@@ -0,0 +1,806 @@
...
+      vasRequestId: vasRequest._id,
+      status: VAS_REQUEST_STATUS.IN_PROGRESS,
+    };
+
+    beforeEach(async () => {
+      adminToken = await addUser(adminUser);
+      await addVasRequest(vasRequest);
+    });
+
+    context('when data and token is valid', () => {
```
andela-hpopoola, line 430: Lets add one more test here to confirm that the status history is properly saved when multiple status are updated at different times. This is to confirm that the first and last value are saved as expected when a VAS Request status is updated to `IN_PROGRESS` and later to `RESOLVED`. We should confirm that the 2 statuses are properly saved in that array.

andela-hpopoola (approved):
Thank you for working on this I noticed you made some other great decisions to ensure that all the comments are properly resolved (Nice work). Please note that I noticed some other edge cases but I believe we would resolve them in other PRs. 

Feel free to merge after fixing and updating the additional tests.

PR #245, opened 2021-09-07
#179479036 remove `$nin` to enable users see owned properties

tobieeniafe:
#### What does this PR do?
Remove `$nin` to enable users see owned properties

Remove `$nin` to enable users see owned properties in search

`GET /api/v1/property/search`

#179479036


andela-hpopoola (changes requested):

Now that we have removed `assignedTo` here, it looks like we can remove it from the property model and use the status in Offer to confirm assigned properties. 

Can you investigate where `assignedTo` is used and replace all instances with the relevant statuses. I am not 100% sure it is still needed, so your investigation will be really helpful.

andela-hpopoola (approved):

PR #248, opened 2021-09-11
#179401104 remove PII when non admin is viewing user information

tobieeniafe:
#### What does this PR do?
Remove PII when non admin is viewing user information

Remove PII when non admin is viewing user information

NA

#179401104


test/controllers/offer.controller.test.js
```diff
@@ -1322,12 +1322,37 @@ describe('Offer Controller', () => {
       });
 
       context('with a valid token & id', () => {
```
andela-hpopoola, line 1324:
```suggestion
      context('with a valid admin token', () => {
```

server/helpers/middleware.js
```diff
@@ -208,3 +208,15 @@ export const isUserOrAdmin = async (req, res, next) => {
...
   return null;
 };
+
+export const partiallyAuthenticate = async (req, res, next) => {
+  if (!req.headers.authorization) {
+    req.user = {};
+    next();
+  } else {
+    const { id } = decodeToken(req.headers.authorization);
+    const user = await getUserById(id);
```
andela-hpopoola, line 218: If the user is banned, we should treat them like an unauthenticated user since authentication is optional here.

// if user is banned
// req.user = {};

test/controllers/property.controller.test.js
```diff
@@ -2879,11 +2879,46 @@ describe('Property Controller', () => {
...
+              expect(res.body.portfolio.videos[0]._id).to.be.eql(approvedVideo._id.toString());
+              expect(res.body.portfolio.videos[0].status).to.be.eql(PROPERTY_VIDEO_STATUS.APPROVED);
+              expect(res.body.portfolio.testimonials.length).to.be.eql(1);
+              expect(res.body.portfolio.testimonials[0].userId).to.be.eql(
+                regularUser._id.toString(),
+              );
+              expect(res.body.portfolio.testimonials[0].user._id).to.be.eql(
+                regularUser._id.toString(),
+              );
+              expect(res.body.portfolio.testimonials[0].user.firstName).to.be.eql(
```
andela-hpopoola, line 2908: To ensure that we are not exposing PII, let's create a test function `expectResponseToExludePII` or update the `expectResponseToExcludeSensitiveUserData` function to ensure that PII is not returned for a user info.

test/controllers/user.controller.test.js
```diff
@@ -4333,4 +4335,114 @@ describe('User Controller', () => {
...
+            expect(res.body.vendor.phone).to.be.eql(vendorUser.phone);
+            expect(res.body.vendor.phone2).to.be.eql(vendorUser.phone2);
+            expect(res.body.vendor.address).to.be.eql(vendorUser.address);
+            expect(res.body.vendor.properties.length).to.be.eql(3);
+            done();
+          });
+      });
+    });
+
+    context('with user and vendor token', () => {
```
andela-hpopoola, line 4386: I don't think this is necessary since we are returning the same information for everyone, we only need to test that an authenticated role can access the endpoint.

test/controllers/user.controller.test.js
```diff
@@ -4333,4 +4335,114 @@ describe('User Controller', () => {
...
+    const endpoint = `/api/v1/user/vendor/${slug}`;
+
+    beforeEach(async () => {
+      adminToken = await addUser(adminUser);
+      userToken = await addUser(regularUser);
+      vendorToken = await addUser(vendorUser);
+      await Property.insertMany([...vendorProperties, ...properties]);
+    });
+
+    context('with admin token', () => {
```
andela-hpopoola, line 4367: We don't need to display PII here for everyone, also an admin can use the get user id endpoint to get all of the vendors information if necessary.

server/schemas/user.schema.js
```diff
@@ -50,6 +50,7 @@ const vendorCompanyName = Joi.object().keys({
 
 const vendor = Joi.object().keys({
   bankInfo,
+  bio: optionalString('Bio'),
```
andela-hpopoola, line 53: Let`s update the related tests to ensure that this is properly saved and returned where necessary.

test/controllers/user.controller.test.js
```diff
@@ -4333,4 +4335,114 @@ describe('User Controller', () => {
...
+          [method](endpoint)
+          .set('authorization', adminToken)
+          .end((err, res) => {
+            expect(res).to.have.status(200);
+            expect(res.body.success).to.be.eql(true);
+            expect(res.body.vendor._id).to.be.eql(vendorUser._id.toString());
+            expect(res.body.vendor.email).to.be.eql(vendorUser.email);
+            expect(res.body.vendor.phone).to.be.eql(vendorUser.phone);
+            expect(res.body.vendor.phone2).to.be.eql(vendorUser.phone2);
+            expect(res.body.vendor.address).to.be.eql(vendorUser.address);
```
andela-hpopoola, line 4379: Not sure if we agreed that the address is also PII.

test/controllers/user.controller.test.js
```diff
@@ -4333,4 +4335,114 @@ describe('User Controller', () => {
...
+            expect(res.body.vendor.properties.length).to.be.eql(3);
```
andela-hpopoola, line 4380: Also we need to add tests to ensure that other relevant information are returned for this endpoint. E.g Companyname, logo and other non personal information should be tested.

andela-hpopoola (changes requested):
Thank you for working on this. I left a couple of comments to ensure that we get this right. Thank you for the hard work.

andela-hpopoola (approved):
Thank you for making the fixes. I left a few comments but this should be ready to merge once fixed.

PR #249, opened 2021-09-15
#179316675 user should pay for VAS - Part 3

tobieeniafe:
#### What does this PR do?
Update transaction model to accommodate multiple models

Update transaction model to accommodate multiple models

`/transaction/`

#179316675


server/services/transaction.service.js
```diff
@@ -72,55 +74,50 @@ export const getPaymentDuration = async (offerId) => {
...
-    const isFirstPayment = await isUserFirstPayment(offer.userId);
+    const isFirstPayment = await isUserFirstPayment(model.userId);
     if (isFirstPayment) {
-      await assignBadgeAutomatically(AUTOMATED_BADGES.USER_FIRST_PAYMENT, offer.userId);
+      await assignBadgeAutomatically(AUTOMATED_BADGES.USER_FIRST_PAYMENT, model.userId);
     }
 
-    const paymentType = await getPaymentDuration(offer._id);
+    if (transaction.model.type === MODEL.OFFER) {
+      await generateNextPaymentDate({
```
andela-hpopoola, line 94: Let's add all of the code here in a new function `processOfferTransaction` to make this function lean. This also makes it easier to add more codes for other model types without bloating this function.

server/models/transaction.model.js
```diff
@@ -73,10 +75,6 @@ const TransactionSchema = new mongoose.Schema(
       type: String,
       required: true,
     },
-    propertyId: {
```
andela-hpopoola, line 76: You might need to confirm that we didn't use `propertyId` to aggregate transactions anywhere. If we do, we need to think of a way to reimplement this.

server/services/transaction.service.js
```diff
@@ -72,55 +74,50 @@ export const getPaymentDuration = async (offerId) => {
...
-
   try {
     const newTransaction = await new Transaction({
       ...transaction,
-      userId: offer.userId,
-      propertyId: offer.propertyId,
-      vendorId: offer.vendorId,
-      addedBy: offer.vendorId,
-      updatedBy: offer.vendorId,
+      vendorId: model.vendorId,
```
andela-hpopoola, line 85: This place is a little bit tricky, however, I think we need to confirm that `userId` exists first and use it here instead of vendorId (vendorId is not used by VAS).

```suggestion
      userId: model.userId,
```

For now, we might leave the propertyId and vendorId but will not make them required for now. When we get it right, maybe we will add another object and add all the related models, I am sure of the best solution for now. We need to think of a better way to make this work flawlessly. 

For `addedBy` and `updatedBy`, let's make them optional too. However, we need to use the Id of the user calling this service (if it is available).

server/services/transaction.service.js
```diff
@@ -72,55 +74,50 @@ export const getPaymentDuration = async (offerId) => {
 };
 
 export const addTransaction = async (transaction) => {
```
andela-hpopoola, line 76: To manage expectations, please note that this function is called in 2 ways:

1. Manually - When an admin is resolving offline payment - `resolveOfflinePayment`
2. Automatically - When payment is made via Paystack - `addPaymentToTransaction`

So, lets put this into consideration when making our changes.

server/services/vasRequest.service.js
```diff
@@ -5,7 +5,9 @@ import httpStatus from '../helpers/httpStatus';
 import { USER_ROLE, VAS_REQUEST_STATUS } from '../helpers/constants';
 // eslint-disable-next-line import/no-cycle
 import { getVasById } from './vas.service';
+// eslint-disable-next-line import/no-cycle
```
andela-hpopoola, line 8: [Tech Debt] To fix the `eslint-disable-next-line` issue in the future, we might need to create a custom file or function for all of the `getMODELById` so that they are not reimported into other files.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to ensure we are on the right track for this. Thank you.

server/models/transaction.model.js
```diff
@@ -30,8 +29,6 @@ import mongoose from 'mongoose';
  *            type: string
  *          userId:
  *            type: string
- *          vendorId:
```
andela-hpopoola, line 36: Lets restore this for now and the one below.

server/services/transaction.service.js
```diff
@@ -73,16 +74,56 @@ export const getPaymentDuration = async (offerId) => {
...
+      referralId: referral._id,
+      transactionId: newTransaction._id,
+      amountPaid: newTransaction.amount,
+      offerId: model._id,
+    });
+  }
+};
+
 export const addTransaction = async (transaction) => {
+  let offerTransactionParameters = {};
```
andela-hpopoola, line 108: This is better as transaction parameters to ensure that it can be reused by other models.
```suggestion
let transactionParameters = {};
```

test/controllers/property.controller.test.js
```diff
@@ -1275,7 +1280,7 @@ describe('Property Controller', () => {
       {
         userId: regularUser._id,
         propertyId: property._id,
-        offerId: offer._id,
+        model: { id: offer._id, type: VALID_TRANSACTION_MODELS.OFFER },
```
andela-hpopoola, line 1283: Using `MODEL.OFFER` is better here to ensure consistency . Please update for other ones too.

```suggestion
model: { id: offer._id, type: MODELS.OFFER },
```

andela-hpopoola (changes requested):
This looks good to me, thank you for working on this. Can you please add some transactions test for VAS (you can replicate some of the tests for offer transactions)?

Also, when getting all transactions, we want to be sure that the updated model loads the right aggregate for offer, property, vas, and vas request.

Lastly, let's not forget to look through the transaction service to be sure that we fix all the updates we made in the model.

Since I will be missing tomorrow's standup, after this, let's please focus on creating Part 4 to conclude the major work and add INITIAL_VAS to the database, Part 5 to fix the `propertyId` and `vendorId` tech debt so that we can accommodate other Ids in the future (like vasRequestId), I will need to work on Part 6 to integrate the updated model into Paystack integration.

Thank you.

tobieeniafe:
@andela-hpopoola  So I came into a jam halfway while trying to add vasRequests to the `/transaction/all` endpoint. The current workflow (which is focused on only offers) is that, when a user visits the endpoint, all transactions made by the user are returned. Using the `userId`. And when it's a vendor, all transactions with the matching `vendorId` are returned.

Now in the case of vasRequests, the vendor can be the user in this case, ie trying to get all vasRequests they've paid for. Which would return zero as a vendor token would match with the `vendorid` key.

Now knowing we would be removing the `vendorId` key soon, this would mean that it's either vendors would not be able to see offer payments made by the user to them, or they wouldn't be able to view their vas requests. For now that is (still trying to figure out a work around for this tho)

andela-hpopoola:
> @andela-hpopoola So I came into a jam halfway while trying to add vasRequests to the /transaction/all endpoint. The current workflow (which is focused on only offers) is that, when a user visits the endpoint, all transactions made by the user are returned. Using the userId. And when it's a vendor, all transactions with the matching vendorId are returned.

Thank you for pointing this out. This actually doesn't look like a blocker. In the previous implementation, the vendor is not making any payments to us, so we are getting the payments the vendor is receiving via `vendorId`. Now, the vendor is making payment to us via VAS, so we should get the payments the vendor is making (via `userId`) and the one the vendor is  receiving (via `vendorId`)

> Now in the case of vasRequests, the vendor can be the user in this case, ie trying to get all vasRequests they've paid for. Which would return zero as a vendor token would match with the vendorid key.

The vendor doesn't need to be the user, we only need to get what they have paid and what the transaction they are expecting via vendorId and userId. When getting all transactions, we would get all transactions that matches where vendorId=`userId` or userId=`userId`. `userId` here refers to the user id of a vendor (or vendor Id).

> Now knowing we would be removing the vendorId key soon, this would mean that it's either vendors would not be able to see offer payments made by the user to them, or they wouldn't be able to view their vas requests. For now that is (still trying to figure out a work around for this tho)

Not sure what you mean by this and it looks like you are making the wrong assumption. We are not removing `vendorId` key soon, we are only refactoring it (the same way we refactored `offerId` but it might be a different implementation). If the `vendorId` is not needed, we won't add it in the first place. We are only refactoring the Ids that are not primarily involved in the transaction (like vendorId, propertyId, vasRequestId and other unforeseen Ids in the future).

The reason for refactoring this is to accommodate future updates when adding other model Ids to that transaction. For example, for transactions involving VAS, we might need to save the `vasRequestId`, but the `vendorId` and `propertyId` that were previously required are not needed in this case. The focus is to ensure that each transaction contains what is required for it and reducing the number of things that are not required.

I hope this clears the confusion. Please let me know if you need more information or have more questions.

server/services/transaction.service.js
```diff
@@ -207,11 +240,7 @@ export const getAllTransactions = async (user, { page = 1, limit = 10, ...query
     transactionOptions.pop();
   }
 
-  if (user.role === USER_ROLE.VENDOR) {
```
andela-hpopoola, line 198: I think we still need to match `vendorId` for vendors.

test/controllers/transaction.controller.test.js
```diff
@@ -152,36 +151,42 @@ describe('Transaction Controller', () => {
...
-        activated: true,
-        vendor: { verified: true },
-      },
+    const vas = VasFactory.build({}, { generateId: true });
+
+    const vendorVasRequest = VasRequestFactory.build(
+      { vasId: vas._id, propertyId: property._id, user: vendorUser },
       { generateId: true },
     );
-    const vendor2Property = PropertyFactory.build({ addedBy: vendor2._id }, { generateId: true });
```
andela-hpopoola, line 158: It looks like you totally removed the property transactions, can we add it back 🙏🏽

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to ensure that we covered all edges cases for transactions, however, it looked like you missed some other parts too.

test/controllers/transaction.controller.test.js
```diff
@@ -205,66 +215,73 @@ describe('Transaction Controller', () => {
...
+                expect(res.body.result[0].userInfo._id).to.have.string(regularUser._id.toString());
+                expect(res.body.result[8]._id).to.be.eql(vendorVasTransactions[0]._id.toString());
+                expect(res.body.result[8].model.id).to.be.eql(vendorVasRequest._id.toString());
+                expect(res.body.result[8].model.type).to.be.eql(MODEL.VAS_REQUEST);
+                expect(res.body.result[8].remittance.amount).to.be.eql(
+                  vendorVasTransactions[0].remittance.amount,
                 );
-                expect(res.body.result[0].remittance.amount).to.be.eql(
-                  vendor2Transactions[0].remittance.amount,
+                expect(res.body.result[8].remittance.by).to.be.eql(
```
andela-hpopoola, line 261: Remittance and its fields should be null for VAS_REQUESTS transactions. Can we confirm that they are null  or undefined instead of testing them against `vendorVasTransactions[0].remittance.by.toString()`.

Also, let's not forget to test for the amount and the userId.

server/services/transaction.service.js
```diff
@@ -314,9 +361,47 @@ export const getOneTransaction = async (transactionId, user) => {
         as: 'vendorInfo',
       },
     },
-    { $unwind: '$propertyInfo' },
+    {
+      $lookup: {
```
andela-hpopoola, line 365: Can we add another test to confirm that the new values are returned for single transaction for VAS_REQUESTS. That means we would have one test for OFFER transaction and another test for VAS_REQUESTS transaction.

andela-hpopoola (approved):
Thank you for making the updates. I left a few comments on the tests, but this looks good overall.

PR #250, opened 2021-09-20
#179637243 add optional ids to support new model type

tobieeniafe:
#### What does this PR do?
Add optional ids to support new model type

Add optional ids to support new model type

`npm run test`

#179637243


server/models/transaction.model.js
```diff
@@ -61,6 +61,21 @@ const TransactionSchema = new mongoose.Schema(
         type: String,
         required: true,
       },
+      offerId: {
```
andela-hpopoola, line 64: I noticed a type confusion pattern here which might give us additional issues in the future.

`vasId` and `offerId` are already part of `model.id` so it is causing unnecessary duplication here. If the the `model.type` is `OFFER` that means that we would need to set `model.id` to the `OFFER` Id and also set `offerId` to the `OFFER` Id. This might cause unnecessary confusion in the future if the value of both fields is different, as we would not be sure of the right one. 

The best fix in my own opinion is to remove `model.id` and use the `modelId` format to get the id. That way, if `model.type` is offer, we will use `offerId`, if it is `vasRequests`, we would use `vasRequestId`. This would prevent unnecessary duplication and future confusion.

The other fix would be to remove the `Id` of those we are certain correlates with transaction (for now VASRequest and Offer). However, the problem of this is that it could cause future issue for situations where we want to save the `offerId` and `vasRequestId` as the same time. We won't be able to do this with `model.id`.

andela-hpopoola (changes requested):
Thank you for making the changes. I left an important comment on some future issues we might have with the current methodology. I hope we can fix that now instead of postponing it till later. Thank you.

server/models/transaction.model.js
tobieeniafe: I understand your concerns, but the `offerId` in this case, refers to the offer that could be gotten from a `vasRequest`.

andela-hpopoola (changes requested):

I have taken a deep look at this and I can confirm that we need to remove the `model.id` implementation to prevent any tight coupling of the model. Lets use the `modelId` format, e.g for `offer` we would use `offerId`, for property, we would use `propertyId`. The main Id will be recognized by the model.type.

This implementation can be done in this PR or a new PR.

Thank you.

PR #252, opened 2021-09-21
#179637260 fix pending bugs and comments - Part 5

tobieeniafe:
#### What does this PR do?
Fix pending bugs and comments

-Add route to return single vas request
-Append `userInfo` and `vasRequestinfo` to transactions when viewing single vas

`GET /api/v1/vas/request/:id`
`GET /api/v1/vas/:id`

#179637260


server/services/vasRequest.service.js
```diff
@@ -81,11 +81,40 @@ export const verifyVasRequestInput = async (vasDetails) => {
   }
 };
 
+export const similarVasRequestExists = async (vasDetails) => {
+  let searchparameters = {
+    userId: ObjectId(vasDetails.userId),
+    vasId: ObjectId(vasDetails.vasId),
```
andela-hpopoola, line 87: I believe the current implementation can be improved. I think you should be able to add the conditional field members here:
```suggestion
    vasId: ObjectId(vasDetails.vasId),
    ...(vasDetails.propertyId && {propertyId: ObjectId(vasDetails.propertyId)}),
    ...(vasDetails.offerId && {offerId: ObjectId(vasDetails.offerId)}),
```

**Source**
https://stackoverflow.com/questions/11704267/in-javascript-how-to-conditionally-add-a-member-to-an-object/40560953#40560953

--


Another way is to add the 4 fields in an array and add them to the `searchParameter` if a value exists for each field.

```js
const fields = ['userId', 'vasId', 'offerId', 'propertyId'];
// loop through fields, checking if vasDetails[field] exists and then assign to searchParameter object.
```

server/services/vasRequest.service.js
```diff
@@ -81,11 +81,40 @@ export const verifyVasRequestInput = async (vasDetails) => {
   }
 };
 
+export const similarVasRequestExists = async (vasDetails) => {
+  let searchparameters = {
```
andela-hpopoola, line 85:
```suggestion
  let searchParameters = {
```

server/services/vas.service.js
```diff
@@ -129,13 +140,21 @@ export const getAllVas = async ({ page = 1, limit = 10, ...query } = {}) => {
 };
 
 export const getOneVas = async (id) => {
-  const vas = await getVasById(id).catch((error) => {
-    throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
-  });
+  const vas = await Vas.aggregate([
+    { $match: { _id: ObjectId(id) } },
+    {
```
andela-hpopoola, line 145: Since we are loading transactions here, I think it is also better to load additional info from the transaction model like vasRequestInfo and userInfo to make the returned transaction more meaningful

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to improve the functionality of the current implementation.

Also, I think you missed adding the `get /vas/request/:id` implementation (the second item in Pivotal Tracker description field for this story)

```
add get vas request by id endpoint (get one vas request)
```


Lastly, in the description part of the PR:

> Description of Task to be completed?
Fix pending bugs and comments

I think we can improve this by listing all the comments and pending bugs that were fixed.

PR #253, opened 2021-09-23
#179637116 update offer status when new payment is made - Part 1

tobieeniafe:
#### What does this PR do?
Update offer status when new payment is made

-Remove `assignedTo` from property model
-Delete `assignPropertyToUser` function

`npm run test`

#179637116


server/services/offer.service.js
```diff
@@ -519,16 +519,10 @@ export const acceptOffer = async (offerToAccept) => {
...
-
   await activatePendingUserReferral(offer);
 
   try {
-    await assignPropertyToUser({
-      propertyId: offer.propertyId,
-      userId: offer.userId,
-      vendor,
-    });
+    // offer should be INTERESTED here
```
andela-hpopoola, line 525: A gentle reminder to remove this comment in next commit.

test/controllers/enquiry.controller.test.js
```diff
@@ -893,13 +893,13 @@ describe('Enquiry Controller', () => {
     );
     const vendorProperties = PropertyFactory.buildList(
       13,
-      { addedBy: vendorUser._id },
+      { addedBy: vendorUser._id, availableUnits: 4 },
```
andela-hpopoola, line 896: Instead of including `availableUnits` here by default, I think it is better to update the property factory to cater for availableUnits (just as we have implemented the `generatedId` field). Something like this (emphasis on the `.after` part)

```js
export default new Factory()
  .option('generateId', false)
  .option('optionalFields', false)
  .after((property, options) => {
    const additionalOptions = {
      ...(options.generateId && { _id: mongoose.Types.ObjectId() }),
      ...(options.optionalFields && { availableUnits: property.units }),
    };
    return { ...additionalOptions, ...property };
  })
  .sequence('name', (i) => `${i} bedroom apartment`)
...

```

With the above implementation, by default,  when `optionalFields` is set to true (like we do for `generateId`), `availableUnits` will be equal to the default value of `units`. However, if available units is given by default, then that value will be used. 

PS: Even though the updates above might not reduce your typing, however, it ensures that the data we are using are consistent. The problem with your current implementation is that the default units in `property.factory.js` is `5` but you are setting it to `4` here; not sure if this is deliberate or a typo, however, newly created properties should have the same value for `availableUnits` and `units` unless we want to explicitly test for `availableUnits`.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on updating the factory to reflect the current changes for consistency. Thank you.

server/services/totalCount.service.js
```diff
@@ -26,9 +26,11 @@ const getTotalCount = async (user) => {
 
   const portfolioMatchObject = { $or: VALID_PORTFOLIO_OFFER };
 
+  const PropertyModel = user.role === USER_ROLE.USER ? Offer : Property;
+
   if (user.role === USER_ROLE.USER) {
     matchObject = { userId: ObjectId(user._id) };
-    propertyMatchObject = { assignedTo: { $in: [ObjectId(user._id)] } };
+    propertyMatchObject = { userId: user._id, status: OFFER_STATUS.ALLOCATED };
```
andela-hpopoola, line 33: I think this should be a match for 
- `OFFER_STATUS.ASSIGNED` - user has made at least 1 payment
-  `OFFER_STATUS.ALLOCATED`- user has reached the allocated threshold set by vendor
-  `OFFER_STATUS.RESOLVED` - user has completed payment for the property. We might need to update this to `COMPLETED_PAYMENT` instead of resolved to make it more explicit.

**Note:** I noticed that `OFFER_STATUS.RESOLVED` is missing in most of the checks we have done. Lets review this and include it where necessary.

Also, it makes more sense if we can write functions that explains what they do -

1. OwnedProperty - Status is ASSIGNED, ALLOCATED OR COMPLETED_PAYMENT 
2. ValidPortfolio - Status is  INTERESTED, ASSIGNED, ALLOCATED OR COMPLETED_PAYMENT
3. PendingOffers - Status is GENERATED, PENDING_ADMIN_APPROVAL, PENDING_VENDOR_REVIEW
4. InvalidOffers - Status is CANCELLED or EXPIRED

**Todo:** 
- We need a cron job to set the status of expired Offers to `EXPIRED`.
- We need to remove the following offer statuses -  `OFFER_STATUS.NEGLECTED` and `OFFER_STATUS.REJECTED`. OFFER_STATUS.NEGLECTED is used in `acceptOffer` service but it is not defined in OFFER_STATUS.

From the above, here is the breakdown of the analyzed `OFFER_STATUS` :

```js
export const OFFER_STATUS = {
  GENERATED: 'Generated', // Offer has been created by vendor
  INTERESTED: 'Interested', // Offer has been signed by the user, no payment has been made yet
  ASSIGNED: 'Assigned', //  A minimum of first Payment has been made on the Offer
  ALLOCATED: 'Allocated', // User has reached the allocated threshold
  REJECTED: 'Rejected', // maybe we should remove this
  REACTIVATED: 'Reactivated', // Same as Generated, just an updated offer
  CANCELLED: 'Cancelled', // Vendor cancels the offer, e.g when a user is no longer interested (need to think of how to refund)
  RESOLVED: 'Resolved', // should be updated to COMPLETED_PAYMENT
  PENDING_ADMIN_APPROVAL: 'Pending Admin Approval',
  PENDING_VENDOR_REVIEW: 'Pending Vendor Review',
  EXPIRED: 'Expired' // Newly added for automatically expired offers
};
```

andela-hpopoola (changes requested):
**PROBLEM STATEMENT**

I noticed that the usage of `OFFER.STATUS` is not consistent with the statuses. This is mainly due to the fact that the statuses we have created are not properly defined. Here are some of the things I noticed:

1. OFFER_STATUS.RESOLVED` - This denotes that a user has completed payment for the property. I think it is better to update this to `COMPLETED_PAYMENT` instead of resolved to make it more explicit.

**Note:** I noticed that `OFFER_STATUS.RESOLVED` is missing in most of the checks we have done. Let's review this and include it where necessary.

2. Some of the places we are comparing multiple statuses seem to be missing one statuses. Also, some statuses are used without being defined. To fix this, lets create custom functions to ensure that the right statuses are used.

1. propertyIsOwnedByUser - Status is ASSIGNED, ALLOCATED OR COMPLETED_PAYMENT 
2. isValidPortfolio - Status is  INTERESTED, ASSIGNED, ALLOCATED OR COMPLETED_PAYMENT
3. isPendingOffers - Status is GENERATED, PENDING_ADMIN_APPROVAL, PENDING_VENDOR_REVIEW, REACTIVATED
4. offerHasBeenTerminated - Status is CANCELLED or EXPIRED

**Todo:** 
- We need a cron job to set the status of expired Offers to `EXPIRED`.
- We need to remove the following offer statuses -  `OFFER_STATUS.NEGLECTED` and `OFFER_STATUS.REJECTED`. OFFER_STATUS.NEGLECTED is used in `acceptOffer` service but it is not defined in OFFER_STATUS.

From the above, here is the breakdown of the analyzed `OFFER_STATUS` :

```js
export const OFFER_STATUS = {
  GENERATED: 'Generated', // Offer has been created by vendor
  INTERESTED: 'Interested', // Offer has been signed by the user, no payment has been made yet
  ASSIGNED: 'Assigned', //  A minimum of first Payment has been made on the Offer
  ALLOCATED: 'Allocated', // User has reached the allocated threshold
  REJECTED: 'Rejected', // maybe we should remove this
  REACTIVATED: 'Reactivated', // Same as Generated, just an updated offer
  CANCELLED: 'Cancelled', // Vendor cancels the offer, e.g when a user is no longer interested (need to think of how to refund)
  RESOLVED: 'Resolved', // should be updated to COMPLETED_PAYMENT
  PENDING_ADMIN_APPROVAL: 'Pending Admin Approval',
  PENDING_VENDOR_REVIEW: 'Pending Vendor Review',
  EXPIRED: 'Expired' // Newly added for automatically expired offers
};
```

server/services/offer.service.js
```diff
@@ -1035,3 +1028,20 @@ export const removeAdditionalTerms = async ({ offerId, vendorId }) => {
...
+
+export const updateStatusOfExpiredOffers = async () => {
+  const today = startOfDay(new Date());
+
+  const expiredOffers = await Offer.aggregate([
+    { $match: { status: OFFER_STATUS.GENERATED } },
+    { $match: { expires: { $lt: add(today, { days: 0 }) } } },
+  ]);
+
+  await Promise.all(
```
andela-hpopoola, line 1040: Let's not forget to send emails, notifications (and maybe Slack notifications in the next PR) to the users and vendors with expired offers.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments on ensuring that this is properly implemented. Lets not forget to add the following implementation in the next PRs:

1. Updating OFFER STATUS in all cases where necessary. We should have endpoints that cover all OFFER STATUSES that we have defined.
2. Ensuring that we create a reminder endpoint before an offer expires (maybe 2 days before expiry - we need to confirm from Namo). A similar implementation is used for Offer reminders.

Thank you.

server/helpers/constants.js
```diff
@@ -107,13 +94,40 @@ export const NOTIFICATION_ACTION = {
...
 ];
 
+export const OWNED_PROPERTY = [
+  { status: OFFER_STATUS.ASSIGNED },
+  { status: OFFER_STATUS.ALLOCATED },
+  { status: OFFER_STATUS.COMPLETED_PAYMENT },
+];
+
+export const PENDING_OFFER = [
+  { status: OFFER_STATUS.GENERATED },
```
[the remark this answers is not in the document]
tobieeniafe: `OFFER_STATUS.REACTIVATED` is the status given to an offer that expired, but was reactivated and a new offer generated from it. So I don't belive it should be a pending offer.

server/helpers/constants.js
```diff
@@ -107,13 +94,40 @@ export const NOTIFICATION_ACTION = {
...
+  { status: OFFER_STATUS.PENDING_ADMIN_APPROVAL },
+  { status: OFFER_STATUS.PENDING_VENDOR_REVIEW },
+];
+
+export const INVALID_OFFER = [{ status: OFFER_STATUS.CANCELLED }, { status: OFFER_STATUS.EXPIRED }];
+
 export const ACTIVE_PORTFOLIO_OFFER = [
```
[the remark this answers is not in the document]
tobieeniafe: `VALID_PORTFOLIO_OFFER` contains ` OFFER_STATUS.COMPLETED_PAYMENT` which means the offer has been completed and can no longer be referred to as active.

server/services/transaction.service.js
```diff
@@ -87,6 +96,12 @@ export const processOfferTransaction = async ({ model, newTransaction }) => {
 
   const paymentType = await getPaymentDuration(model._id);
 
+  const isFirstPaymentForOffer = await isFirstOfferPayment(model._id);
+
+  if (isFirstPaymentForOffer) {
+    await assignOffer(model._id);
```
andela-hpopoola, line 102: We need to add tests for this.

server/helpers/constants.js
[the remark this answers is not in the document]
andela-hpopoola: I am not sure of your definition of `active` but looking at the codebase and where the `ACTIVE_PORTFOLIO_OFFER` constant is used, I think it should contain `OFFER_STATUS.COMPLETED_PAYMENT`.

https://github.com/Highrachy/ballers-api/blob/501f98f42ef9ec1ec9b5c95e3b490f49612a1ef0/server/services/transaction.service.js#L283

If you don't add `OFFER_STATUS.COMPLETED_PAYMENT` here or change the implementation, then the person that has a contribution reward of `40 million`, will now have a contribution reward of `0 Naira` after he has completed payment.

server/helpers/constants.js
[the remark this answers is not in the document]
andela-hpopoola: `OFFER_STATUS.REACTIVATED` is the same as `OFFER_STATUS.GENERATED` but with the explanation you have given.

> OFFER_STATUS.REACTIVATED is the status given to an offer that expired, but was reactivated and a new offer generated from it. So I don't belive it should be a pending offer.


The only reason we used `OFFER_STATUS.REACTIVATED` is to differentiate it from `OFFER_STATUS.REACTIVATED`. 

`PENDING_OFFER` should contain offers that have not been signed by the user but are active, so `REACTIVATED` falls under this category too.

server/controllers/offer.controllers.js
```diff
@@ -242,6 +231,23 @@ const OfferController = {
       })
       .catch((error) => next(error));
   },
+
+  processExpiredOffers(req, res, next) {
+    processExpiredOffers()
+      .then((offers) => {
+        const message = `${pluralizeCount(offers.length, 'offer')} ${pluralize(
```
andela-hpopoola, line 238:
```suggestion
        const message = `${pluralizeCount(offers.length, 'offer has', 'offers have')}
```

server/services/offer.service.js
```diff
@@ -568,15 +566,17 @@ export const assignOffer = async (offerId, user) => {
...
   }
 
+  const property = await getOneProperty(offer.propertyId);
+
   try {
-    await Offer.findByIdAndUpdate(
+    await Property.findByIdAndUpdate(property._id, {
+      $set: { availableUnits: property.availableUnits - 1 },
+    });
+    return await Offer.findByIdAndUpdate(
```
andela-hpopoola, line 575: Before we assign the property here, I think it makes sense to check if the offer is a `PENDING_OFFER`. Not sure if this has been done.

andela-hpopoola (changes requested):
Thank you for working on this. Let's try to resist the urge to add more functionality to this PR, lets try to conclude this by fixing relevant comments now, adding necessary tests and fixing other functionalities in a different PR. Thank you.

server/helpers/constants.js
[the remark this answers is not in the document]
andela-hpopoola:
```
// REACTIVATED: 'Reactivated', // Same as Expired, just informing us that a new offer was created from this.
```

server/helpers/filters.js
```diff
@@ -220,6 +220,7 @@ export const USER_FILTERS = {
 export const PROPERTY_FILTERS = {
   ...ADDRESS_FILTERS,
   addedBy: { type: FILTER_TYPE.OBJECT_ID },
+  availableUnits: { type: FILTER_TYPE.INTEGER },
```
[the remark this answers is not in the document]
andela-hpopoola: I am unresolving this as some tests are still missing that we need to confirm that available unit is returned as expected. The fastest way to add this is to update the `expectResponseToContainNecessaryPropertyData` test function to include availableUnits as a 3rd parameter. If is is given we will use the given value, else we will use the `units`.

andela-hpopoola (approved):
Thank you for working on this. I left a few more comments on removing the OfferStatus allocation to enable us to write a better service to cater for this. I am approving this PR now as it is already a mess and very difficult to review. In the future, when the number of edited files exceeds 5 files (without tests), let's start thinking of breaking it down into multiple PR and sticking to one functionality (with discretion).

Here is the collection of the pending stuffs from this PR:

1. offer reminder - add in-app, sms, slack, email notifications
2. revisit cancel offer endpoint (remove enquiry set to false)
3. update available units when a property `units` is updated
4. consolidate assignOffer service to cater for completed payment, allocated, and assigned
5. confirm that `getAllUserOffers` service is still needed(differentiate from get all offers)

PR #254, opened 2021-09-24
#179637260 fix pending bugs or comments part 6

tobieeniafe:
#### What does this PR do?
Fix pending bugs or comments part 6

Remove `model.id` from transaction and offline payment models

`npm run test`

#179637260


test/controllers/offlinePayment.controller.test.js
```diff
@@ -208,7 +208,7 @@ describe('Offline Payment Controller', () => {
       context('when model id empty', () => {
         it('retruns error', (done) => {
```
andela-hpopoola, line 209:
```suggestion
        it('returns error', (done) => {
```

server/schemas/offlinePayment.schema.js
```diff
@@ -22,12 +22,14 @@ const optionalModelType = Joi.string()
   .optional();
 
 const requiredModel = Joi.object().keys({
-  id: requiredObjectId('Model id'),
+  offerId: optionalObjectId('Offer id'),
```
andela-hpopoola, line 25: When adding offline payments or transactions, we can make the `offerId` and `vasRequestId` to be required based on the selected modelType:

You can read this as: `When type is MODEL.OFFER then make OfferId required else make it optional`.

https://joi.dev/api/?v=17.4.2#anywhencondition-options

```js
  offerId: Joi.when('type', {
    is: MODEL.OFFER,
    then: requiredObjectId('Offer id'),
    otherwise: optionalObjectId('Offer id'),
  }),
```

The same implementation should be used for `vasRequestId`.

test/controllers/offlinePayment.controller.test.js
```diff
@@ -208,7 +208,7 @@ describe('Offline Payment Controller', () => {
       context('when model id empty', () => {
```
andela-hpopoola, line 208: Lets add a similar test for `vasRequestId`.

```suggestion
      context('when offer id empty', () => {
```

Also, let's add another test to ensure that an error is returned when the associated model is not given based on the model type (as advised on the Joi implementation).

Something like this:

```js
 context('when offer id is not given', () => {
        it('retruns error', (done) => {
          const offlinePayment = OfflinePaymentFactory.build({
            model: { type: MODEL.OFFER },
            bankId: bank._id,
          });
          request()
            .post('/api/v1/offline-payment')
            .set('authorization', userToken)
            .send(offlinePayment)
            .end((err, res) => {
              expect(res).to.have.status(412);
              expect(res.body.success).to.be.eql(false);
              expect(res.body.error).to.be.eql('"Offer id" is required');
              done();
            });
        });
      });

```

Similar tests should be added for VasRequestId too. Also, similar updates should be made in `transaction.controller.test.js` for both models.

andela-hpopoola (approved):
Thank you for the hard work on this. This actually look more straightforward than I thought and the implementation looks awesome. 

I left a few comments on updating the `Joi` validations and adding more test to ensure that model Ids are made required from Joi.schema.

Feel free to merge after making the changes but let me know if there are any blockers with my suggestions. Thank you.

PR #257, opened 2021-10-02
#179637116 update offer status when new payment is made - Part 2

tobieeniafe:
#### What does this PR do?
Update offer status when a new payment is made

-Add notifications for expired, assigned & allocated offers

`npm run test`

#179637116


server/services/offer.service.js
```diff
@@ -626,7 +644,7 @@ export const cancelOffer = async ({ offerId, vendorId }) => {
   try {
     await Enquiry.findByIdAndUpdate(offer.enquiryId, { $set: { approved: false } });
```
andela-hpopoola, line 645: When an offer is cancelled, I don't think we need to set Enquiry to false as explained in last PR.

server/services/offer.service.js
```diff
@@ -1059,13 +1077,41 @@ export const processExpiredOffers = async () => {
...
+    { $unwind: '$propertyInfo' },
   ]);
 
   await Promise.all(
     expiredOffers.map(async (offer) => {
       await Offer.findByIdAndUpdate(offer._id, { $set: { status: OFFER_STATUS.EXPIRED } });
+
+      const description = `Your offer for ${getFormattedName(offer.propertyInfo.name)} has expired`;
+
+      await createNotification(NOTIFICATIONS.OFFER_EXPIRED, offer.userId, {
```
andela-hpopoola, line 1106: We also need to create a Notification for the vendor for expired Notifications.

test/services/offer.service.test.js
```diff
@@ -1335,6 +1346,17 @@ describe('Offer Service', () => {
         const expiredOffer = await Offer.findById(expiredOffers[0]._id);
         expect(expiredOffer.status).to.eql(OFFER_STATUS.EXPIRED);
       });
+
+      context('when new notification is added', () => {
```
andela-hpopoola, line 1350:
```suggestion
      context('when an expired offer is processed', () => {
```

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments to ensure that we get this right.

server/services/property.service.js
```diff
@@ -40,7 +40,7 @@ export const addProperty = async (property) => {
   try {
     const addedProperty = await new Property({
       ...property,
-      availableUnits: property.units,
+      availableUnits: property.availableUnits || property.units,
```
andela-hpopoola, line 43: Not sure why you added this, I believe we should not allow `availableUnits` to be defined from a variable. Moreover, this is not defined in Joi schema, so I am not sure why you are adding it.

server/services/property.service.js
```diff
@@ -53,12 +53,39 @@ export const updateProperty = async (updatedProperty) => {
     throw new ErrorHandler(httpStatus.INTERNAL_SERVER_ERROR, 'Internal Server Error', error);
   });
 
+  let { units, availableUnits } = property;
+
+  if (updatedProperty.units) {
+    const initialUnits = property.units;
```
andela-hpopoola, line 59: I believe line 56 to 79 can be replaced with:

```js
const availableUnits = updatedProperty.units ? updatedProperty.units - property.units + property.availableUnits : property.availableUnits;

if (availableUnits < 0) {
   throw new ErrorHandler(
      httpStatus.PRECONDITION_FAILED,
      'Available property units cannot be less than zero',
   );
}
```

server/services/property.service.js
```diff
@@ -53,12 +53,39 @@ export const updateProperty = async (updatedProperty) => {
...
+    const updatedunits = updatedProperty.units;
```
andela-hpopoola, line 60: When writing compound words in Javascript, always ensure you use camelCasing:

```suggestion
    const updatedUnits = updatedProperty.units;
```

test/controllers/property.controller.test.js
```diff
@@ -650,6 +653,76 @@ describe('Property Controller', () => {
...
+              expect(res.body).to.have.property('property');
+              expect(res.body.property._id).to.be.eql(updatedUnit.id.toString());
+              expect(res.body.property.units).to.be.eql(updatedUnit.units);
+              expect(res.body.property.availableUnits).to.be.eql(0);
+              done();
+            });
+        });
+      });
+
+      context('when updated unit is less than initial unit', () => {
```
andela-hpopoola, line 703:
```suggestion
      context('when updated unit is less than available unit', () => {
```

server/services/property.service.js
```diff
@@ -53,12 +53,39 @@ export const updateProperty = async (updatedProperty) => {
...
+      if (availableUnits < 0) {
+        throw new ErrorHandler(
+          httpStatus.PRECONDITION_FAILED,
+          'Available property units cannot be less than zero',
+        );
+      }
+    }
+  }
+
+  const propertyToUpdate = { ...updatedProperty, units, availableUnits };
```
andela-hpopoola, line 81: Units is already defined in `updatedProperty`.

```suggestion
  const propertyToUpdate = { ...updatedProperty, availableUnits };
```

mailer/index.js
```diff
@@ -50,6 +50,16 @@ const EMAIL_CONTENT = {
     contentTop:
       'You have received an offer via ballers.ng for a [Property type] in [Property name]. This offer is Valid till [expiry date]. Check your dashboard for more details.',
   },
+  OFFER_EXPIRED_USER: {
+    subject: 'Offer Expired',
+    contentTop:
+      'Your offer for [Property name] has expired. Check your dashboard for more details.',
+  },
+  OFFER_EXPIRED_VENDOR: {
```
andela-hpopoola, line 58: It looks like this isn't used anywhere in the code. If we are not sending emails to vendors when an offer has expired, then we should remove this and use the one defined for Notifications.

mailer/index.js
```diff
@@ -72,6 +82,14 @@ const EMAIL_CONTENT = {
     contentBottom:
       'Kindly contact us if you have any questions concerning the payment or proposed payment methods.',
   },
+  PROPERTY_ALLOCATED_USER: {
```
andela-hpopoola, line 85: Same comment as above. If we are not sending a mail, lets remove.

mailer/index.js
```diff
@@ -72,6 +82,14 @@ const EMAIL_CONTENT = {
...
+    subject: 'Property Allocated',
+    contentTop: 'Congratulations!!! Property [Property name] has been allocated to you.',
+  },
+  PROPERTY_ASSIGNED: {
```
andela-hpopoola, line 89: Same comment as the 2 comments above.

andela-hpopoola (changes requested):
Thank you for working on this. I left a few comments, however, I noticed that you still have some unresolved comments (from last week). Since this is the only story you have a Pull Request for, I expect you to have finished working on this and resolved all comments (based on the number of hours you have worked on this). 

Can you please put more effort into this and improve your productivity 🙏🏽 ?

server/services/offer.service.js
```diff
@@ -585,29 +603,81 @@ export const assignOffer = async (offerId) => {
...
-  if (currentPercentagePaid >= allocationInPercentage) {
+  if (
+    isFirstPaymentForOffer &&
+    !paymentType.isLastPayment &&
+    currentPercentagePaid < allocationInPercentage
+  ) {
+    await assignOffer(offer._id);
+  }
+
+  if (currentPercentagePaid >= allocationInPercentage && !paymentType.isLastPayment) {
```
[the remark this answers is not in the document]
tobieeniafe: I believe the right status addition here should be `&& offer.status !== OFFER_STATUS.ALLOCATED`. Because if for a property of 10 million and an allocation percentage of 70%. If the first payment is 8 million, this would not run as the status would be `INTERESTED`

PR #263, opened 2021-10-12
#179881904 add needs attention endpoint

tobieeniafe:
#### What does this PR do?
Add needs attention endpoint

Add needs attention endpoint

`GET /api/v1/total-count/need-attention`

#179881904


server/services/totalCount.service.js
```diff
@@ -85,4 +85,28 @@ const getTotalCount = async (user) => {
...
-export default getTotalCount;
+export const getNeedAttention = async () => {
+  /* Admin
+   offer - PENDING_ADMIN_APPROVAL
+   property video - PENDING_ADMIN_REVIEW
+   offline payment - resolved false
+   reported property - resolved false
+   vas request - requested
+   bank account - approved false
+   property - flagged true
```
andela-hpopoola, line 96: Flagged Property doesn't require an admin's attention.

server/services/totalCount.service.js
```diff
@@ -85,4 +85,28 @@ const getTotalCount = async (user) => {
...
+   property - approved true
```
andela-hpopoola, line 97: I think this should be approved - false

server/services/totalCount.service.js
```diff
@@ -85,4 +85,28 @@ const getTotalCount = async (user) => {
...
+
```
andela-hpopoola, line 98: We might also want to add `remittance: false` for OFFER transactions to always remind the admin that they have not remitted a paid transaction to a vendor.

andela-hpopoola (changes requested):
Thank you for working on this. Your analysis looks great and sounds like a good start to make the implementation easier.

server/services/totalCount.service.js
```diff
@@ -85,4 +96,83 @@ const getTotalCount = async (user) => {
...
+    const pendingOfflinePayments = await OfflinePayment.countDocuments({
+      'resolved.status': false,
+    });
+    const pendingReportedProperties = await ReportedProperty.countDocuments({
+      'resolved.status': false,
+    });
+    const pendingVasRequests = await VasRequest.countDocuments({
+      status: VAS_REQUEST_STATUS.REQUESTED,
+    });
+    const pendingBankAccounts = await BankAccount.countDocuments({
```
andela-hpopoola, line 118: Since it is the admin that is adding and approving bank account, I don't think this is necessary for disapproved accounts. However, we can use this for `deactivated` accounts so that it can serve as a gentle reminder to the admin since that is supposed to be temporary.

server/services/totalCount.service.js
```diff
@@ -85,4 +96,83 @@ const getTotalCount = async (user) => {
   }
 };
 
-export default getTotalCount;
+export const getNeedAttention = async (user) => {
+  let response = {};
+
+  if (user.role === USER_ROLE.ADMIN) {
+    const pendingOffers = await Offer.countDocuments({
```
andela-hpopoola, line 103: Lets remove `pending` from the variable name (it is confusing for some of the variables that are not pending and would cause more harm than good in understanding the code in the future). Lets use the variable name directly, anyone consuming the endpoint will know the models that need attention. 

```suggestion
    const offers = await Offer.countDocuments({
```

server/services/totalCount.service.js
```diff
@@ -85,4 +96,83 @@ const getTotalCount = async (user) => {
...
+  if (user.role === USER_ROLE.VENDOR) {
+    const pendingEnquiries = await Enquiry.countDocuments({
+      vendorId: ObjectId(user._id),
+      approved: false,
+    });
+    const pendingOffers = await Offer.countDocuments({
+      vendorId: ObjectId(user._id),
+      status: OFFER_STATUS.PENDING_VENDOR_REVIEW,
+    });
+    const pendingPropertyVideos = await PropertyVideo.countDocuments({
```
andela-hpopoola, line 150: This is one of the reasons why adding pending before the variable name doesn't work. This is returning disapproved videos not pending videos - so it is safer to just use `propertyVideos` 😏 .

test/controllers/totalCount.controller.test.js
```diff
@@ -338,4 +344,286 @@ describe('TotalCount Controller', () => {
       user: editorUser,
     });
   });
+
+  describe('Get Needs Attention Count', () => {
+    const method = 'get';
+    const endpoint = '/api/v1/total-count/need-attention';
```
andela-hpopoola, line 350: I was contemplating if I should ask you to use `needs-attention` instead of the singular form. Your description on line 348 fixes that for me 😅. Lets update the endpoint to `needs-attention`, it feels more natural to read.

```suggestion
    const endpoint = '/api/v1/total-count/needs-attention';
```

andela-hpopoola (approved):
Thank you for working on this. This looks good to me, feel free to merge after fixing the comments.

